PR #3, opened 2021-06-17
Lots of buildscript 

i509VCB:


build-logic/src/main/groovy/qsl.library.gradle
```diff
@@ -0,0 +1,135 @@
...
+	id("net.kyori.indra.git")
+	id("qsl.common")
+}
+
+def extension = project.extensions.create("qslLibrary", QSLLibraryExtension, project)
+
+// We seem to have to use afterEvaluate to verify this stuff
+afterEvaluate {
+	// TODO: Validate the extension parts
+}
```
TheGlitch76, line 28: figure out what's going on with this

build-logic/src/main/groovy/qsl.module.gradle
```diff
@@ -0,0 +1,172 @@
...
+	// Version is set automatically in extension, verify it is set.
+	if (project.version == "unspecified") { // Gradle will return `unspecified` given no version being set.
+		throw new GradleException("Module $extension.moduleName needs a version to be set in the qslModule extension")
+	}
+
+	// Apply module dependencies for maven.
+	extension.setupModuleDependencies()
+	// TODO: Anything else to validate.
+
+	// TODO: Set qsl-library thing
```
TheGlitch76: ?

build-logic/src/main/java/qsl/internal/Versions.java
```diff
@@ -0,0 +1,27 @@
+package qsl.internal;
+
+/**
+ * Constants used across the convention scripts.
+ */
+public final class Versions {
```
TheGlitch76, line 8: why does this have to be in java, couldn't it be a .properties file or something?

build-logic/src/main/groovy/qsl.module.gradle
i509VCB: forgot to remove that todo

build-logic/src/main/java/qsl/internal/Versions.java
i509VCB: Can't be groovy since other convention scripts cant access it then. I dislike properties here since we have to load it for every single project then. Java here is the least painful and is on every project's classpath instantly.

build-logic/src/main/groovy/qsl.library.gradle
i509VCB: I believe I have an idea why:

This is simply based on plugin load order:

 * Plugin is loaded and empty extension is instantiated.

 * Entire convention buildscript is run.

 * extension is filled in.

Because of the above steps, this is required

i509VCB: @zml2008 could you explain if I am correct above?

TheGlitch76: nvm, misunderstood the comment, you're fine

build-logic/src/main/java/qsl/internal/Versions.java
LambdAurora: It's a bit unfortunate because for average/new contributors this is extremely un-intuitive and hard to find.

LambdAurora (approved):
I admit the buildscript scares me, but at least the build.gradle files in the library and modules are very clean and understandable.

build-logic/src/main/groovy/qsl.library.gradle
```diff
@@ -0,0 +1,135 @@
...
+				source(project.sourceSets.testmod)
+				property("quilt.autoTest")
+				programArg("--nogui")
+			}
+		}
+	}
+}
+
+// A library isn't truly a distributed artifact, rather it is just maven metadata to depend on the modules of a library.
+// For that reason, we have no ruse for the jar or remapJar tasks.
```
egelja:
```suggestion
// For that reason, we have no use for the jar or remapJar tasks.
```

build-logic/src/main/groovy/qsl.module.gradle
```diff
@@ -0,0 +1,171 @@
...
+dependencies {
+	// testmod sourceSet should depend on everything in the main source set.
+	testmodImplementation sourceSets.main.output
+}
+
+configurations {
+	// The reason for this `dev` configuration has been lost to time.
+	// If you happen to figure out why extension is needed, or a way to eliminate it, please tell us.
+	dev
+}
```
egelja, line 69: Where is the configuration defined?

build-logic/src/main/java/qsl/internal/Versions.java
egelja: Maybe make it clear in the README?

build.gradle
```diff
@@ -103,12 +84,13 @@ task checkVersion {
...
-			def versions = metadata.versioning.versions.version*.text();
+			def versions = metadata.versioning.versions.version*.text()
+
 			if (versions.contains(version)) {
 				throw new RuntimeException("${version} has already been released!")
 			}
-		} catch (FileNotFoundException ex) {
-			// Ignored
+		} catch (FileNotFoundException ignored) {
+			// No existing version of library found
```
MrAwesomeRocks, line 93: Maybe it may be better to print something, just in case someone later on ends up moving a library and Gradle can't find it.

build-logic/src/main/groovy/qsl.library.gradle
egelja, line 28: This is defined twice (L26 and L40)... is that going to break something?

library/core/crash-report-info/build.gradle
```diff
@@ -1,5 +1,12 @@
...
+	id("qsl.module")
 }
 
-coreDependencies(project, ["qsl-base"])
+qslModule {
+	moduleName = "crash-report-info"
+	version = "1.0.0"
+	library = "core"
+	coreDependencies([
+			"qsl-base"
```
egelja:
```suggestion
		"qsl-base"
```
Better indentation.

egelja (changes requested):
Everything looks good for the most part, just noticed a couple of little things.

build-logic/src/main/groovy/qsl.module.gradle
i509VCB: It is defined right there

build-logic/src/main/groovy/qsl.library.gradle
i509VCB: Nope, both after Evaluate blocks are run in the order they are declared.

egelja: Oh ok

build-logic/src/main/groovy/qsl.module.gradle
egelja: Isn't this just telling Gradle to use it, though?

OroArmor (approved):

build-logic/src/main/groovy/qsl.module.gradle
LambdAurora: The configurations block declare and defines the different configurations.

See https://docs.gradle.org/current/dsl/org.gradle.api.artifacts.ConfigurationContainer.html.

build.gradle
```diff
@@ -1,52 +1,22 @@
...
-			}
-		}
+sourceSets {
+	testmod {
+		compileClasspath += main.compileClasspath
+		runtimeClasspath += main.runtimeClasspath
 	}
 }
 
-jar {
```
TheGlitch76, line 46: you have a `jar { enabled = false }` for the libraries but not here, which is it?

settings.gradle
```diff
@@ -14,22 +14,29 @@ pluginManagement {
...
 
-			if (!properties.qsl_disabled) {
+	rootProject.projectDir.toPath().resolve("library/$library/").toFile().listFiles().each {
+		// Is the module disabled?
+		if (it.isDirectory()
+				&& it.name != "src" // Ignore sources
+				&& it.name != "build" // Ignore build artifacts
+				&& !it.name.startsWith(".") // Ignore anything hidden on unix-like OSes
+		) {
+			// If this file does not exist, include the project in the build.
```
TheGlitch76:
```suggestion
			// Libraries can be disabled by adding a file named DISABLE at the root of its directory
```

build.gradle
i509VCB: Probably need to disable it here. Might have over done it

i509VCB:
> Maybe make it clear in the README?

@MrAwesomeRocks I added an `UPDATING.md` to explain where the file is and documented why it's a class file inside of the Versions.java.

build.gradle
i509VCB: Done

README.md
```diff
@@ -0,0 +1,22 @@
+# QSL Buildscript Prototype
```
TheGlitch76: if you're going to start implementing QSL proper this is no longer a buildscript prototype

TheGlitch76 (changes requested):

i509VCB:
glitch said it's good on discord.

build-logic/src/main/java/qsl/internal/extension/QSLModuleExtension.java
```diff
@@ -0,0 +1,117 @@
...
+import org.gradle.api.plugins.JavaPlugin;
+import org.gradle.api.provider.Property;
+import org.gradle.api.publish.PublicationContainer;
+import org.gradle.api.publish.PublishingExtension;
+import org.gradle.api.publish.maven.MavenPublication;
+import org.gradle.api.tasks.Input;
+import qsl.internal.GroovyXml;
+import qsl.internal.json.ModJsonObject;
+
+public class QSLModuleExtension {
```
Prospector, line 26: I realize I'm a bit late and this is an extremely minor thing, but this class name does violate the Java class naming convention we usually use. Acronyms should be treated as words, so Qsl, not QSL

PR #9, opened 2021-07-07
Initial event api, license headers

i509VCB:
The first task here is the new event api for using with QSL. It is quite similar to the api fabric has minus some slight semantic differences and names. 

**one thing needs to be done before I merge this**

- [ ] Setup license headers


TheGlitch76 (changes requested):
I don't think that copy-pasting Fabric classes is the right way to do this at all. It was not sustainable for Patchwork and I don't want to support that approach here.

Maintaining a fork of Fabric API with the implementation modified to wrap around QSL is infinitely more maintainable because we can merge upstream and keep track of when they change things.

We're not assuming that *anything* in QSL is considered stable, so it is OK to create Fabric's APIs again almost verbatim (name style changing as needed) until we determine what changes we want.

i509VCB:
Hmm glitch that seems more reasonable tbf. However I'd want a few conditions:

Our backend reimplementation of fabric api MUST depend on a fixed version of QSL. Reason being I'll need to expose some internal compat bridges.

TheGlitch76:
Whatever you need--it's easier to change those kinds of things with experimentation, but it would be hard to lift FAPI out of QSL into a fork at a later date, so that's what's important to me.

i509VCB:
Ready for review. Grammar checking especially. 

Note the api is not final yet.

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/Event.java
```diff
@@ -0,0 +1,118 @@
...
+	 * @param invoker the invoker
+	 */
+	protected final void setInvoker(T invoker) {
+		this.invoker = invoker;
+	}
+
+	/**
+	 * Register a callback to the event.
+	 *
+	 * @param callback the callback.
```
LambdAurora:
```suggestion
	 * @param callback the callback
```

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/Event.java
```diff
@@ -0,0 +1,118 @@
...
+
+	/**
+	 * Sets the invoker instance.
+	 *
+	 * <p>This method is protected and final since it is only intended for implementors of the {@link Event} class to
+	 * set the invoker instance.
+	 *
+	 * <p>Since the invoker field is volatile, the invoker field should be available immediately for use.
+	 *
+	 * @param invoker the invoker
```
LambdAurora:
```suggestion
	 * @param invoker the invoker instance
```

For consistency with the getter.

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/ParameterInvokingEvent.java
```diff
@@ -0,0 +1,79 @@
...
+ *     	   // Invoke the callbacks implemented by this event
+ *         callback.spooked(ghost, spookedEntity);
+ *     }
+ * });
+ * </code></pre>
+ */
+@Documented // Documentation
+@Retention(RetentionPolicy.CLASS) // For static analysis
+@Target({ ElementType.FIELD, ElementType.METHOD })
+public @interface ParameterInvokingEvent {
```
LambdAurora: That annotation name is still super obscure, at least the doc is clear.

Technici4n: Do you have a better idea?

Technici4n:
I'm not sure why you're randomly changing some names (besides `ParameterInvokingEvent` which is a fine rename). `implementation` for example is really confusing. And `fallback` instead of `empty` is worse as well imo.

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/Event.java
```diff
@@ -0,0 +1,118 @@
...
+	 */
+	protected final void setInvoker(T invoker) {
```
Technici4n: Why? What is the benefit compared to a `protected` field?

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/Event.java
```diff
@@ -0,0 +1,118 @@
...
+ * of implementing an invoker and only allows callback implementations to be done by implementing an interface onto a
+ * class or extending a class.
+ *
+ * <h2>Example: Registering callbacks</h2>
+ *
+ * The most common use of an event will be registering a callback which is executed by the event. To register a callback,
+ * pass an instance of {@code T} into {@link #register}.
+ *
+ * <pre><code>
+ * public static final Event&lt;Runnable&gt; EXAMPLE = ...; // implementation
```
Pyrofab: I think this example should use a custom functional interface, especially since it shows implementing a callback onto a class. Using existing functional interfaces for event callbacks prevents implementing multiple on the same object, so I'd consider it bad practice.

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/ParameterInvokingEvent.java
LambdAurora: Sadly, no.
I don't think this can be named in a clear way anyway, and this is the best name so far.

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/Event.java
LambdAurora: I would say this is closer to OOP conventions, but I'm not sure why either.

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/ParameterInvokingEvent.java
i509VCB: Writing the documentation is probably the best way to communicate what the annotation does.

i509VCB:
Fallback vs empty I'd be fine changing with some discussion.

Implementation was an intentional choice since I have interpreted an Event as a holder object for a generated implementation of T. In the case of an array event, T is implemented via multiple callbacks that implement T.

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/Event.java
```diff
@@ -0,0 +1,118 @@
...
+	protected final void setInvoker(T invoker) {
+		this.invoker = invoker;
+	}
+
+	/**
+	 * Register a callback to the event.
+	 *
+	 * @param callback the callback.
+	 */
+	public abstract void register(T callback);
```
Haven-King: If we wanted to add callback IDs and dependencies, now would be the time to do it. The signature of `register` would then be something like
```java
/**
	 * Register a callback to the event.
	 *
     * @param id the identifier of this listener
	 * @param callback the callback
	 * @param after an array of identifiers this callback needs to be run after, if they are present
	 */
public abstract void register(Identifier id, T callback, Identifier... after);
```
This can fairly easily be implemented in a way that does not impact invocation-time performance. It does not solve all possible problems that might arise from invoker order, but it does offer one more (relatively pain free) mechanism to mod developers.

It's worth noting that for the purpose of Fabric mod compatibility, any implementation needs to be capable of handling cases where `id` is null, invoking it at any time.

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/ParameterInvokingEvent.java
Haven-King: `ParameterImplementedEvent` would be more clear to me, but at this point the documentation is good enough at describing its function that the name probably doesn't matter 😄

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/Event.java
Technici4n: `before` is necessary as well.

Haven-King: > `before` is necessary as well.

That complicates things slightly as we can no longer just use varargs, but would be fine with some overrides (and probably better method names 😛).
```java
/**
 * Register a callback to the event.
 *
 * @param id the identifier of this listener
 * @param callback the callback
 */
public void register(Identifier id, T callback) {
  this.register(id, callback, new Identifier[0], new Identifier[0]);
}

/**
 * Register a callback to the event.
 *
 * @param id the identifier of this listener
 * @param callback the callback
 * @param before an array of identifiers this callback needs to be run before, if they are present
 */
public void registerBefore(Identifier id, T callback, Identifier... before) {
  this.register(id, callback, before, new Identifier[0]);
}

/**
 * Register a callback to the event.
 *
 * @param id the identifier of this listener
 * @param callback the callback
 * @param after an array of identifiers this callback needs to be run after, if they are present
 */
public void registerAfter(Identifier id, T callback, Identifier... after) {
  this.register(id, callback, new Identifier[0], after);
}

/**
 * Register a callback to the event.
 *
 * @param id the identifier of this listener
 * @param callback the callback
 * @param before an array of identifiers this callback needs to be run before, if they are present
 * @param after an array of identifiers this callback needs to be run after, if they are present
 */
public abstract void register(Identifier id, T callback, Identifier[] before, Identifier[] after);
```

Leo40Git: This seems very workable!

Though, shouldn't those `T[]`/`T...` arrays be `Identifier[]`/`Identifier...`, according to the javadocs?

Haven-King: > 
> 
> This seems very workable!
> 
> Though, shouldn't those `T[]`/`T...` arrays be `Identifier[]`/`Identifier...`, according to the javadocs?

Oops, yes, I knew I missed something.

i509VCB: Varargs cannot be arbitrarily placed, only one per method and it must be last argument

Leo40Git: I know that, I was pointing out the incorrect typing on the varargs/arrays.

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/ParameterInvokingEvent.java
i509VCB: Ah the age old naming question.

We could go for a more Gradle-like naming scheme with something like `InvokesParameterAsCallback` for a name.

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/Event.java
i509VCB: Yeah it might be worth making it clear that people should make purpose specific functional interfaces.

i509VCB: Went ahead and made that change

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/Event.java
i509VCB: OOP, also allows a slightly nicer deprecation method if needed.

i509VCB:
@Technici4n returned to empty naming on the special `createArrayEvent`

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/Event.java
i509VCB: Not fully convinced on this. Do not the API is NOT final yet

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/EventFactory.java
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * <h2>Array event implementation</h2>
+ *
+ * The array event implementation provided by the Quilt Standard Libraries is a thread safe, copy-on-write event that
+ * generates an implementation of {@code T} through a user provided {@link Function function} to implement {@code T}
+ * using an array of callbacks.
+ *
+ * <p>Executing an event is the same as defined in the documentation of the {@link Event} class.
+ */
+public final class EventFactory {
```
Technici4n: What is the reasoning for this class? Couldn't these methods be in `Event` directly?

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/Event.java
```diff
@@ -0,0 +1,125 @@
...
+ * <ul>
+ *     <li>Synchronization when registering callbacks.
+ *     <ul>
+ *         <li>Registration of callbacks may occur on multiple threads, so thread safety when registering callbacks may
+ *         be useful.
+ *         <li>When the invoker is being executed, a thread safe implementation should ensure the current invoker's
+ *         state should not change. This means a newly registered invoker should only be visible for execution after
+ *         any current open executions of the invoker finish.
+ *     </ul>
+ * </ul>
```
Technici4n: Do note that fabric's `Event<T>` is not for extension by other mods (what would be the point after all). The javadoc is very lacking in that regard but I have plans to improve it.

Maybe having `@ApiStatus.NonExtendable` here would make sense?

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/Event.java
Technici4n: Not needed if it's not for extension.

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/Event.java
Technici4n: Not convinced either, as Lambda explained on discord it makes every ID a rigid part of a mod's API.

Technici4n (changes requested):
I think this is a good occasion to remove useless API complexity. Maybe `ArrayEvent` should even be inlined into `Event` directly.

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/Event.java
i509VCB: Some mods providing an implementation of a cascading event implementation is probably worth keeping extensibility for.

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/EventFactory.java
i509VCB: See ordering discussion on discord, if I do make array event public api, factory methods would be moved to the class then

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/Event.java
i509VCB: Screenshot of a compromise I might accept for int priorities given some surrounding conditions:

https://media.discordapp.net/attachments/522124386128494599/865243382997844008/Screenshot_20210715-094323_Discord.jpg

Haven seems to be on the other side of the coin on identifiers, I do note that identifiers have some advantages but some disadvantages that integers handle properly

> changing priority ordering is an intentional act, you must have something to be ordered and know what you are ordered before.
>
> no reordering is done given the absence of dependencies

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/ArrayEvent.java
```diff
@@ -0,0 +1,191 @@
...
+import java.util.Arrays;
+import java.util.Objects;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Function;
+
+/**
+ * An object which stores event callbacks in an array.
+ *
+ * <p>The factory methods for ArrayEvent allow the user to provide an implementation of {@code T} which is used to
```
TheGlitch76:
```suggestion
 * <p>The factory methods for ArrayEvent allows the user to provide an implementation of {@code T} which is used to
```

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/ArrayEvent.java
```diff
@@ -0,0 +1,191 @@
+package org.quiltmc.qsl.api.event;
```
TheGlitch76: don't forget to update this depending on what we decide the package structure is in the vote

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/ArrayEvent.java
```diff
@@ -0,0 +1,191 @@
...
+ * annotation on the type. You can let T not be a functional interface, however it heavily complicates the process
+ * of implementing an invoker and only allows callback implementations to be done by implementing an interface onto a
+ * class or extending a class.
+ *
+ * <h2>Example: Registering callbacks</h2>
+ *
+ * The most common use of an event will be registering a callback which is executed by the event. To register a callback,
+ * pass an instance of {@code T} into {@link #register}.
+ *
+ * <pre><code>
```
TheGlitch76: use `{@code` instead of `<code>` so you don't have to escape `@` and `<>`

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/ParameterInvokingEvent.java
```diff
@@ -0,0 +1,79 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * A marker interface that indicates an event stored in a field or returned by a method may invoke the event callback on
+ * one or more context objects passed in some method parameters.
+ *
+ * <p>The primary use of this annotation is for static analysis and documentation. Where a field or method return value
```
TheGlitch76:
```suggestion
 * <p>The primary use of this annotation is for static analysis and documentation. When a field or method return value
```

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/ParameterInvokingEvent.java
```diff
@@ -0,0 +1,79 @@
...
+ * is annotated as a parameter invoking, you are encouraged to note in the documentation which context parameters may be
+ * invoked.
+ *
+ * <p>Event implementations may have an undefined order in which the parameter may be invoked, see the documentation of
```
TheGlitch76: Is this what you mean?
```suggestion
 * <p>Event implementations may have an undefined order in which the parameters may be invoked, see the documentation of
```

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/ParameterInvokingEvent.java
```diff
@@ -0,0 +1,79 @@
...
+ * entities. With a parameter invoking event, if you directly control the object the action is being performed on such
+ * as the object being one of your own modded entities, you can implement the callback interface, {@code T}, and have
+ * the event execute the callback on your own object, removing the need to find your specific entity through further
+ * checks.
+ *
+ * <p>Assuming {@code ScaryEvents.SPOOKED}'s callback interface is {@code Spooked} and the event is parameter invoking,
+ * you may implement {@code ScaryEvents.Spooked} directly on to your Entity to be notified that when your entity is
+ * spooked.
+ *
+ * <pre><code>
```
TheGlitch76: Same here, use `{@code`

TheGlitch76 (changes requested):
Mostly just some documentation clunk that needs fixed, and this shouldn't be merged until we get the vote on #10 tomorrow or so.

TheGlitch76:
I'm still unsure if we should resolve the debate of int/string/no priority before merging this PR.
We can swap it out at any time before API is stabilized, of course, but that doesn't mean we should wait until the last moment.

Leo40Git:
Heads up: decided package structure is `org.quiltmc.qsl.$module.api/impl`!

i509VCB:
I think we are done with this.

library/core/qsl-base/src/main/java/org/quiltmc/qsl/api/event/Event.java
i509VCB: There is now an issue for this chain: https://github.com/QuiltMC/quilt-standard-libraries/issues/14

LambdAurora (approved):

PR #15, opened 2021-07-22
I made a fatal mistake: event api, license headers

i509VCB:
Given the big oops I just did by accidentally merging #9 this PR is here again


i509VCB:
Okay so, make an issue if you saw something wrong...

PR #17, opened 2021-07-23
Resource loader module.

LambdAurora:
Here's the resource loader module!

What is done:

 - [x] Resource reloaders.
 - [x] Mod resource packs.
 - [x] Built-in resource pack API.
 - [x] Programmer art API.
 - [x] Group resource pack API.
 - [x] Replace MC's vanilla resource pack implementation with a NIO implementation.
 - [x] Replace MC's default resource pack with a group resource pack containing a NIO replacement of MC's resource pack and the mod resource packs.
 - [ ] Virtual resource packs API (bonus only, could be done later).
 
I already open the PR as a draft so I can get early reviews to avoid having a big blob to review in the end.


library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/GroupResourcePack.java
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * Represents a group resource pack, which holds multiple resource packs as one.
+ */
+public abstract class GroupResourcePack implements ModResourcePack {
+	protected final ResourceType type;
+	protected final List<ModResourcePack> packs;
+	protected final Map<String, List<ModResourcePack>> namespacedPacks = new Object2ObjectOpenHashMap<>();
+
+	public GroupResourcePack(ResourceType type, List<ModResourcePack> packs) {
```
Kroppeb: I would prefer `List<? extends ModResourcePack> packs`. This should avoid unnecessary unchecked casts in cases like where someone has a `List<GroupResourcePack>` and wants to combine them into a single `GroupResourcePack`.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/GroupResourcePack.java
```diff
@@ -0,0 +1,128 @@
...
+	public List<ModResourcePack> getPacks(String namespace) {
+		return this.namespacedPacks.get(namespace);
+	}
+
+	@Override
+	public InputStream open(ResourceType type, Identifier id) throws IOException {
+		List<ModResourcePack> packs = this.namespacedPacks.get(id.getNamespace());
+
+		if (packs != null) {
+			for (int i = packs.size() - 1; i >= 0; i--) {
```
Kroppeb, line 74: I'd suggest adding a comment explaining why we are iterating backwards. I'd assume it's something like
```java
// Iterating backwards as later packs should override the resources defined in earlier packs
```

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/GroupResourcePack.java
```diff
@@ -0,0 +1,128 @@
...
+
+	@Override
+	public boolean contains(ResourceType type, Identifier id) {
+		List<ModResourcePack> packs = this.namespacedPacks.get(id.getNamespace());
+
+		if (packs == null) {
+			return false;
+		}
+
+		for (int i = packs.size() - 1; i >= 0; i--) {
```
Kroppeb, line 117: I'd suggest adding a comment explaining why we are iterating backwards. I don't actually understand why this is the case.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/GroupResourcePack.java
```diff
@@ -0,0 +1,128 @@
...
+	public Collection<Identifier> findResources(ResourceType type, String namespace, String prefix, int maxDepth, Predicate<String> pathFilter) {
+		List<ModResourcePack> packs = this.namespacedPacks.get(namespace);
+
+		if (packs == null) {
+			return Collections.emptyList();
+		}
+
+		Set<Identifier> resources = new HashSet<>();
+
+		for (int i = packs.size() - 1; i >= 0; i--) {
```
Kroppeb, line 98: I'd suggest adding a comment explaining why we are iterating backwards. I don't actually understand why this is the case.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceLoaderImpl.java
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	public static void sort(ResourceType type, List<ResourceReloader> reloaders) {
+		get(type).sort(reloaders);
+	}
+
+	@Override
+	public void registerReloader(IdentifiableResourceReloader resourceReloader) {
+		if (!this.addedListenerIds.add(resourceReloader.getQuiltId())) {
+			LOGGER.warn("Tried to register resource reloader " + resourceReloader.getQuiltId() + " twice!");
```
TheGlitch76: Do we really want this to fail silently?
Is there any instance where this would happen that isn't a bug?

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ModNioResourcePack.java
```diff
@@ -0,0 +1,215 @@
...
+					LOGGER.warn("findResources at " + path + " in namespace " + namespace + ", mod " + modInfo.getId() + " failed!", e);
+				}
+			}
+		}
+
+		return ids;
+	}
+
+	protected void warnInvalidNamespace(String s) {
+		LOGGER.warn("Fabric NioResourcePack: ignored invalid namespace: {} in mod ID {}", s, modInfo.getId());
```
TheGlitch76: Shouldn't this be Quilt?
```suggestion
		LOGGER.warn("Quilt NioResourcePack: ignored invalid namespace: {} in mod ID {}", s, modInfo.getId());
```

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ModNioResourcePack.java
```diff
@@ -0,0 +1,215 @@
...
+		}
+
+		stream = ModResourcePackUtil.openDefault(this.modInfo, this.type, filename);
+
+		if (stream != null) {
+			return stream;
+		}
+
+		// ReloadableResourceManagerImpl gets away with FileNotFoundException.
+		throw new FileNotFoundException("\"" + filename + "\" in Fabric mod \"" + modInfo.getId() + "\"");
```
TheGlitch76:
```suggestion
		throw new FileNotFoundException("\"" + filename + "\" in Quilt mod \"" + modInfo.getId() + "\"");
```

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourcePackActivationType.java
```diff
@@ -0,0 +1,46 @@
...
+	 * Always enabled. The user cannot disable the resource pack.
+	 */
+	ALWAYS_ENABLED;
+
+	/**
+	 * Returns whether this resource pack will be enabled by default or not.
+	 *
+	 * @return {@code true} if enabled by default, else {@code false}
+	 */
+	public boolean isEnabledByDefault() {
```
TheGlitch76, line 41: We've dropped get/is prefixes in Loader. Should we do the same here?
cc @i509VCB

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceLoaderImpl.java
LambdAurora: I guess in the case that some mods register reloaders with the same ID?
Shouldn't happen at all though, unsure, might make it throw a IllegalArgumentException there.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourcePackActivationType.java
i509VCB: Not just a plain drop, sometimes get/is is required depending on the context, but yes.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ModResourcePack.java
```diff
@@ -0,0 +1,34 @@
...
+package org.quiltmc.qsl.resource.loader.api;
+
+import net.fabricmc.loader.api.metadata.ModMetadata;
+
+import net.minecraft.resource.ResourcePack;
+
+/**
+ * Interface implemented by mod-provided resource packs.
+ */
+public interface ModResourcePack extends ResourcePack {
```
i509VCB: I think we should avoid this scenario where people have to upcast the game types to an high level type in our API. Probably better to just have a method to lookup this additional metadata.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/GroupResourcePack.java
```diff
@@ -0,0 +1,128 @@
...
+
+import net.minecraft.resource.ResourceNotFoundException;
+import net.minecraft.resource.ResourcePack;
+import net.minecraft.resource.ResourceType;
+import net.minecraft.util.Identifier;
+
+/**
+ * Represents a group resource pack, which holds multiple resource packs as one.
+ */
+public abstract class GroupResourcePack implements ModResourcePack {
```
i509VCB: Hmm what was the purpose of this again? Might be worth describing the use case in docs

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourceLoader.java
```diff
@@ -0,0 +1,45 @@
...
+
+import net.minecraft.resource.ResourceType;
+
+import org.quiltmc.qsl.resource.loader.api.reloader.IdentifiableResourceReloader;
+import org.quiltmc.qsl.resource.loader.impl.ResourceLoaderImpl;
+
+/**
+ * Represents the resource loader. Contains different register methods.
+ */
+public interface ResourceLoader {
```
i509VCB, line 33: We don't have sealed classes till 17-ga, mark this as NonExtendable with JB annotations

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/reloader/IdentifiableResourceReloader.java
```diff
@@ -0,0 +1,50 @@
...
+ * Interface for "identifiable" resource reloaders.
+ *
+ * <p>"Identifiable" resource reloaders have an unique identifier, which can be depended on,
+ * and can provide dependencies that they would like to see executed before themselves.
+ *
+ * @see ResourceReloaderKeys
+ */
+public interface IdentifiableResourceReloader extends ResourceReloader {
+	/**
+	 * {@return the unique identifier of this resource reloader}
```
i509VCB, line 35: Is the return tag supposed to be wrapped in brackets?

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/reloader/ResourceReloaderKeys.java
```diff
@@ -0,0 +1,44 @@
...
+package org.quiltmc.qsl.resource.loader.api.reloader;
+
+import net.minecraft.util.Identifier;
+
+/**
+ * This class contains default keys for various Minecraft resource reloaders.
+ *
+ * @see IdentifiableResourceReloader
+ */
+public final class ResourceReloaderKeys {
```
i509VCB, line 26: Do we want to keep these seperated between client and server? Could be done via two seperate classes or add a Client and server static inner class for this

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/reloader/SimpleResourceReloader.java
```diff
@@ -0,0 +1,78 @@
...
+ * </ul>
+ *
+ * <p>The load stage should be self-contained as it can run on any thread! However,
+ * the apply stage is guaranteed to run on the game thread.
+ *
+ * <p>For a fully synchronous alternative, consider using
+ * {@link SynchronousResourceReloader} in conjunction with
+ * {@link IdentifiableResourceReloader}.
+ *
+ * @param <T> The data object.
```
i509VCB: Might be more accurate to describe T as the type of the intermediate reloading state?

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/KeyedResourceReloaderMixin.java
```diff
@@ -0,0 +1,83 @@
...
+
+import org.quiltmc.qsl.resource.loader.api.reloader.IdentifiableResourceReloader;
+import org.quiltmc.qsl.resource.loader.api.reloader.ResourceReloaderKeys;
+
+@Mixin({
+		RecipeManager.class, ServerAdvancementLoader.class, FunctionLoader.class,
+		LootManager.class, TagManagerLoader.class
+})
+public abstract class KeyedResourceReloaderMixin implements IdentifiableResourceReloader {
+	private Identifier quilt$id;
```
i509VCB: Throw some `@Unique` annotations on these fields and ditch the quilt prefix

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/NamespaceResourceManagerAccessor.java
```diff
@@ -0,0 +1,36 @@
...
+import net.minecraft.resource.ResourceType;
+import net.minecraft.util.Identifier;
+
+@Mixin(NamespaceResourceManager.class)
+public interface NamespaceResourceManagerAccessor {
+	@Accessor("type")
+	ResourceType getType();
+
+	@Invoker("getMetadataPath")
+	static Identifier quilt$accessor_getMetadataPath(Identifier id) {
```
i509VCB: Mixin makes a guarantee that accessor Mixins will never have conflicting methods, the issue only arises when a duck exists that is named the same thing.

quilt prefix here is probably not needed.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/reloader/IdentifiableResourceReloader.java
LambdAurora: Yes, it generates the summary "Returns <what's in the brackets>." automatically, though the Intellij previewer has a bit of a hard time with it.

i509VCB: So it's doclet stuff then

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/KeyedResourceReloaderMixin.java
LambdAurora: Wished the `@Unique` annotation would make it clear which mod adds the fields instead of throwing random characters.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/NamespaceResourceManagerAccessor.java
TheGlitch76: IMO it's still a good idea to give these prefixes--the other major projects (Sponge, Fabric [iirc]) and their derivatives (Patchwork) use this style.

And in general it's best to avoid anything that could fail in Mixin because Mixin loves failing silently and it's awful to debug

LambdAurora: Everytime I used an `Invoker` without a prefix I ended up with a crash because of "same name" somehow.

i509VCB: To clarify, the `quilt$` and `accessor_` are redundant here, one or the other has to go

LambdAurora: I think I remember why the `accessor_` part is there, because I was the one who PRed this accessor to FAPI, and originally it was just `namespace$` prefixed, but I was told that I should've made it clear that it's an accessor and not a new method.

But the issue is, if I use `accessor_` and someone else put an invoker with the same `accessor_` prefix, won't I risk a crash?

i509VCB: Mixin seems to handle invokers properly with this aliasing thing, as many mods put accessors on very common things

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/GroupResourcePack.java
LambdAurora: In FAPI it's impl-only, but the goal of this is to be able to group multiple resource packs as one unique entry in the resource pack list.

Which will be used for mod resource packs and vanilla resource pack, as it'll be one unique entry (Default) in the resource pack screen.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/NamespaceResourceManagerAccessor.java
OroArmor: Not anymore. That was fixed in mixin 0.8.3 iirc

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ModResourcePack.java
LambdAurora: I'm not sure to get what you're saying?

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/NamespaceResourceManagerAccessor.java
TheGlitch76: accessors shouldn't ever conflict, when you got the crash were you using `accessor_getMetadataPath` or just `getMetadataPath`?

LambdAurora: With just `getMetadataPath` it crashed I think?
It's been a while tbh so it might have changed since.

TheGlitch76: Yeah, you were conflicting with a non-invoker then

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/KeyedResourceReloaderMixin.java
```diff
@@ -0,0 +1,83 @@
...
+			} else if (self instanceof ServerAdvancementLoader) {
+				this.quilt$id = ResourceReloaderKeys.ADVANCEMENTS;
+			} else if (self instanceof FunctionLoader) {
+				this.quilt$id = ResourceReloaderKeys.FUNCTIONS;
+			} else if (self instanceof LootManager) {
+				this.quilt$id = ResourceReloaderKeys.LOOT_TABLES;
+			} else if (self instanceof TagManagerLoader) {
+				this.quilt$id = ResourceReloaderKeys.TAGS;
+			} else {
+				this.quilt$id = new Identifier("minecraft",
```
Kroppeb: Shouldn't this throw instead?

LambdAurora: Unsure, that's inheriting old FAPI code.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/reloader/SimpleResourceReloader.java
LambdAurora: Hopefully the new doc is better?

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/reloader/ResourceReloaderKeys.java
LambdAurora: Separated everything into static inner classes.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ModResourcePack.java
Leo40Git: I think what he's saying is, instead of:
```java
ModMetadata modMeta;
if (resPack instanceof ModResourcePack modResPack)
    modMeta = modResPack.getQuiltModMetadata();
else
    modMeta = null; // vanilla/standard resource pack
```
we should have:
```java
ModMetadata modMeta = ModResourcePack.getQuiltModMetadata(resPack);
```
or... something like that?

IMO, `instanceof` pattern matching makes this point moot.

LambdAurora: Well, tbh the first question we should ask is "Is the mod metadata really that useful in resource packs?"

All resource packs are named by default, which makes them already identifiable, and the only use of ModMetadata in the whole resource loader module is:
 - getting the mod path out of it, but that's usually internal, it happens once when registering a built-in resource pack, and once when every mod resource packs are built.
 - generating a default `pack.mcmeta` for mod resource packs
 
 Externally there's literally no use, so is it even that useful to expose this in API?
 The pack.mcmeta generation could be exposed in API, and if a mod needs it it just can fetch the corresponding mod metadata if needed because the only use case of it would be if the mod makes its own resource pack, so no identification issue there.

spaceclouds42:
Concern about where features should be between qloader and qsl:
 what if the quilt mod loader, _loaded_ mods' resources. An API pertaining to resources makes sense in qsl, but the actual loading of resources from mods should be loaded by the thing that _loads mods_. Shouldn't have to depend on a qsl module or write a resource loader yourself for your mod's resources to be loaded.

OroArmor:
This make sense, but loader is version agnostic. Hooking into minecraft internals is not the purpose of the loader, which is why it is part of qsl

LambdAurora:
> Concern about where features should be between qloader and qsl:
> what if the quilt mod loader, _loaded_ mods' resources. An API pertaining to resources makes sense in qsl, but the actual loading of resources from mods should be loaded by the thing that _loads mods_. Shouldn't have to depend on a qsl module or write a resource loader yourself for your mod's resources to be loaded.

Resource loader is much more than just "feed MC mods as resource packs", it significantly modifies Vanilla behavior:
 - it groups multiple resource packs as one (all the mods resource packs + Vanilla pack -> Default pack)
 - it offers an API for mods to add more resource packs (built-in, virtual)
 - API to trigger code when a resource reload happens (to load custom resources)
 - programmer art API (aka, allows mods to define programmer art textures for their own blocks/items/etc.)
 It wouldn't fit in loader and can't exist as a loader plugin, with all of these features it has to be a QSL module.

spaceclouds42:
As I said, some of those features make sense as a part of qsl, but specifically the part that loads mods' resources into a resource pack, that part should be a part of the loader. However, I hadn't thought of the loader having to be version agnostic.
> can't exist as a loader plugin

Why cant strictly the basic resource loading part be a plugin? Or are plugins also supposed to be version agnostic and not touch mc

FoundationGames:
> Why cant strictly the basic resource loading part be a plugin? Or are plugins also supposed to be version agnostic

@SpaceClouds42 Loader plugins don't touch game code, you absolutely need to touch and modify game code to implement a resource loader

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ModResourcePack.java
LambdAurora: I removed the ModResourcePack interface entirely.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/reloader/ResourceReloaderKeys.java
```diff
@@ -0,0 +1,60 @@
...
+	 */
+	public static final class Client {
+		public static final Identifier SOUNDS = new Identifier("minecraft:sounds");
+		public static final Identifier FONTS = new Identifier("minecraft:fonts");
+		public static final Identifier MODELS = new Identifier("minecraft:models");
+		public static final Identifier LANGUAGES = new Identifier("minecraft:languages");
+		public static final Identifier TEXTURES = new Identifier("minecraft:textures");
+
+		private Client() {
+			throw new UnsupportedOperationException("Someone tried to instantiate a static-only class. How?");
```
TheGlitch76: IMO these UnsupportedOperationExceptions just add noise to the code. It would be better to just leave it as an empty code block.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourcePackActivationType.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+public enum ResourcePackActivationType {
+	/**
+	 * Normal activation. The user has full control over the activation of the resource pack.
+	 */
+	NORMAL,
+	/**
+	 * Enabled by default. The user has still full control over the activation of the resource pack.
+	 *
+	 * <p>Note: this setting can only be satisfied on data packs, client resource packs cannot be by default enabled.
```
TheGlitch76: This is really nit-picky, but this word order is really awkward.
```suggestion
	 * Enabled by default. The user still has full control over the activation of the resource pack.
	 *
	 * <p>Note: this setting can only be satisfied on data packs, client resource packs cannot be enabled by default.
```

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourcePackActivationType.java
TheGlitch76: I'd say this is a case where `enabledByDefault` is fine.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/package-info.java
```diff
@@ -0,0 +1,35 @@
...
+ *  <li>Resource Pack refers to both client-sided resource pack and data pack.</li>
+ *  <li>Virtual Resource Pack refers to a resource pack that may be generated at runtime, or simply doesn't exist directly on disk.</li>
+ *  <li>Group Resource Pack refers to a virtual resource pack that groups multiple resource packs together.</li>
+ * </ul>
+ * </p>
+ *
+ * <p><h3>Modded Resource Pack Handling</h3></p>
+ * <p>The Resource Loader will create a resource pack for each mod that provides resources in {@code assets} or {@code data}
+ * sub-directories.
+ * Those mod resource packs are grouped into the default resource pack, they can override the default resource pack but unrecommended.</p>
```
TheGlitch76:
```suggestion
 * Those mod resource packs are grouped into the default resource pack. They can override the default resource pack, but it is not recommended.</p>
```
Also, why is it not recommended? This seems like it could use a "because".

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/package-info.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * <p><h4>Built-in Mod Resource Pack</h4></p>
+ * <p>The Resource Loader adds manually registered mod resource packs. Those resource packs are registered with
+ * {@link org.quiltmc.qsl.resource.loader.api.ResourceLoader#registerBuiltinResourcePack(net.minecraft.util.Identifier, net.fabricmc.loader.api.ModContainer, ResourcePackActivationType)}</p>
+ *
+ * <p><h4>Programmer Art Resource Pack</h4></p>
+ * <p>The Resource Loader will inject resources into the Programmer Art resource pack for each mod that provides
+ * Programmer Art resources in the {@code programmer_art} top-level directory of the mod
```
TheGlitch76, line 29: You refer to what I am presuming is the same location (the subfolders you create in the `resources` folder) as being both a sub-directory and a top-level directory in this documentation.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ModNioResourcePack.java
```diff
@@ -0,0 +1,215 @@
...
+			return Files.newInputStream(path);
+		}
+
+		stream = ModResourcePackUtil.openDefault(this.modInfo, this.type, filename);
+
+		if (stream != null) {
+			return stream;
+		}
+
+		// ReloadableResourceManagerImpl gets away with FileNotFoundException.
```
TheGlitch76: gets away? what does that mean?

TheGlitch76 (changes requested):
Code is looking good, but I have some minor issues with the documentation

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceLoaderImpl.java
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	public static void sort(ResourceType type, List<ResourceReloader> reloaders) {
+		get(type).sort(reloaders);
+	}
+
+	@Override
+	public void registerReloader(IdentifiableResourceReloader resourceReloader) {
+		if (!this.addedListenerIds.add(resourceReloader.getQuiltId())) {
+			throw new RuntimeException("Tried to register resource reloader " + resourceReloader.getQuiltId() + " twice!");
```
i509VCB: There may be a better exception type than runtime exception for these? Runtime exception is quite broad

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/DefaultResourcePackMixin.java
```diff
@@ -0,0 +1,103 @@
...
+	@Unique
+	private ModNioResourcePack locateAndLoad() {
+		return ResourceLoaderImpl.locateAndLoadDefaultResourcePack(
+				this.getClass().equals(DefaultResourcePack.class) ?
+						ResourceType.SERVER_DATA : ResourceType.CLIENT_RESOURCES
+		);
+	}
+
+	/**
+	 * @author QuiltMC
```
i509VCB: Author should refer to the code author(s) probably and not quiltmc

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/DefaultResourcePackMixin.java
```diff
@@ -0,0 +1,103 @@
...
+	 * @reason Rewrite default resource access to avoid resource leaking.
+	 */
+	@Overwrite
```
i509VCB, line 63: Are these overwrites to fix that eclipse classpath bug? Or is there more context. If this fixes classpath for eclipse then do mention it in class docs so people understand the purpose of these.

library/core/resource-loader/src/testmod/java/org/quiltmc/qsl/resource/loader/test/ResourceReloaderTestMod.java
```diff
@@ -0,0 +1,129 @@
...
+	private static boolean clientResources = false;
+	private static boolean serverResources = false;
+
+	@Override
+	public void onInitialize() {
+		this.setupClientReloadListeners();
+		this.setupServerReloadListeners();
+
+		// No lifecycle events yet
+		/*ServerTickEvents.START_WORLD_TICK.register(world -> {
```
i509VCB: Should be changed to server started when lifecycle events are brought back in

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ModNioResourcePack.java
i509VCB: To me it reads like "resource not found, so fail" in surrounding code.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/DefaultResourcePackMixin.java
LambdAurora: In the mixin overwrite context it mostly refers to who provide the mixin imo.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ModNioResourcePack.java
LambdAurora: It means that here the resource manager will catch the exception and return `null` iirc.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/DefaultResourcePackMixin.java
LambdAurora: I mean, the reason field does mention that this is to fix the resource leaking issue.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/DefaultResourcePackMixin.java
i509VCB: So two conflicting overwrites will note the authors?

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/reloader/ResourceReloaderKeys.java
LambdAurora: IMO static-only classes being initialized should be considered an error.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/DefaultResourcePackMixin.java
LambdAurora: I'm pretty sure it does?

mcdev warns about it needing an author and reason fields in the doc at least.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/reloader/ResourceReloaderKeys.java
i509VCB: I think javac is a good enough guard here. If someone does something stupid we can't stop them

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceLoaderImpl.java
LambdAurora: Replaced with IllegalStateExceptions.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/package-info.java
LambdAurora: Now it's top-level directory everywhere.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/KeyedResourceReloaderMixin.java
LambdAurora: This actually makes more sense in the client version of the mixin as some of them fall into the else block.

i509VCB: I'd probably add a message guarded by the development environment noting the class if we don't have it covered by the statement. In log

LambdAurora: Meh, from the look of the old code those classes having for fallback this is pretty much intended.

LambdAurora:
Ready for review, all the essential features and parity features with Fabric API are implemented.
Virtual resource pack stuff can be done later as the base feature of this module are required to test and make other modules work.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/GroupResourcePack.java
```diff
@@ -39,6 +39,14 @@
 
 /**
  * Represents a group resource pack, which holds multiple resource packs as one.
+ *
+ * <p>
+ * The possible use cases are:
+ * <ul>
+ *  <li>bundling multiple resource packs as one to pollute less the user's UI</li>
```
Leo40Git: Grammar/spelling fixes:
```html
<ul>
  <li>bundling multiple resource packs as one to reduce pollution of the user's UI</li>
  <li>replacing the default resource pack with a combination of the default one and all mods' resource packs</li>
  <li>etc.</li>
</ul>
```

build-logic/src/main/groovy/qsl.module.gradle
```diff
@@ -123,7 +123,29 @@ java {
...
+		source = "16"
+		encoding = "UTF-8"
+		charSet = "UTF-8"
+		memberLevel = JavadocMemberLevel.PACKAGE
+		links(
+				"https://guava.dev/releases/21.0/api/docs/",
+				"https://asm.ow2.io/javadoc/",
+				"https://docs.oracle.com/javase/8/docs/api/",
+				"https://jenkins.liteloader.com/job/Mixin/javadoc/",
+				"https://logging.apache.org/log4j/2.x/log4j-api/apidocs/"
```
OroArmor: Should we add Yarn/QuiltMappings links to this?

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ModResourcePackProvider.java
```diff
@@ -0,0 +1,81 @@
...
+			5. (Invisible) High-priority virtual resource packs
+		 */
+
+		ResourceLoaderImpl.registerBuiltinResourcePacks(this.type, profileAdder, factory);
+	}
+
+	private static ResourcePackSource nameAndSource(String source) {
+		Text text = new TranslatableText(source);
+		return name -> (new TranslatableText("pack.nameAndSource", name, text)).formatted(Formatting.GRAY);
+	}
```
OroArmor: Is this method needed? Its only used in one place and can easily be inlined.

OroArmor (approved):
Looks great to me. ~~Need this to be merged for my own pull request.~~

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ModResourcePackProvider.java
LambdAurora: The method allows caching of the source text.

build-logic/src/main/groovy/qsl.module.gradle
LambdAurora: If it's already generated, yes.

OroArmor: It does not look like it is published. This can be resolved once quilt mappings is out

OroArmor: ok so its being published, but there doesn't seem to be a forward facing URL to read the docs. I don't know how its automatically done on the fabric repository so ill look into it

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/MinecraftServerMixin.java
```diff
@@ -0,0 +1,54 @@
...
+@Mixin(MinecraftServer.class)
+public class MinecraftServerMixin {
+	@Redirect(method = "loadDataPacks", at = @At(value = "INVOKE", target = "Ljava/util/List;contains(Ljava/lang/Object;)Z"))
+	private static boolean onCheckDisabled(List<String> list, Object o, ResourcePackManager resourcePackManager) {
+		var profileName = (String) o;
+		boolean contains = list.contains(profileName);
+
+		if (contains) {
+			return true;
+		}
```
BasiqueEvangelist: Why not just
```suggestion
		if (list.contains(profileName)) {
			return true;
		}
```

LambdAurora: Pretty sure that was a remain of in-dev code for when I made it months ago.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourceLoader.java
```diff
@@ -0,0 +1,107 @@
...
+	 * @param id             the identifier of the resource pack, its namespace must be the same as the mod id
+	 * @param activationType the activation type of the resource pack
+	 * @return {@code true} if successfully registered the resource pack, else {@code false}
+	 * @see #registerBuiltinResourcePack(Identifier, ModContainer, ResourcePackActivationType)
+	 */
+	static boolean registerBuiltinResourcePack(Identifier id, ResourcePackActivationType activationType) {
+		return FabricLoader.getInstance().getModContainer(id.getNamespace())
+				.map(container -> registerBuiltinResourcePack(id, container, activationType))
+				.orElse(false);
+	}
```
BasiqueEvangelist: Why does this just return false if the mod couldn't be found? Wouldn't throwing an exception be better, since modders probably won't be trying to register invalid packs?

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ModResourcePackProvider.java
BasiqueEvangelist: Isn't this just a copy of `ResourcePackSource.nameAndSource`? Why not use that instead?

LambdAurora: Reference equality is different, which is used to our advantage.

BasiqueEvangelist: You mean like for the lambda?
Won't that be created every time anyway, considering it uses a variable from the outer scope (`text`)

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/reloader/SimpleResourceReloader.java
```diff
@@ -0,0 +1,78 @@
...
+ *     <li>{@linkplain #apply(Object, ResourceManager, Profiler, Executor)}: apply the information from the data object
+ * to the game instance.
+ * </ul>
+ *
+ * <p>The load stage should be self-contained as it can run on any thread! However,
+ * the apply stage is guaranteed to run on the game thread.
+ *
+ * <p>For a fully synchronous alternative, consider using
+ * {@link SynchronousResourceReloader} in conjunction with
+ * {@link IdentifiableResourceReloader}.
```
BasiqueEvangelist: This should probably recommend `SimpleSynchronousResourceReloader` instead.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ModResourcePackProvider.java
LambdAurora: What I mean is the lambda is of the type `ResourcePackSource`, a `ResourcePackProfile` has a method to get the source which returns `ResourcePackSource`, we then use a reference equality to differentiate the quilt's one from the Vanilla's one.

BasiqueEvangelist: Well, yes, but can't you use the same method to make the quilt `ResourcePackSource`?

LambdAurora: It's package-private and I found the use of an accessor quite ridiculous in this case.

BasiqueEvangelist: > It's package-private

Is it though? I made line 39 use `ResourcePackSource`'s method and it built completely fine.

LambdAurora: I have been deceived.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourceLoader.java
LambdAurora: It now throws an IllegalArgumentException if it cannot find the mod.

BasiqueEvangelist (approved):

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/reloader/SimpleResourceReloader.java
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * @param <T> the type for the intermediate reloading state
+ */
+public interface SimpleResourceReloader<T> extends IdentifiableResourceReloader {
+	@Override
+	default CompletableFuture<Void> reload(ResourceReloader.Synchronizer helper, ResourceManager manager,
+										   Profiler loadProfiler, Profiler applyProfiler,
+										   Executor loadExecutor, Executor applyExecutor) {
+		return this.load(manager, loadProfiler, loadExecutor).thenCompose(helper::whenPrepared)
+				.thenCompose(o -> this.apply(o, manager, applyProfiler, applyExecutor));
```
TheGlitch76, line 52: i assume this method has a contract of only ever running on the game thread?

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/DefaultResourcePackMixin.java
TheGlitch76: yeah, but plenty of people have no idea what this "well-known" bug is.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/ReloadableResourceManagerImplMixin.java
```diff
@@ -0,0 +1,73 @@
...
+			at = @At(value = "INVOKE", target = "Lorg/apache/logging/log4j/Logger;isDebugEnabled()Z", remap = false)
+	)
+	private void reload(Executor prepareExecutor, Executor applyExecutor, CompletableFuture<Unit> initialStage,
+						List<ResourcePack> packs,
+						CallbackInfoReturnable<ResourceReload> info) {
+		ResourceLoaderImpl.sort(this.type, this.reloaders);
+	}
+
+	// private static synthetic method_29491(Ljava/util/List;)Ljava/lang/Object;
+	// Supplier lambda in beginMonitoredReload method.
```
TheGlitch76: Put this text in the value of an `@Dynamic` annotation. It hints to your IDE that the method missing isn't actually an error.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/ReloadableResourceManagerImplMixin.java
```diff
@@ -0,0 +1,73 @@
...
+	@Inject(method = "method_29491", at = @At("HEAD"), cancellable = true, remap = false)
+	private static void getResourcePackNames(List<ResourcePack> packs, CallbackInfoReturnable<String> cir) {
+		cir.setReturnValue(packs.stream().map(pack -> {
+			if (pack instanceof GroupResourcePack groupResourcePack) {
+				return groupResourcePack.getFullName();
+			} else {
+				return pack.getName();
+			}
+		}).collect(Collectors.joining(", ")));
+	}
```
TheGlitch76, line 82: Hello, unconditional return! This should be an overwrite. If it didn't work before, try it again with the `@Dynamic` annotation I mentioned above.
If even that doesn't fix it, at least mark in a comment why this isn't an overwrite.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/VanillaDataPackProviderMixin.java
```diff
@@ -0,0 +1,49 @@
...
+										 CallbackInfo ci) {
+		// Register built-in resource packs after vanilla built-in resource packs are registered.
+		ModResourcePackProvider.SERVER_RESOURCE_PACK_PROVIDER.register(profileAdder, factory);
+	}
+
+	// Synthetic method register(Consumer;ResourcePackProfile$Factory;)V -> lambda in ResourcePackProfile.of
+	@Inject(method = "method_14454", at = @At("RETURN"), cancellable = true, remap = false)
+	private void onPackGet(CallbackInfoReturnable<ResourcePack> cir) {
+		cir.setReturnValue(ResourceLoaderImpl.buildMinecraftResourcePack((DefaultResourcePack) cir.getReturnValue()));
+	}
```
TheGlitch76: Same as previous comments

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/client/ClientBuiltinResourcePackProviderMixin.java
```diff
@@ -0,0 +1,65 @@
...
+		var originalPack = (AbstractFileResourcePack) cir.getReturnValue();
+		cir.setReturnValue(ResourceLoaderImpl.buildProgrammerArtResourcePack(originalPack));
+	}
+
+	// ClientBuiltinResourcePackProvider#method_25454 second lambda.
+	@Inject(method = "method_25456", at = @At("RETURN"), cancellable = true, remap = false)
+	private static void onSupplyDirProgrammerArtPack(File file, CallbackInfoReturnable<ResourcePack> cir) {
+		var originalPack = (AbstractFileResourcePack) cir.getReturnValue();
+		cir.setReturnValue(ResourceLoaderImpl.buildProgrammerArtResourcePack(originalPack));
+	}
```
TheGlitch76: And these

library/core/resource-loader/src/testmod/java/org/quiltmc/qsl/resource/loader/test/BuiltinResourcePackTestMod.java
```diff
@@ -0,0 +1,37 @@
...
+import org.quiltmc.qsl.resource.loader.api.ResourceLoader;
+import org.quiltmc.qsl.resource.loader.api.ResourcePackActivationType;
+
+public class BuiltinResourcePackTestMod implements ModInitializer {
+	private static final Logger LOGGER = LogManager.getLogger();
+
+	@Override
+	public void onInitialize() {
+		if (!ResourceLoader.registerBuiltinResourcePack(id("test"), ResourcePackActivationType.DEFAULT_ENABLED)) {
+			LOGGER.warn("Could not register built-in resource pack.");
```
TheGlitch76: this should probably fail hard so it's easily caught when someone is naively running all the test mods at once

TheGlitch76 (changes requested):

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/reloader/SimpleResourceReloader.java
LambdAurora: Apply stage always run on the game thread (due to the apply executor), load stage can be asynchronous, so on different threads.
The contract is already specified in the class' javadoc.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/DefaultResourcePackMixin.java
LambdAurora: The bug is mentioned on the class javadoc.

library/core/resource-loader/src/testmod/java/org/quiltmc/qsl/resource/loader/test/BuiltinResourcePackTestMod.java
LambdAurora: Throwing a RuntimeException now.

TheGlitch76 (approved):

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourcePackActivationType.java
i509VCB: Guessing this was a no for the response?

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/api/reloader/IdentifiableResourceReloader.java
```diff
@@ -0,0 +1,50 @@
...
+
+	/**
+	 * Returns the identifiers of resource reloaders this resource reloader expects to have been
+	 * executed before itself.
+	 * <p>
+	 * Please keep in mind that this only takes effect during the application stage!
+	 *
+	 * @return the identifiers this resource reloader depends on
+	 */
+	default Collection<Identifier> getQuiltDependencies() {
```
i509VCB, line 48: Maybe add an `@see` comment referring to the keys?

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ModResourcePackProvider.java
```diff
@@ -0,0 +1,73 @@
...
+	public static final ResourcePackSource PACK_SOURCE_MOD_BUILTIN = ResourcePackSource.nameAndSource("pack.source.builtin");
+	public static final ModResourcePackProvider CLIENT_RESOURCE_PACK_PROVIDER = new ModResourcePackProvider(ResourceType.CLIENT_RESOURCES);
+	public static final ModResourcePackProvider SERVER_RESOURCE_PACK_PROVIDER = new ModResourcePackProvider(ResourceType.SERVER_DATA);
+
+	private final ResourceType type;
+	private final ResourcePackProfile.Factory factory;
+
+	public ModResourcePackProvider(ResourceType type) {
+		this.type = type;
+		this.factory = (name, text, bl, supplier, metadata, initialPosition, source) ->
```
i509VCB: hmm what's `bl`

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ModResourcePackUtil.java
```diff
@@ -0,0 +1,61 @@
...
+		if ("pack.mcmeta".equals(filename)) {
+			String description = info.getName();
+
+			if (description == null) {
+				description = "";
+			} else {
+				description = description.replaceAll("\"", "\\\"");
+			}
+
+			var pack = String.format("""
```
i509VCB, line 45: This formatting is a bit bizzare?

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceLoaderImpl.java
```diff
@@ -0,0 +1,325 @@
...
+			);
+		}
+	}
+
+	/**
+	 * Sorts the given resource reloaders to satisfy dependencies.
+	 *
+	 * @param reloaders the resource reloaders to sort
+	 */
+	private void sort(List<ResourceReloader> reloaders) {
```
i509VCB, line 104: Would be nice to place comments inside the method dictating each step.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceLoaderImpl.java
```diff
@@ -0,0 +1,325 @@
...
+	/**
+	 * Appends mod resource packs to the given list.
+	 *
+	 * @param packs   the resource pack list to append
+	 * @param type    the type of resource
+	 * @param subPath the resource pack sub path directory in mods, may be {@code null}
+	 */
+	public static void appendModResourcePacks(List<ResourcePack> packs, ResourceType type, @Nullable String subPath) {
+		for (var container : FabricLoader.getInstance().getAllMods()) {
+			if (container.getMetadata().getType().equals("builtin")) {
```
i509VCB, line 189: Not sure this api is going to live in loader, I'm guessing you need this anyways.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/DefaultResourcePackMixin.java
```diff
@@ -0,0 +1,108 @@
...
+	 * @author QuiltMC, LambdAurora
+	 * @reason Rewrite default resource access to avoid resource leaking.
+	 */
+	@Overwrite
+	public Collection<Identifier> findResources(ResourceType type, String namespace, String prefix, int maxDepth,
+												Predicate<String> pathFilter) {
+		return this.internalPack.findResources(type, namespace, prefix, maxDepth, pathFilter);
+	}
+
+	@Inject(method = "close", at = @At("HEAD"), remap = false)
```
i509VCB, line 104: Hmm we need to test this on a real client, remap=false raises an eye brow

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/client/GameOptionsMixin.java
```diff
@@ -0,0 +1,91 @@
...
+import org.quiltmc.qsl.resource.loader.impl.ModResourcePackProvider;
+
+@Mixin(GameOptions.class)
+public abstract class GameOptionsMixin {
+	@Shadow
+	public List<String> resourcePacks;
+
+	@Shadow
+	private static List<String> parseList(String content) {
+		throw new UnsupportedOperationException("Injection failed.");
```
i509VCB: I think we want to use assertion error in these cases? Unless there is a more appropriate type than that.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ModResourcePackUtil.java
LambdAurora: How should this be formatted then?
A text block expression requires the string to begin on a new line.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/DefaultResourcePackMixin.java
LambdAurora: This is `Closeable#close`, so no remapping.

build-logic/src/main/groovy/qsl.module.gradle
i509VCB: Crash reports are not going to build without this block, so I'm going to copy it

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/client/GameOptionsMixin.java
LambdAurora: To me a `IllegalStateException` would be more appropriate:

> Signals that a method has been invoked at an illegal or inappropriate time. In other words, the Java environment or Java application is not in an appropriate state for the requested operation.

i509VCB: Yeah that makes sense, we probably need a document noting these little conventions that aren't strictly within scope of code style

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/ReloadableResourceManagerImplMixin.java
LambdAurora: Turns out `@Dynamic` doesn't work, neither does remap = false on the `@Overwrite`.
Which is definitely very weird.

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceLoaderImpl.java
LambdAurora: Done.

i509VCB (approved):
Have tested this on macos so I can assure it works there

PR #18, opened 2021-07-26
Item Setting API module

OroArmor:
This pull request creates the Quilt Items Library.

What is done:
- [x] Item Settings
- [ ] ~~Item Tooltips~~ Part of Tooltip API
- [ ] ~~Fuel Registry~~ Moving to Item Data PR
- [ ] ~~Composting Registry~~ Moving to Item Data PR
- [ ] ~~Tool Attributes~~ - Delayed until later, lets get some apis in
- [x] ~~Item Groups~~ removed

What has/can be added from Fabric PRs:
 - [x] Custom Item Setting API (HavenKing)
 - [x] ~~Recipe Remainer API (BoogieMonster1O1)~~ Moving to follow up PR, slightly complex system, but not horrible
 - [ ] ~~Item Extensions - Elytra, Bow, Crossbow (OroArmor)~~ Separate PR


library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/impl/QuiltCreativeGuiComponents.java
```diff
@@ -0,0 +1,89 @@
...
+
+		public ItemGroupButtonWidget(int x, int y, Type type, CreativeGuiExtensions extensions) {
+			super(x, y, 11, 10, type.text, (bw) -> type.clickConsumer.accept(extensions));
+			this.extensions = extensions;
+			this.type = type;
+			this.gui = (CreativeInventoryScreen) extensions;
+		}
+
+		@Override
+		public void render(MatrixStack matrixStack, int mouseX, int mouseY, float float_1) {
```
LambdAurora: What happened to those params?
```suggestion
		public void render(MatrixStack matrices, int mouseX, int mouseY, float delta) {
```

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/api/QuiltItemGroupBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.function.Supplier;
+
+import net.minecraft.item.ItemGroup;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.collection.DefaultedList;
+import net.minecraft.util.Identifier;
+
+import org.quiltmc.qsl.itemgroup.impl.ItemGroupExtensions;
+
+public final class QuiltItemGroupBuilder {
```
LambdAurora: Class javadoc?

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/impl/ItemGroupExtensions.java
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.quiltmc.qsl.itemgroup.impl;
+
+public interface ItemGroupExtensions {
+	void expandArray();
```
LambdAurora: Shouldn't those be prefixed with `quilt$` to avoid possible conflicts?

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/mixin/client/CreativePlayerInventoryGuiMixin.java
```diff
@@ -0,0 +1,166 @@
...
+import net.minecraft.entity.player.PlayerInventory;
+import net.minecraft.item.ItemGroup;
+import net.minecraft.text.Text;
+
+import org.quiltmc.qsl.itemgroup.impl.CreativeGuiExtensions;
+import org.quiltmc.qsl.itemgroup.impl.QuiltCreativeGuiComponents;
+
+@Mixin(CreativeInventoryScreen.class)
+public abstract class CreativePlayerInventoryGuiMixin extends AbstractInventoryScreen<CreativeInventoryScreen.CreativeScreenHandler> implements CreativeGuiExtensions {
+	public CreativePlayerInventoryGuiMixin(CreativeInventoryScreen.CreativeScreenHandler screenHandler_1, PlayerInventory playerInventory_1, Text textComponent_1) {
```
LambdAurora: This is really weird.
```suggestion
	public CreativePlayerInventoryGuiMixin(CreativeInventoryScreen.CreativeScreenHandler screenHandler, PlayerInventory playerInventory, Text title) {
```

library/items/item/src/main/java/org/quiltmc/qsl/item/impl/AppliedItemSettingHooks.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package org.quiltmc.qsl.item.impl;
+
+import org.quiltmc.qsl.item.api.item.v1.CustomDamageHandler;
+import org.quiltmc.qsl.item.api.item.v1.EquipmentSlotProvider;
+
+public interface AppliedItemSettingHooks {
+	/* @Nullable */ EquipmentSlotProvider getEquipmentSlotProvider();
```
LambdAurora: This annotation being commented looks wrong to me.

library/items/item/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,11 @@
+{
+  "schemaVersion": 1,
+  "id": "qsl-items-item",
```
LambdAurora: so we do kebab-case modid?
I still find it weird, especially with how in Identifiers the path is snake_case (in Vanilla at least), which hints toward that as the standard.

library/items/item/src/testmod/java/org/quiltmc/qsl/item/test/CustomDamageTest.java
```diff
@@ -0,0 +1,59 @@
...
+	public void onInitialize() {
+		Registry.register(Registry.ITEM, new Identifier("qsl-items-item", "weird_pickaxe"), new WeirdPick());
+	}
+
+	public static final CustomDamageHandler WEIRD_DAMAGE_HANDLER = (stack, amount, entity, breakCallback) -> {
+		// If sneaking, apply all damage to vanilla. Otherwise, increment a tag on the stack by one and don't apply any damage
+		if (entity.isSneaking()) {
+			return amount;
+		} else {
+			NbtCompound tag = stack.getOrCreateNbt();
```
LambdAurora: Nitpick, but `tag` name is kinda wrong here now.

LambdAurora (commented):
I also wonder if we should mandate the use of `this` when calling fields or methods as it really helps in readability and it tells directly where it comes from and what scope it affects.

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/mixin/ItemGroupMixin.java
```diff
@@ -0,0 +1,42 @@
...
+
+@Mixin(ItemGroup.class)
+public abstract class ItemGroupMixin implements ItemGroupExtensions {
+	@Shadow
+	@Final
+	@Mutable
+	public static ItemGroup[] GROUPS;
+
+	@Override
+	public void expandArray() {
```
pluiedev: I think that we should have a registry for item groups, instead of growing the item group array each time an item group is created. That way we can have all of the item groups in one place, and register them all at once in one go.
Also I feel that registring an item group *implicitly* when it's created is a bit unintuitive and non-conforming when most of the rest of the API requires explicit registration.

library/items/item/src/main/java/org/quiltmc/qsl/item/impl/AppliedItemSettingHooks.java
pluiedev: Probs some old FAPI relic where nullable annotations aren't allowed for some reason

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/mixin/ItemGroupMixin.java
Pyrofab: I believe the idea with modifying the array each time is so that mods can inspect all the tabs as they would in vanilla, or even add tabs themselves without the standard library.

pluiedev: >so that mods can inspect all the tabs as they would in vanilla

My idea is to have the QSL expand and populate the array in bulk and on demand. Maybe we could update the vanilla "mirror" every time it's accessed and the QSL registry changes, and that way it syncs up.

>or even add tabs themselves without the standard library.

Sure, modders can just use mixins to expand the array this way, but the standard library need not to do that - I would expect that most modders would use the QSL to solve problems that have already been solved before (i.e. not reinventing the wheel), so IMO should a mod reimplement something that already exists in the QSL, it's entirely on its own to be compatible.

tl;dr a mod can do whatever it likes, but it's not the QSL's problem if they are incompatible

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/api/QuiltItemGroupBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+		return new ItemGroup(ItemGroup.GROUPS.length - 1, String.format("%s.%s", identifier.getNamespace(), identifier.getPath())) {
+			@Override
+			public ItemStack createIcon() {
+				return stackSupplier.get();
+			}
+
+			@Override
+			public void appendStacks(DefaultedList<ItemStack> stacks) {
+				if (stacksForDisplay != null) {
+					stacksForDisplay.accept(stacks);
```
i509VCB: We probably need to figure out a better way to handle appending of extra item stacks. At the moment setting the supplier on the builder removes the original contents of the item group.

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/api/QuiltItemGroupBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	/**
+	 * This is a single method that makes creating an {@link ItemGroup} with an icon one call.
+	 *
+	 * @param identifier    the id will become the name of the {@link ItemGroup} and will be used for the translation key
+	 * @param stackSupplier the supplier should return the {@link ItemStack} that you wish to show on the tab
+	 * @return An instance of the built {@link ItemGroup}
+	 */
+	public static ItemGroup build(Identifier identifier, Supplier<ItemStack> stackSupplier) {
```
i509VCB: This method always felt weird being named build and being static.

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/mixin/ItemGroupMixin.java
i509VCB: A registry implies removals are also valid, so any implementation would need to take that into account.

i509VCB (dismissed):

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/api/QuiltItemGroupBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+	/**
+	 * This is used to add an icon to the {@link ItemGroup}.
+	 *
+	 * @param stackSupplier the supplier should return the {@link ItemStack} that you wish to show on the tab
+	 * @return this
+	 */
+	public QuiltItemGroupBuilder icon(Supplier<ItemStack> stackSupplier) {
+		this.stackSupplier = stackSupplier;
+		return this;
+	}
```
Haven-King: I'd like to see an analogue to this method that just takes a single `ItemStack`. There are some issues with forward references, but they can be mitigated by some restructuring of the rest of the item API that I'll try and detail in some of my other comments.

Additionally, a potential change that would require some more mixins would be to allow for multiple stacks, stack suppliers, and item tags to be provided to enable animated tab group icons that switch between all provided icons. This doesn't necessarily need to be added for the initial port, but is something that's been on my mind.

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/api/QuiltItemGroupBuilder.java
Haven-King: A restructure of this part of the API that makes sense to me would be to remove the `public ItemGroup build()` method that is currently called before item registration and instead expose a `public void register()` method that would be called *after* all items are registered to it. Ideally the `register` method could be omitted entirely, and Quilt could just register all item groups at once, but there is not currently a defined place where that would be done.

The `QuiltItemSettings#group` method would need to be changed to take a builder, and items would need to have their groups assigned lazily once the builder is built, but this would make a lot of things much cleaner. This will allow stacks to be appended programmatically in mod initializers.

Mods might around and expose their item group builders:
```java
public class MegaMod implements ModInitializer {
  public static final QuiltItemGroupBuilder ITEMS = QuiltItemGroupBuilder.create(new Identifier("megamod:items"));

  public static Item MEGA_INGOT, MEGA_WAND, MEGA_SWORD, MEGA_BELT;

  @Override
  public void onInitialize() {
    MEGA_INGOT = new Item(new QuiltItemSettings().group(ITEMS));
    MEGA_WAND = new Item(new QuiltItemSettings().group(ITEMS));
    MEGA_SWORD = new Item(new QuiltItemSettings().group(ITEMS));
    MEGA_BELT = new Item(new QuiltItemSettings().group(ITEMS));

    // Ideally the call to register here will not be necessary
    ITEMS.icon(MEGA_INGOT).register();
  }
}
```

Another mod might add extensions to that mod, and can easily add its stacks with
```java
public class MegaModExtensions implements ModInitializer {
  public static Item MEGA_HELMET, MEGA_CHESTPLATE, MEGA_LEGGINGS, MEGA_BOOTS;

  @Override
  public void onInitialize() {
    MEGA_HELMET = new ArmorItem(MegaModArmor.MATERIAL, EquipmentSlot.HEAD, new QuiltItemSettings().group(MegaMod.ITEMS));
    MEGA_CHESTPLATE= new ArmorItem(MegaModArmor.MATERIAL, EquipmentSlot.CHEST, new QuiltItemSettings().group(MegaMod.ITEMS));
    MEGA_LEGGINGS= new ArmorItem(MegaModArmor.MATERIAL, EquipmentSlot.LEGS, new MEGA_LEGGINGS().group(MegaMod.ITEMS));
    MEGA_BOOTS= new ArmorItem(MegaModArmor.MATERIAL, EquipmentSlot.FEET, new QuiltItemSettings().group(MegaMod.ITEMS));

    // If we can find a good place to "freeze" the registries, this call can be omitted,
    // making the API even cleaner
    MegaMod.ITEMS.register();
  }
}
```

These changes aren't necessary for this PR to be merged, but should be seriously considered prior to release. Implementation for all of these changes would be fairly easy after we've hashed out the details of the API.

library/items/item/src/main/java/org/quiltmc/qsl/item/api/item/v1/QuiltItemSettings.java
```diff
@@ -0,0 +1,107 @@
...
+import org.quiltmc.qsl.item.impl.QuiltItemInternals;
+
+/**
+ * Quilts's version of {@link Item.Settings}. Adds additional methods and hooks
+ * not found in the original class.
+ *
+ * <p>To use it, simply replace {@code new Item.Settings()} with
+ * {@code new QuiltItemSettings()}.
+ */
+public class QuiltItemSettings extends Item.Settings {
```
Haven-King: This would probably be as good a time as any to add support for [custom item settings](https://github.com/FabricMC/fabric/pull/1359).

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/mixin/ItemGroupMixin.java
Haven-King: A full-on registry is likely unnecessary. Lazy initialization of item groups should be sufficient.

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/impl/ItemGroupExtensions.java
FoundationGames: <https://github.com/QuiltMC/quilt-standard-libraries/pull/19> uses `qsl$`

OroArmor: Is that still needed? I thought that mixin fixed that issue. Or does it also help with crash logs

BasiqueEvangelist: > I thought that mixin fixed that issue

This isn't an invoker, though?

library/items/item/src/main/resources/fabric.mod.json
Haven-King: Per the [quilt.mod.json](https://github.com/QuiltMC/rfcs/blob/master/loader/0002-quilt.mod.json.md#the-id-field) spec, snake_case is best practice.

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/api/QuiltItemGroupBuilder.java
pluiedev: Yeah that's pretty much what I mean, although if we can find the place to "freeze" the registries then it would be better

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/mixin/client/CreativePlayerInventoryGuiMixin.java
OroArmor: this is how old the item group API is

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/api/QuiltItemGroupBuilder.java
OroArmor: `build` -> `create`

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/api/QuiltItemGroupBuilder.java
OroArmor: Yeah, im mostly focused on ported the current apis for a main review. If there is something big enough that we want to change, that can be done later in this PR or in a different one

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/api/QuiltItemGroupBuilder.java
OroArmor: I have a couple issues with the above idea. Mostly that there would then be a dependency on the item-group API from the item API. As well, I think that supplier `ItemGroup`s are fine and make it clearer that stuff needs to be "lazy". Im not sure about the best way to set the icon. But a subclass of item group that allows for setting the icon directly might be better. (Im going to implement this now)

OroArmor: @Haven-King take a look at the javadocs and the expanded test mod. I think it balances your suggestion along with making it simple for users and keeping the API implementation simple

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/api/QuiltItemGroup.java
```diff
@@ -0,0 +1,61 @@
...
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemGroup;
+import net.minecraft.item.ItemStack;
+
+/**
+ * Extensions for the {@link ItemGroup} class. Currently, the only extension is setting the icon with an {@link Item} after the item has been created ({@link QuiltItemGroup#setIcon(Item)}).
+ */
+public class QuiltItemGroup extends ItemGroup {
```
Haven-King: Having an extended `ItemGroup` is good, but it would make more sense to me for it to be an interface implemented on the standard `ItemGroup` class so that every `ItemGroup` is a `QuiltItemGroup` since the behavior is purely additive, similar to how every `BakedModel` is a `FabricBakedModel`.

library/items/item/src/main/java/org/quiltmc/qsl/item/api/item/v1/QuiltItemSettings.java
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	@Override
+	public QuiltItemSettings maxDamage(int maxDamage) {
+		super.maxDamage(maxDamage);
+		return this;
+	}
+
+	@Override
+	public QuiltItemSettings recipeRemainder(Item recipeRemainder) {
```
LemmaEOF: I think it would be really good for Quilt to add a `selfRemainder` method, since by nature of how this method works it's kind of impossible to have, say, a philosopher's stone not get used up when crafting, since it doesn't exist as an item yet when QuiltItemSettings is being created.

OroArmor: Yeah, iirc there was a much more in depth remainder API that someone made. i will look into port that one

LemmaEOF: I drafted something for Woven a lil while back, but it was kinda weird.

Platymemo:
So are the [custom item settings](https://github.com/FabricMC/fabric/pull/1359) and the [added recipe remainder settings](https://github.com/FabricMC/fabric/pull/1180) planned to be added to this PR, or PR'd after this one is merged? I'm of the mind that the priority is the base item library, and these others can be PR'd in later. The same thing with the lazy item group initialization, priority should be finishing and merging this current PR, then implementing these changes later.

OroArmor:
I think the importance along with pretty much an ok for those other apis in fabric (but with some bikeshed hell) means that they can be merged in. Most of the changes are pretty small or refactors to the api that need to happen anyway, so the impact is minimal on porting time

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/mixin/ItemGroupMixin.java
OroArmor: Yeah, dynamic removal of item groups is likely unnessicary

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/api/QuiltItemGroupBuilder.java
OroArmor: @i509VCB I made it so that it no longer returns, allowing for items to be added and item stacks.

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/api/QuiltItemGroup.java
```diff
@@ -0,0 +1,70 @@
...
+	protected QuiltItemGroup(int index, String id, @Nullable Supplier<ItemStack> iconSupplier) {
+		super(index, id);
+		this.iconSupplier = iconSupplier;
+	}
+
+	/**
+	 * Sets the {@link Item} to use as the icon for the {@link ItemGroup}
+	 * @param icon The {@link Item} for the icon
+	 */
+	public void setIcon(@NotNull ItemConvertible icon) {
```
LambdAurora: Those `NotNull`s shouldn't be here. Only nullable stuff should be annotated.

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/mixin/client/CreativePlayerInventoryGuiMixin.java
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	@Shadow
+	protected abstract void setSelectedTab(ItemGroup itemGroup_1);
+
+	@Shadow
+	public abstract int getSelectedTab(); /* XXX getSelectedTab XXX */
+
+	// "static" matches selectedTab
+	private static int qsl$currentPage = 0;
```
LambdAurora:
```suggestion
	@Unique
	private static int qsl$currentPage = 0;
```

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/mixin/client/CreativePlayerInventoryGuiMixin.java
```diff
@@ -0,0 +1,166 @@
...
+
+	private int getPageOffset(int page) {
```
LambdAurora:
```suggestion
	@Unique
	private int getPageOffset(int page) {
```

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/mixin/client/CreativePlayerInventoryGuiMixin.java
```diff
@@ -0,0 +1,166 @@
...
+		return switch (page) {
+			case 0 -> 0;
+			case 1 -> 12;
+			default -> 12 + ((12 - QuiltCreativeGuiComponents.ALWAYS_SHOWN_GROUPS.size()) * (page - 1));
+		};
+	}
+
+	private int getOffsetPage(int offset) {
```
LambdAurora:
```suggestion
	@Unique
	private int getOffsetPage(int offset) {
```

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/mixin/client/CreativePlayerInventoryGuiMixin.java
```diff
@@ -0,0 +1,166 @@
...
+		}
+
+		if (type == QuiltCreativeGuiComponents.Type.PREVIOUS) {
+			return qsl$currentPage != 0;
+		}
+
+		return false;
+	}
+
+	private void qsl$updateSelection() {
```
LambdAurora:
```suggestion
	@Unique
	private void qsl$updateSelection() {
```

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/mixin/client/CreativePlayerInventoryGuiMixin.java
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	@Inject(method = "renderTabIcon", at = @At("HEAD"), cancellable = true)
+	private void renderTabIcon(MatrixStack matrixStack, ItemGroup itemGroup, CallbackInfo info) {
+		if (qsl$isGroupNotVisible(itemGroup)) {
+			info.cancel();
+		}
+	}
+
+	private boolean qsl$isGroupNotVisible(ItemGroup itemGroup) {
```
LambdAurora:
```suggestion
	@Unique
	private boolean qsl$isGroupNotVisible(ItemGroup itemGroup) {
```

library/items/item/src/testmod/java/org/quiltmc/qsl/item/test/QuiltItemSettingsTests.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.concurrent.ThreadLocalRandom;
+
+public class QuiltItemSettingsTests implements ModInitializer {
+	public static final CustomItemSetting<String> CUSTOM_DATA_TEST = CustomItemSetting.create(() -> null);
+
+	@Override
+	public void onInitialize() {
+		// Registers an item with a custom equipment slot.
+		Item testItem = new Item(new QuiltItemSettings().group(ItemGroup.MISC).equipmentSlot(stack -> EquipmentSlot.CHEST));
+		Registry.register(Registry.ITEM, new Identifier("qsl_items_item_testmod", "test_item"), testItem);
```
LambdAurora: imo, the namespace should be in a static final field for cleanliness.

library/items/item/src/main/java/org/quiltmc/qsl/item/mixin/AbstractFurnaceBlockEntityMixin.java
```diff
@@ -0,0 +1,73 @@
...
+		super(type, pos, state);
+	}
+
+	@SuppressWarnings("ConstantConditions")
+	@Inject(method = "tick", at = @At(value = "INVOKE", target = "Lnet/minecraft/item/Item;getRecipeRemainder()Lnet/minecraft/item/Item;"), locals = LocalCapture.CAPTURE_FAILHARD)
+	private static void setRemainder(World world, BlockPos pos, BlockState state, AbstractFurnaceBlockEntity blockEntity, CallbackInfo ci, boolean bl, boolean bl2, ItemStack itemStack, Recipe<?> recipe, int i, Item item) {
+		ItemStack stack = CustomItemSettingImpl.RECIPE_REMAINDER_PROVIDER.get(item).getRecipeRemainder(
+				itemStack,
+				blockEntity,
+				((AbstractFurnaceBlockEntityMixin) (Object) blockEntity).recipeType,
```
LambdAurora: I would use an intermediate local just for readability imo.

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/api/QuiltItemGroup.java
OroArmor:
```suggestion
	public void setIcon(ItemConvertible icon) {
```

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/api/QuiltItemGroup.java
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	@Override
+	public ItemStack createIcon() {
+		if (iconSupplier != null && !iconSupplier.get().isEmpty()) {
+			return iconSupplier.get();
+		} else {
+			if (icon == null) {
+				return ItemStack.EMPTY;
+			}
```
BasiqueEvangelist: This should probably just throw an exception if there is no item. There aren't that many cases in which a mod will want the group to not have an item, and in even that case it would be better (imo) to explicitly specify that the group doesn't have an item.

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/mixin/ItemGroupMixin.java
```diff
@@ -0,0 +1,42 @@
...
+	@Mutable
+	public static ItemGroup[] GROUPS;
+
+	@Override
+	public void qsl$expandArray() {
+		ItemGroup[] tempGroups = GROUPS;
+		GROUPS = new ItemGroup[GROUPS.length + 1];
+
+		System.arraycopy(tempGroups, 0, GROUPS, 0, tempGroups.length);
+	}
```
BasiqueEvangelist: Why is this a duck method that only accesses static fields? Wouldn't just making an accessor for `GROUPS` be much simpler?

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/mixin/client/CreativePlayerInventoryGuiMixin.java
```diff
@@ -0,0 +1,171 @@
...
+public abstract class CreativePlayerInventoryGuiMixin extends AbstractInventoryScreen<CreativeInventoryScreen.CreativeScreenHandler> implements CreativeGuiExtensions {
+	public CreativePlayerInventoryGuiMixin(CreativeInventoryScreen.CreativeScreenHandler screenHandler, PlayerInventory playerInventory, Text title) {
+		super(screenHandler, playerInventory, title);
+	}
+
+	@Shadow
+	protected abstract void setSelectedTab(ItemGroup itemGroup_1);
+
+	@Shadow
+	public abstract int getSelectedTab(); /* XXX getSelectedTab XXX */
```
i509VCB: Why the comment at the end?

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/mixin/client/CreativePlayerInventoryGuiMixin.java
```diff
@@ -0,0 +1,171 @@
...
+
+	// "static" matches selectedTab
+	@Unique
+	private static int qsl$currentPage = 0;
```
i509VCB: Probably could use better documentation, something like

"In order to match the behavior of Vanilla where closing and opening the creative inventory brings you to the previously open item group, we also replicate this behavior with the current page"

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/api/QuiltItemGroupBuilder.java
```diff
@@ -0,0 +1,147 @@
...
+	private Supplier<ItemStack> stackSupplier = () -> ItemStack.EMPTY;
+	private Consumer<List<ItemStack>> stacksForDisplay;
+
+	/**
+	 * Create a new {@link QuiltItemGroupBuilder}. Using the constructor allows for the use of the {@link QuiltItemGroupBuilder#icon(Supplier)} and {@link QuiltItemGroupBuilder#appendItems(Consumer)} methods.
+	 * Manually setting the icon with {@link QuiltItemGroup#setIcon(Item)} is possible after calling {@link QuiltItemGroupBuilder#build()}
+	 *
+	 * @param identifier the {@link Identifier} will become the name of the {@link ItemGroup} and will be used for the translation key
+	 */
+	public QuiltItemGroupBuilder(Identifier identifier) {
```
i509VCB: This constructor should probably be private given the class is final and we have factory methods

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/api/QuiltItemGroup.java
```diff
@@ -0,0 +1,70 @@
...
+
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemConvertible;
+import net.minecraft.item.ItemGroup;
+import net.minecraft.item.ItemStack;
+
+/**
+ * Extensions for the {@link ItemGroup} class. Currently, the only extension is setting the icon with either an {@link ItemConvertible} or {@link ItemStack} after the item has been created ({@link QuiltItemGroup#setIcon(ItemConvertible)}, {@link QuiltItemGroup#setIcon(ItemStack)}).
+ */
+public class QuiltItemGroup extends ItemGroup {
```
i509VCB: Is this intended for extension?

Based on the builder impl I'd say no?

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/api/QuiltItemGroupBuilder.java
```diff
@@ -0,0 +1,147 @@
...
+ *   public static Item MY_ITEM;
+ *
+ *   @Override
+ *   public void onInitialize() {
+ *     MY_ITEM = new Item(new QuiltItemSettings().group(MY_ITEM_GROUP));
+ *   }
+ * }
+ * }</pre>
+ */
+public final class QuiltItemGroupBuilder {
```
i509VCB: Do we want to move this to be a subclass of `QuiltItemGroup`?

i509VCB:
Hmm Oro, we may want to consider splitting this big library into multiple pull requests, I imagine that would be easier to get through and review in 1 sitting?

OroArmor:
I guess, but right now there isn't too much. I'll break off the item group api into its own as there is more work for the item setting api that need to be done

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/api/QuiltItemGroup.java
OroArmor: it shouldnt be, but there could be a case where a modder wishes to extend it, though that should be done from `ItemGroup` in most cases

OroArmor:
Turning `ItemGroup` into its own pull request.

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/mixin/ItemGroupMixin.java
OroArmor: see #40 for item group api

library/items/item-group/src/main/java/org/quiltmc/qsl/itemgroup/api/QuiltItemGroupBuilder.java
OroArmor: see #40 for api implementation

OroArmor:
I think this is ready for review. I have scaled back the scope of the PR so that it is easier to merge in.

library/items/build.gradle
```diff
@@ -0,0 +1,8 @@
+plugins {
+	id("qsl.library")
+}
+
+qslLibrary {
+	libraryName = "items"
```
Leo40Git: Pretty sure this is supposed to be `item`, not `items`.

OroArmor: :blob_waaaaa:

OroArmor: I think i chose it because the items library had the item api

TheGlitch76:
Please resolve merge conflicts!

OroArmor:
Fixed!

library/items/build.gradle
TheGlitch76: Per the RFC this needs to be `item`

OroArmor: :pain: give me a minute

library/item/item/build.gradle
```diff
@@ -0,0 +1,12 @@
+plugins {
+	id("qsl.module")
+}
+
+qslModule {
+	moduleName = "item"
+	version = "1.0.0"
+	library = "items"
```
LambdAurora:
```suggestion
	library = "item"
```

library/item/item/src/main/java/org/quiltmc/qsl/item/api/item/v1/CustomItemSetting.java
```diff
@@ -0,0 +1,86 @@
...
+ * <h2>Usage Example</h2>
+ * <pre>{@code
+ * // Create the setting instance. You can think of this as the "setting key".
+ * public static final CustomItemSetting<String> CUSTOM_TOOLTIP = CustomItemSetting.create(() -> null);
+ *
+ * @Override
+ * public void onInitializeClient() {
+ *     ItemTooltipCallback.EVENT.register((stack, context, lines) -> {
+ *         // Gets the setting from the specified item.
+ *         String tooltip = CUSTOM_TOOLTIP.getValue(stack.getItem());
```
IotaBread: This seems outdated, `getValue` doesn't exist

library/item/item/src/main/java/org/quiltmc/qsl/item/api/item/v1/QuiltItemSettings.java
```diff
@@ -0,0 +1,179 @@
...
+
+	/**
+	 * Sets a custom setting of the item.
+	 * @param setting the unique type for this setting
+	 * @param value the object containing the setting itself
+	 * @return this builder
+	 */
+	public <T> QuiltItemSettings customSetting(CustomItemSetting<T> setting, T value) {
+		if (!(setting instanceof CustomItemSettingImpl)) {
+			throw new UnsupportedOperationException("CustomItemSetting should not be ");
```
IotaBread: Shouldn't the class name be appended here?

library/item/item/src/main/java/org/quiltmc/qsl/item/api/client/item/v1/ItemTooltipCallback.java
```diff
@@ -0,0 +1,53 @@
...
+
+	/**
+	 * Called when an item stack's tooltip is rendered. Text added to {@code lines} will be
+	 * rendered with the tooltip. Values from {@code lines} can also be modified or removed.
+	 *
+	 * @param stack The {@link ItemStack} to build the tooltip for
+	 * @param context The {@link TooltipContext} for the current Tooltip being built
+	 * @param lines the list containing the lines of text displayed on the stack's tooltip
+	 */
+	void tooltipCallback(ItemStack stack, TooltipContext context, List<Text> lines);
```
williambl: Will this name be changed? I don't think it fits with most other event callback method names, which tend to look more like `onX` than `xCallback`.

library/item/item/src/main/java/org/quiltmc/qsl/item/api/client/item/v1/ItemTooltipCallback.java
```diff
@@ -0,0 +1,53 @@
...
+import net.minecraft.item.ItemStack;
+import net.minecraft.text.Text;
+
+import net.fabricmc.api.EnvType;
+import net.fabricmc.api.Environment;
+
+/**
+ * A callback for modifying the lines of text for the rendered tooltip displayed when the mouse hovers over an item stack.
+ */
+@Environment(EnvType.CLIENT)
```
williambl:
```suggestion
@Environment(EnvType.CLIENT)
@FunctionalInterface
```
This looks like it should be	 annotated with FunctionalInterface?

library/item/item/src/main/java/org/quiltmc/qsl/item/api/item/v1/CustomItemSetting.java
```diff
@@ -0,0 +1,86 @@
...
+ *     }
+ * }}</pre>
+ *
+ * <p>You should not implement this interface.
+ *
+ * <p>Use {@link CustomItemSetting#create(Supplier)} to retrieve an instance of QSL's implementation.
+ *
+ * @param <T> the type of the setting to be attached
+ */
+public interface CustomItemSetting<T> {
```
williambl: Maybe `@ApiStatus.NonExtendable`?

library/item/item/src/main/java/org/quiltmc/qsl/item/api/item/v1/CustomItemSetting.java
OroArmor: `get`

library/item/item/src/main/java/org/quiltmc/qsl/item/api/item/setting/CustomDamageHandler.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * <p>Custom damage handlers can be set with {@link QuiltItemSettings#customDamage}.
+ */
+@FunctionalInterface
+public interface CustomDamageHandler {
+	/**
+	 * Called to apply damage to the given stack.
+	 * This can be used to e.g. drain from a battery before actually damaging the item.
+	 *
+	 * @param stack The {@link ItemStack} that is being damaged
```
LambdAurora: PR-wide comment: `@param`/`@return`/`@throws`/etc should not have the first word capitalized (in the conventions).

OroArmor: ill fix with my next commit

library/item/item_setting/src/main/java/org/quiltmc/qsl/itemsetting/api/CustomItemSetting.java
```diff
@@ -0,0 +1,88 @@
...
+
+import java.util.function.Supplier;
+
+import net.minecraft.item.Item;
+
+import org.jetbrains.annotations.ApiStatus;
+import org.quiltmc.qsl.itemsetting.impl.CustomItemSettingImpl;
+
+/**
+ * A type of setting that can be passed to item constructors.
```
TheGlitch76: what does "can be passed to item constructors" mean? it's not clear from this class alone.

library/item/item_setting/src/main/java/org/quiltmc/qsl/itemsetting/api/CustomItemSetting.java
```diff
@@ -0,0 +1,88 @@
...
+ *     }
+ * }}</pre>
+ *
+ * <p>You should not implement this interface.
+ *
+ * <p>Use {@link CustomItemSetting#create(Supplier)} to retrieve an instance of QSL's implementation.
+ *
+ * @param <T> the type of the setting to be attached
+ */
+@ApiStatus.Internal
```
TheGlitch76: Definitely the wrong annotation. I think you're looking for `ApiStatus.NonExtendable`?

library/item/item_setting/src/main/java/org/quiltmc/qsl/itemsetting/api/EquipmentSlotProvider.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * A provider for the preferred {@link EquipmentSlot} of an item.
+ * This can be used to give non-armor items, such as blocks,
+ * an armor slot that they can go in.
+ *
+ * <p>{@link EquipmentSlotProvider} can be set with {@link QuiltItemSettings#equipmentSlot(EquipmentSlotProvider)}.
+ *
+ * <p>Note that items extending {@link net.minecraft.item.ArmorItem} don't need to use this
+ * as there's {@link net.minecraft.item.ArmorItem#getSlotType()}.
```
TheGlitch76:
```suggestion
 * <p>Note that items extending {@link net.minecraft.item.ArmorItem} should use
 * {@link net.minecraft.item.ArmorItem#getSlotType()} instead.
```

library/item/item_setting/src/main/java/org/quiltmc/qsl/itemsetting/api/EquipmentSlotProvider.java
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+package org.quiltmc.qsl.itemsetting.api;
+
+import net.minecraft.entity.EquipmentSlot;
+import net.minecraft.item.ItemStack;
+
+/**
+ * A provider for the preferred {@link EquipmentSlot} of an item.
```
TheGlitch76: What does  "preferred" mean in this context? Does it mean that an EquipmentSlotProvider can be put in any slot, but just prefers this one?

library/item/item_setting/src/main/java/org/quiltmc/qsl/itemsetting/api/package-info.java
```diff
@@ -0,0 +1,17 @@
+/**
+ * <h2>The Quilt Item Setting sub API.</h2>
```
TheGlitch76: what does sub api mean?

TheGlitch76:
```suggestion
 * <h2>The Quilt Item Settings sub API.</h2>
```

TheGlitch76 (changes requested):
javadoc comments, as usual.
Only got through the API classes so far.

library/item/item_setting/src/main/java/org/quiltmc/qsl/itemsetting/api/package-info.java
OroArmor: Mostly that its a part of the larger item API. This should be changed as I moved this into its own module

TheGlitch76 (approved):

Haven-King (approved):

OroArmor:
Moving this pr into FCP

library/item/item_setting/src/testmod/java/org/quiltmc/qsl/item/test/CustomDamageTest.java
```diff
@@ -0,0 +1,59 @@
...
+import net.minecraft.nbt.NbtCompound;
+import net.minecraft.text.Text;
+import net.minecraft.util.Identifier;
+import net.minecraft.util.registry.Registry;
+
+import net.fabricmc.api.ModInitializer;
+import org.quiltmc.qsl.itemsetting.api.CustomDamageHandler;
+import org.quiltmc.qsl.itemsetting.api.QuiltItemSettings;
+
+public class CustomDamageTest implements ModInitializer {
```
LambdAurora, line 33: Please use QSL's entrypoint (1.18.2).

library/item/item_setting/src/testmod/java/org/quiltmc/qsl/item/test/QuiltItemSettingsTests.java
```diff
@@ -0,0 +1,44 @@
...
+
+import net.minecraft.entity.EquipmentSlot;
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemGroup;
+import net.minecraft.util.Identifier;
+import net.minecraft.util.registry.Registry;
+
+import net.fabricmc.api.ModInitializer;
+
+public class QuiltItemSettingsTests implements ModInitializer {
```
LambdAurora, line 31: Please use QSL's entrypoint (1.18.2).

library/item/item_setting/src/main/java/org/quiltmc/qsl/itemsetting/api/CustomItemSetting.java
```diff
@@ -0,0 +1,88 @@
...
+ * and users that wish to expose a custom setting for use in other mods should do so by exposing
+ * the CustomItemSetting instance.
+ *
+ * <h2>Usage Example</h2>
+ * <pre>{@code
+ * // Create the setting instance. You can think of this as the "setting key".
+ * public static final CustomItemSetting<String> CUSTOM_TOOLTIP = CustomItemSetting.create(() -> null);
+ *
+ * @Override
+ * public void onInitializeClient() {
```
LambdAurora:
```suggestion
 * public void onInitializeClient(ModContainer mod) {
```

QSL entrypoint added a new parameter.

LambdAurora (changes requested):
The package really should be `item.setting` instead of `itemsetting`.
Sorry for pointing this out this late.

LambdAurora (approved):

PR #19, opened 2021-07-26
Tool Interaction Recipes API module

Leo40Git:
This PR creates the Tool Interaction Recipes API module (and also its containing blocks library).

**TODO:**
- [x] Code-based path via registries
- [ ] Data-driven path (requires Resource Loader)


library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/api/InWorldRecipeRegistries.java
```diff
@@ -0,0 +1,42 @@
...
+	public static final Registry<InWorldRecipe> AXE = create("axe");
+	public static final Registry<InWorldRecipe> SHOVEL = create("shovel");
+	public static final Registry<InWorldRecipe> HOE = create("hoe");
+
+	public static @NotNull Optional<InWorldRecipe> findMatchingRecipe(@NotNull Registry<InWorldRecipe> registry,
+			@NotNull ItemUsageContext context, @NotNull Block targetBlock) {
+		return registry.stream()
+				.filter(recipe -> recipe.targetBlock() == targetBlock && recipe.predicate().test(context))
+				.findFirst();
+	}
```
TheGlitch76: Would this method ever be called in hot code? As it stands right now this is a pretty allocation-heavy and slow way of doing things, but that might be fine.

Leo40Git: I could probably transform this into a regular `for (Entry : registry.getEntries)` ~~(especially since I actually need to change it to return a list).~~

LambdAurora: This definitely needs to be a `for` because the stream API is a disappointment so it will actually go through the entire list and never return early despite you needing just the first match.

BasiqueEvangelist: > I actually need to change it to return a list

Couldn't you add a `doRecipes` method which would try all matching (by block) recipes until one succeeds?

Leo40Git: > Couldn't you add a `doRecipes` method which would try all matching (by block) recipes until one succeeds?

What a coincidence, that's what I just did 🙃

library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/api/InWorldRecipe.java
```diff
@@ -0,0 +1,22 @@
...
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Represents an <em>in-world recipe</em> - a recipe that can be performed on a specific block type in the world.
+ */
+public interface InWorldRecipe {
+	/**
+	 * {@return the specific block type required for this recipe}
+	 */
+	@NotNull Block targetBlock();
```
FoundationGames: Could this use a `Predicate<BlockState>` rather than just a single Block filter? Some may want to filter by block type (as vanilla recipes do) or filter by a specific block state.

As for the data driven part, the predicate can be specified similar to the way loot tables work
(Example, Revision 2)
```json
{
    "predicates": [
        {
            "type": "quilt:block_state",
            "block": "minecraft:oak_slab",
            "properties": {
                "half": "lower"
            }
        },
        {
            "type": "quilt:block_tag",
            "tag": "minecraft:flowers"
        },
        {
            "type": "custom:custom",
            "custom_condition_data": [ "foo", "bar" ]
        }
    ],
    "result": {
        "block": "minecraft:birch_slab",
        "properties": {}
    }
}
```

FoundationGames (changes requested):

library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/api/InWorldRecipe.java
Leo40Git: This can be done by checking the `BlockState` in the `tryPerform` method before actually doing anything.

williambl: What if you want to use a block tag as an input?

Leo40Git: ...good point. Maybe I should just remove `targetBlock()` entirely.

BasiqueEvangelist: > ...good point. Maybe I should just remove `targetBlock()` entirely.

Couldn't you just make it optional? So that if a recipe will only work for one specific block, that can be optimized, but all other recipes will be run every time?

FoundationGames: > This can be done by checking the `BlockState` in the `tryPerform` method before actually doing anything.

As Will said, now you lose the flexibility of allowing multiple block types via tags (though my json example wouldn't work then)

Leo40Git: > Couldn't you just make it optional? So that if a recipe will only work for one specific block, that can be optimized, but all other recipes will be run every time?

That could work, though I'd need registry events to actually implement that optimization.

williambl:
```suggestion
	boolean matches(BlockState state);
```

the most basic impl would just be:
```java
if (this.block != null) {
	return state.getBlock() == this.block;
} else if (this.tag != null) {
	return tag.contains(state.getBlock());
} else return false;
```
but it can be extended to the other blockstate predicate stuff mentioned earlier

BasiqueEvangelist: > the most basic impl would just be:
> 
> ```java
> if (this.block != null) {
> 	return state.getBlock() == this.block;
> } else if (this.tag != null) {
> 	return tag.contains(state.getBlock());
> } else return false;
> ```
> 
> but it can be extended to the other blockstate predicate stuff mentioned earlier

You can just use `tryPerform` for this, instead of adding a new `matches` method.

library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/mixin/PickaxeItemMixin.java
```diff
@@ -0,0 +1,32 @@
...
+
+@Mixin(PickaxeItem.class)
+public abstract class PickaxeItemMixin extends Item {
+	public PickaxeItemMixin() {
+		super(new Settings());
+		throw new AssertionError();
+	}
+
+	@Override
+	public ActionResult useOnBlock(ItemUsageContext context) {
```
Platymemo: overriding methods in mixins is not recommended. I think the ideal path is to inject in a supermixin (mixin to Item) with an empty body and override that handler method in a submixin (this mixin here).

library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/mixin/HoeItemMixin.java
```diff
@@ -0,0 +1,33 @@
...
+import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
+
+@Mixin(HoeItem.class)
+public abstract class HoeItemMixin {
+	@Inject(method = "useOnBlock",
+			at = @At(value = "INVOKE", target = "Ljava/util/Map;get(Ljava/lang/Object;)Ljava/lang/Object;", ordinal = 0),
+			cancellable = true, locals = LocalCapture.CAPTURE_FAILHARD)
+	public void qsl$doCustomRecipes(ItemUsageContext context, CallbackInfoReturnable<ActionResult> cir,
+									World world, BlockPos blockPos) {
+		if (!InWorldRecipeRegistries.tryPerform(InWorldRecipeRegistries.AXE, context, world.getBlockState(blockPos).getBlock()))
```
Platymemo:
```suggestion
		if (!InWorldRecipeRegistries.tryPerform(InWorldRecipeRegistries.HOE, context, world.getBlockState(blockPos).getBlock()))
```

library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/mixin/ShovelItemMixin.java
```diff
@@ -0,0 +1,33 @@
...
+import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
+
+@Mixin(ShovelItem.class)
+public abstract class ShovelItemMixin {
+	@Inject(method = "useOnBlock",
+			at = @At(value = "INVOKE", target = "Ljava/util/Map;get(Ljava/lang/Object;)Ljava/lang/Object;", ordinal = 0),
+			cancellable = true, locals = LocalCapture.CAPTURE_FAILHARD)
+	public void qsl$doCustomRecipes(ItemUsageContext context, CallbackInfoReturnable<ActionResult> cir,
+									World world, BlockPos blockPos, BlockState blockState, PlayerEntity playerEntity) {
+		if (!InWorldRecipeRegistries.tryPerform(InWorldRecipeRegistries.AXE, context, blockState.getBlock()))
```
Platymemo:
```suggestion
		if (!InWorldRecipeRegistries.tryPerform(InWorldRecipeRegistries.SHOVEL, context, blockState.getBlock()))
```

library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/mixin/SwordItemMixin.java
```diff
@@ -0,0 +1,32 @@
...
+
+@Mixin(SwordItem.class)
+public abstract class SwordItemMixin extends Item {
+	public SwordItemMixin() {
+		super(new Settings());
+		throw new AssertionError();
+	}
+
+	@Override
+	public ActionResult useOnBlock(ItemUsageContext context) {
```
Platymemo: Same comment as with the pickaxe mixin, override is not recommended.

Platymemo (commented):
How do we feel about including damage effects as default in `tryPerform` so data recipes can customize how much damage they inflict on the items used? Will need javadocs to recommend calling super

library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/mixin/HoeItemMixin.java
Leo40Git: oops, already fixed this locally! copy+paste is a double-edged sword

library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/api/InWorldRecipe.java
Leo40Git: Ended up changing `Block targetBlock()` to `Set<Block> targetBlocks()`, I think this resolves the issue nicely (as I said, block state can be checked in `tryPerform`).

library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/api/InWorldRecipe.java
```diff
@@ -4,19 +4,26 @@
...
 	 * Tries to perform the recipe.
 	 * @param context item usage context
 	 * @return {@code true} if recipe was successfully performed, {@code false} otherwise
 	 */
 	boolean tryPerform(@NotNull ItemUsageContext context);
+
+	/**
+	 * Invoked on data pack reload. Use this to update your {@linkplain #targetBlocks() target block set}.
+	 */
+	default void onReload() { }
```
BasiqueEvangelist: Is this needed? You could just recalculate the set every time in `targetBlocks`, considering it'll only be used on datapack reload to recreate the multimap.

Leo40Git: Fair point, method yeeted.

library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/mixin/ItemMixin.java
```diff
@@ -0,0 +1,22 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+@Mixin(Item.class)
+public abstract class ItemMixin {
+	@Inject(method = "useOnBlock", at = @At("HEAD"), cancellable = true)
+	public void qsl$useOnBlockOverride(ItemUsageContext context, CallbackInfoReturnable<ActionResult> cir) {
+		cir.setReturnValue(qsl$useOnBlock(context));
```
Platymemo: This affects every item's useOnBlock method that isn't overridden. You want this to not change vanilla behavior unnecessarily.
```suggestion
		ActionResult result = qsl$useOnBlock(context);
		if (result.isAccepted()) {
				cir.setReturnValue(result);
		}
```

Platymemo: Also, overriding the handler method was my original intention, but this also works just as well

Leo40Git:
Adding a "stupid simple" recipe now looks like:
```java
InWorldRecipeRegistries.PICKAXE.register(new Identifier("coolmod", "pick_cool_ore"), new SimpleInWorldRecipe(CoolModBlocks.COOL_ORE) {
    @Override
    public boolean tryPerform(@NotNull ItemUsageContext context) {
        World world = context.getWorld();
        world.setBlockState(context.getBlockPos(), Blocks.STONE.getDefaultState(), 3);
        return super.tryPerform(context);
    }
});
```

library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/mixin/ItemMixin.java
Leo40Git: It returns `ActionResult.PASS` by default, which is the same behavior as vanilla (and I doubt this will change, unless Mojang wants to break interacting with _every single block_).

Leo40Git: I'll allow it to return `null`, which will cause it to defer to the actual method.

Platymemo: oops, for some reason I thought it returned `ActionResult.FAIL`. My b

Platymemo: I think `ActionResult.PASS` should defer to the actual method tho, and only `ActionResult.SUCCESS` or `ActionResult.FAIL` should cause it to deviate

i509VCB:
Could you explain the use case for this module?

Also given we are still quite early and need core things in place in QSL your PR may not be immediately addressed

i509VCB:
Oops

Leo40Git:
> Could you explain the use case for this module?

Normally, adding a stripping/tilling/shoveling(?) recipe requires a mixin. This module reduces the number of mixins from `number_of_mods_that_add_those_kinds_of_recipes` to one.

Also, this extends stripping and shoveling to support more than just "replace any state of this block with this state of this other block" (or even "with the default state of this block" in the case of stripping, though that's because it does property copying by default), and it adds "right click recipe actions" to pickaxes and swords as well.

i509VCB:
So this is really a module to act as a registry for recipe actions based on tool actions?

Leo40Git:
Yes. I was asking about a better name in the Discord, "tool interaction recipe" was suggested.

i509VCB:
Okay, since this involves recipes, @shedaniel do you have any particular requirements for REI et al. for custom recipe systems?

library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/api/InWorldRecipeRegistries.java
```diff
@@ -0,0 +1,31 @@
...
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * <b>NOTE:</b> You need to handle action sound playback yourself.<br>
+ * <b>NOTE2:</b> You <em>can</em> override vanilla recipes with this.
+ */
+public final class InWorldRecipeRegistries {
+	private InWorldRecipeRegistries() { throw new AssertionError(); }
+
+	public static final Registry<InWorldRecipe> SWORD = create("sword");
```
i509VCB: These seem to be based on tool and not action, such as tilling, pathing and debarking. Is there a reason for this name scheme? Or is it more abstract than those specific actions.

Leo40Git: It's more abstract than those specific actions, I feel.

Like, right clicking with a sword could mean carving, stabbing, slicing, etc... the same ambiguity could be applied to the other tool types.

library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/api/InWorldRecipeRegistries.java
```diff
@@ -0,0 +1,31 @@
...
+	public static final Registry<InWorldRecipe> PICKAXE = create("pickaxe");
+	public static final Registry<InWorldRecipe> AXE = create("axe");
+	public static final Registry<InWorldRecipe> SHOVEL = create("shovel");
+	public static final Registry<InWorldRecipe> HOE = create("hoe");
+
+	private static @NotNull Registry<InWorldRecipe> create(@NotNull String toolName) {
+		RegistryKey<Registry<InWorldRecipe>> key =
+				RegistryKey.ofRegistry(new Identifier("qsl-blocks-in-world-recipes", "in_world/" + toolName));
+		Registry<InWorldRecipe> registry = new SimpleRegistry<>(key, Lifecycle.stable());
+		// FIXME add to root registry - not working for some reason???
```
i509VCB: This is because the root registry stuff is a little more complex than that. Soon to come registry api will handle that

library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/impl/InWorldRecipeMaps.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Map;
+
+public final class InWorldRecipeMaps {
+	private InWorldRecipeMaps() { throw new AssertionError(); }
+
+	public static Multimap<Block, InWorldRecipe> sword, pickaxe, axe, shovel, hoe;
+
+	// TODO update this to run on datapack reload
+	// TODO data-drive!
+	public static void update() {
```
i509VCB: Could this cause a thread safety issue if the client uses these maps and tries to display the recipe to a player and the server signals an update, especially on integrated server.

library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/impl/InWorldRecipeMaps.java
```diff
@@ -0,0 +1,50 @@
...
+			if (!recipe.targetBlocks().contains(targetBlock))
+				continue;
+			if (recipe.tryPerform(context))
+				return true;
+		}
+		return false;
+	}
+
+	private static Multimap<Block, InWorldRecipe> fromRegistry(@NotNull Registry<InWorldRecipe> registry) {
+		//noinspection UnstableApiUsage
```
i509VCB: A note: these noinspection lines are not handled by eclipse and the java language server fyi.

library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/api/BlockPredicateUtil.java
```diff
@@ -0,0 +1,21 @@
...
+import net.minecraft.block.Block;
+import net.minecraft.predicate.BlockPredicate;
+import net.minecraft.tag.Tag;
+import org.jetbrains.annotations.NotNull;
+import org.quiltmc.qsl.inworldrecipes.mixin.BlockPredicateAccessor;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public final class BlockPredicateUtil {
```
i509VCB: Hmm what is the primary purpose of the util method in here?

library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/api/BlockPredicateUtil.java
```diff
@@ -0,0 +1,21 @@
...
+	private BlockPredicateUtil() { throw new AssertionError(); }
+
+	public @NotNull Set<Block> getBlocks(@NotNull BlockPredicate predicate) {
+		Tag<Block> tag = ((BlockPredicateAccessor) predicate).getTag();
+		if (tag != null)
+			return new HashSet<>(tag.values()); // /shrug
```
i509VCB: Just do oneliner `if (condition) statement;` if you aren't going to put `{ .. }` on these statements

library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/impl/InWorldRecipeMaps.java
```diff
@@ -0,0 +1,50 @@
+package org.quiltmc.qsl.inworldrecipes.impl;
+
+import com.google.common.collect.*;
```
i509VCB: Please avoid star imports

i509VCB (commented):
Not done yet, trying to address api surface first

library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/api/InWorldRecipeRegistries.java
Leo40Git: This comment was because `Registry.register(Registry.REGISTIRES, Identifier, Registry<InWorldRecipe>)` wasn't resolving.

library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/impl/InWorldRecipeMaps.java
Leo40Git: Will convert to an annotation.

library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/api/BlockPredicateUtil.java
Leo40Git: This is intended for the future data-driven implementation, so we can extract the block set from a recipe's predicate(s) to put them in the multimap.

If a recipe has no block predicates (or the `ANY` block predicate), it'll be added to an alternate list (NYI).

library/blocks/in-world-recipes/src/main/java/org/quiltmc/qsl/inworldrecipes/impl/InWorldRecipeMaps.java
Leo40Git: damn you auto-import

shedaniel:
I am not too sure about the reasoning and scope of this API, the implementation is also, weird.

1. InWorldRecipe, not sure what purpose does targetBlocks serve, it may be for faster lookup, but could be extremely limiting. Even worse is that in InWorldRecipeMaps#tryPerform, it is doing a secondary contains check, I know it would be O(1), but this seems quite redundant and the direction of the api is confusing.

2. InWorldRecipe could not be easily extended by mods (this is a simple fix), InWorldRecipe#tryPerform offers no easy way for mods to alter the functionality. You could possibly make it so there would be an abstract method to do extra things, or just leave the default implementation out.

3. Types of items, this seems to be quite limiting, with each type of tool holding a registry and map, and a mixin for every type of tool. I am not sure but I personally think that it would even be easier for modders to listen to an item use event and have their own system

4. I am not sure how you would data driven this without being more complicated then just allowing people to extend the vanilla recipes / serializers system, vanilla recipes syncs to the clients for you, etc

As for REI requirements, any mods that wish to add REI functionality for something like this would need to have their own plugin for showing the actual functionality (placing widgets, rendering texts, etc), mods would just need to have this info on the client.
(so syncing data driven content to client)

Also, I am not sure why you need a registry and then copy to a map

Leo40Git:
> InWorldRecipe, not sure what purpose does targetBlocks serve, it may be for faster lookup, but could be extremely limiting.

I'll implement an alternate iteration-based path for recipes that doesn't require a set of target blocks.

> Even worse is that in InWorldRecipeMaps#tryPerform, it is doing a secondary contains check,

That's a leftover from an earlier version, oops

> InWorldRecipe#tryPerform offers no easy way for mods to alter the functionality.

??? It can be overriden entirely without calling super (if you don't want the tool to get damaged or want it to be damaged by like 10 or something), not sure what this means - it's entirely customizable, all it needs to do is to return `true` to indicate success or `false` to indicate failure (and failure means the game will try to apply another recipe).

> I am not sure but I personally think that it would even be easier for modders to listen to an item use event and have their own system

I'll probably make the system use an item use event rather than this "mix into each and every tool class" approach.

> I am not sure how you would data driven this

It'll be a fun experience to learn how to do this.

Leo40Git:
Renaming to "tool interaction recipes".

Leo40Git:
~~Screw it I'm adding a block render layer thing as well~~

nvm it just muddles up this PR needlessly

library/blocks/render-layers/src/main/java/org/quiltmc/qsl/blockrenderlayer/mixin/RenderLayersMixin.java
```diff
@@ -0,0 +1,22 @@
...
+import org.spongepowered.asm.mixin.Shadow;
+
+import java.util.Map;
+
+@Mixin(RenderLayers.class)
+public abstract class RenderLayersMixin {
+	@Shadow @Final private static Map<Block, RenderLayer> BLOCKS;
+	@Shadow @Final private static Map<Fluid, RenderLayer> FLUIDS;
+
+	static {
```
Leo40Git: I'm assuming this is executed _after_ the pre-existing static block? Not sure how static blocks work in mixin, honestly.

BasiqueEvangelist:
Why does this have the two modules together? Wouldn't it be better to PR render layers separately, considering that it is ported from FAPI, while in-world recipes is a new module?

Leo40Git:
Closing for now, will be reopened/recreated later.

PR #21, opened 2021-08-01
Blocks API library

Leo40Git:
This pull request creates the Quilt Blocks library.

#### Ports from Fabric API
- [ ] Block render layers
- [ ] Extended block settings
- [ ] Block place/break/interact events

#### Other things

- [ ] [Climbable API](https://github.com/Haven-King/Climbable/blob/master/src/main/java/dev/hephaestus/climbable/api/ClimbingSpeedRegistry.java) except you get context

### DIFF
--- a/library/blocks/build.gradle
+++ b/library/blocks/build.gradle
@@ -1,0 +1,8 @@
+plugins {
+	id("qsl.library")
+}
+
+qslLibrary {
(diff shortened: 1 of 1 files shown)


Leo40Git:
on second thought this should be several smol PRs

PR #22, opened 2021-08-01
Block Extensions API module

Leo40Git:
This PR adds the Block Extensions module (and its containing blocks module).

- [x] Block Render Layer Map
- [x] Extended Block Settings
- [x] Extended Material Builder
- [x] Extendable Extra Block Data system


Leo40Git:
Ready for review!

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/BlockRenderLayerMap.java
```diff
@@ -0,0 +1,85 @@
...
+import net.minecraft.fluid.Fluid;
+import org.jetbrains.annotations.ApiStatus;
+
+import java.util.Map;
+
+/**
+ * Provides methods to set the {@link RenderLayer} of blocks and fluids.
+ */
+@Environment(EnvType.CLIENT)
+public final class BlockRenderLayerMap {
```
LambdAurora: Shouldn't this have an impl counterpart that is `ApiStatus.Internal` to avoid exposing the maps and the initialize method in the API?
And as this is client, this should be in a `client` subpackage.

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/QuiltBlockSettings.java
```diff
@@ -0,0 +1,254 @@
...
+
+	protected QuiltBlockSettings(Material material, MapColor mapColor) {
+		super(material, mapColor);
+	}
+
+	protected QuiltBlockSettings(AbstractBlock.Settings settings) {
+		super(((AbstractBlockSettingsAccessor) settings).getMaterial(), ((AbstractBlockSettingsAccessor) settings).getMapColorProvider());
+
+		// [VanillaCopy] AbstractBlock.Settings#copy(AbstractBlock.Settings)
+		AbstractBlockSettingsAccessor thisAccessor = (AbstractBlockSettingsAccessor) this;
```
LambdAurora: I would use `var` here to shorten the lines and the cast make it obvious which type the local is.
```suggestion
		var thisAccessor = (AbstractBlockSettingsAccessor) this;
```

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/QuiltBlockSettings.java
```diff
@@ -0,0 +1,254 @@
...
+		AbstractBlockSettingsAccessor otherAccessor = (AbstractBlockSettingsAccessor) settings;
```
LambdAurora:
```suggestion
		var otherAccessor = (AbstractBlockSettingsAccessor) settings;
```

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/QuiltBlockSettings.java
```diff
@@ -0,0 +1,254 @@
...
+
+	@Override
+	public QuiltBlockSettings resistance(float resistance) {
+		super.resistance(resistance);
+		return this;
+	}
+
+	/* QUILT ADDTIIONS */
+
+	public QuiltBlockSettings luminance(int luminance) {
```
LambdAurora, line 271: Javadoc would be welcomed.

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/QuiltMaterialBuilder.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+package org.quiltmc.qsl.block.extensions.api;
+
+import net.minecraft.block.MapColor;
+import net.minecraft.block.Material;
+import net.minecraft.block.piston.PistonBehavior;
+import org.quiltmc.qsl.block.extensions.mixin.MaterialBuilderAccessor;
+
+public class QuiltMaterialBuilder extends Material.Builder {
```
LambdAurora, line 27: Doc?

LambdAurora (commented):
Aside from the comments I made, it's nice overall, would love a package-info javadoc on the API package.

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/BlockRenderLayerMap.java
Leo40Git: Maps are `private`, so they're not actually exposed. The `initialize` method, however...

Leo40Git:
Javadoc has been attempted.

LambdAurora (approved):
After re-reading I still can't get what's QuiltBlockInternals, but aside from that it's all good imo.

Leo40Git:
`QuiltBlockInternals` is for future work concerning extra block data. It's empty for now, will have more added onto it later.

Also I should probably make `ExtraData` extensible somehow, hm...

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/client/BlockRenderLayerMap.java
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	/**
+	 * Sets the render layer of the specified fluids.
+	 *
+	 * @param layer new render layer
+	 * @param fluids target fluids
+	 */
+	public static void put(RenderLayer layer, Fluid... fluids) {
+		for (Fluid fluid : fluids)
```
i509VCB: Please use braces here for for loops.

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/data/BlockDataKey.java
```diff
@@ -0,0 +1,52 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.quiltmc.qsl.block.extensions.api.data;
+
+/**
+ * Represents a key that can be used to get a value from an {@link ExtraBlockData} instance.
+ *
+ * @param <T> value type
```
i509VCB: doesn't the doclet set the method docs for a record component when using the `@param` tag?

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/data/BlockDataKey.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+public record BlockDataKey<T>(Class<T> type, String name) {
+	/**
+	 * Creates a new key.
+	 *
+	 * @param type value type
+	 * @param name name
+	 */
+	public BlockDataKey { }
```
i509VCB: I'd argue this is redundant?

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/data/ExtraBlockData.java
```diff
@@ -0,0 +1,127 @@
...
+	/**
+	 * A builder to construct a {@code ExtraBlockData} collection.
+	 */
+	public static final class Builder {
+		private final Map<BlockDataKey<?>, Object> values;
+
+		/**
+		 * Creates a new builder.
+		 */
+		public Builder() {
```
i509VCB: Wouldn't it make more sense to have a `builder` method on `ExtraBlockData` and make this constructor private?

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/data/ExtraBlockData.java
```diff
@@ -0,0 +1,127 @@
...
+		 */
+		public ExtraBlockData build() {
+			return new ExtraBlockData(new HashMap<>(values));
+		}
+	}
+
+	/**
+	 * Invoked to compute an {@code ExtraBlockData} collection for a specific block.
+	 */
+	public interface OnBuild {
```
i509VCB: I need to think about this a bitr

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/data/BlockDataKey.java
Leo40Git: This is the only way to add Javadoc to record constructors/getters, AFAIK.

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/data/BlockDataKey.java
Leo40Git: AFAIK, no.

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/QuiltMaterialBuilder.java
```diff
@@ -0,0 +1,99 @@
...
+	 */
+	public QuiltMaterialBuilder pistonBehavior(PistonBehavior behavior) {
+		((MaterialBuilderAccessor) this).setPistonBehavior(behavior);
+		return this;
+	}
+
+	@Override
+	public Material build() {
+		return super.build();
+	}
```
BasiqueEvangelist: Is this needed, considering you can just use the superclass's `build` method?

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/client/BlockRenderLayerMap.java
```diff
@@ -0,0 +1,77 @@
...
+
+	/**
+	 * Sets the render layer of the specified block.
+	 *
+	 * @param block target block
+	 * @param layer new render layer
+	 */
+	public static void put(Block block, RenderLayer layer) {
+		BlockRenderLayerMapImpl.put(block, layer);
+	}
```
BasiqueEvangelist: Is the single block overload of this method needed, considering you can just use the multi block overload anyway?

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/client/BlockRenderLayerMap.java
```diff
@@ -0,0 +1,77 @@
...
+
+	/**
+	 * Sets the render layer of the specified fluid.
+	 *
+	 * @param fluid target fluid
+	 * @param layer new render layer
+	 */
+	public static void put(Fluid fluid, RenderLayer layer) {
+		BlockRenderLayerMapImpl.put(fluid, layer);
+	}
```
BasiqueEvangelist: Is the single fluid overload of this method needed, considering you can just use the multi fluid overload anyway?

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/data/ExtraBlockData.java
```diff
@@ -0,0 +1,133 @@
...
+		ArrayEvent<OnBuild> EVENT = ArrayEvent.create(OnBuild.class, callbacks -> (block, settings, builder) -> {
+			for (OnBuild callback : callbacks)
+				callback.append(block, settings, builder);
+		});
```
BasiqueEvangelist: Shouldn't this event be parameter invoking?

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/QuiltMaterialBuilder.java
Leo40Git: Oops, accidentally overrode this. You'd think an IDE would say something about overriding a method just to call `super`, huh?

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/client/package-info.java
```diff
@@ -0,0 +1,15 @@
+/**
+ * <h2>Client-side Block Extensions</h2>
+ *
+ * <p>
+ * <h3>{@linkplain org.quiltmc.qsl.block.extensions.api.client.BlockRenderLayerMap Block render layer map}</h3>
+ * <ul>
+ *     <li>Provides methods for setting the render layer of
+ *     {@linkplain org.quiltmc.qsl.block.extensions.api.client.BlockRenderLayerMap#put(net.minecraft.block.Block, net.minecraft.client.render.RenderLayer) blocks}
+ *     and {@linkplain org.quiltmc.qsl.block.extensions.api.client.BlockRenderLayerMap#put(net.minecraft.fluid.Fluid, net.minecraft.client.render.RenderLayer) fluids}.</li>
```
BasiqueEvangelist: This should be removed since the single-block and single-fluid overloads were removed.

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/data/ExtraBlockData.java
```diff
@@ -0,0 +1,137 @@
...
+
+		/**
+		 * Appends key to value pairs to the specified builder.
+		 *
+		 * @param block block
+		 * @param settings block settings
+		 * @param builder collection builder
+		 */
+		@ParameterInvokingEvent
```
BasiqueEvangelist: The `ParameterInvokingEvent` annotation is used incorrectly here. You're supposed to put it on a field that contains an event or a method which *returns* an event.

> ```
>  * <h2>How to use this annotation</h2>
>  *
>  * An event which may invoke a parameter should first place the annotation on the field of the event instance or the
>  * method which returns an event instance.
> ```

BasiqueEvangelist (approved):

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/data/ExtraBlockData.java
i509VCB: ~~Actually leo was correct here. Maybe I need to clarify it goes on the field/method which returns the event instance~~

i509VCB: Never mind I was wrong

pluiedev (approved):

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/impl/QuiltBlockInternals.java
```diff
@@ -0,0 +1,52 @@
...
+
+		ExtraBlockData extraData = internals.qsl$getExtraData();
+		if (extraData == null) {
+			ExtraBlockData.Builder builder = ExtraBlockData.builder();
+			Map<BlockDataKey<?>, Object> map = internals.qsl$getSettingsMap();
+			if (map != null) {
+				copyFromMap(builder, map);
+			}
+			ExtraBlockData.OnBuild.EVENT.invoker().append(block, settings, builder);
+			internals.qsl$setExtraData(extraData = builder.build());
```
BasiqueEvangelist: Couldn't you give the QuiltBlockSettings extra data map directly to the builder, instead of copying everything to the builder?

Leo40Git: I don't think I understood this. This is pretty much the most direct way to get the data map into the builder while not exposing a way to create an invalid `ExtraBlockData` collection (I.E keys mapped to values with incompatible types).

BasiqueEvangelist: How would an invalid `ExtraBlockData` be created here?

Leo40Git: Exposing a constructor that allows passing the map into the data collection/builder directly could allow this to happen:
```java
BlockDataKey<Integer> a = new BlockDataKey<>(Integer.class, new Identifier("badmod", "hehe"));
HashMap<BlockDataKey<?>, Object> map = new HashMap<>();
map.put(a, "This is not an integer!");
ExtraBlockData.Builder builder = ExtraBlockData.builder(map); // this builder will now construct an invalid collection!
```

BasiqueEvangelist: Couldn't the constructor be an implementation detail? Yes, it may cause issues if it's public API ~~you could just cast the BlockDataKey anyway~~, but I'm saying it should be used *here* specifically.

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/impl/QuiltBlockInternals.java
```diff
@@ -0,0 +1,49 @@
...
+
+		return extraData;
+	}
+
+	@SuppressWarnings("unchecked")
+	private static void copyFromMap(ExtraBlockData.Builder builder, Map<BlockDataKey<?>, Object> map) {
+		for (Map.Entry<BlockDataKey<?>, Object> entry : map.entrySet()) {
+			builder.put((BlockDataKey<Object>) entry.getKey(), entry.getValue());
+		}
+	}
```
BasiqueEvangelist: This isn't used anywhere now.

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/impl/QuiltBlockInternals.java
```diff
@@ -0,0 +1,49 @@
...
+public final class QuiltBlockInternals {
+	private QuiltBlockInternals() { }
+
+	public static ExtraBlockData computeExtraData(Block block) {
+		Block.Settings settings = ((AbstractBlockAccessor) block).getSettings();
+		BlockSettingsInternals internals = (BlockSettingsInternals) settings;
+
+		ExtraBlockData extraData = internals.qsl$getExtraData();
+		if (extraData == null) {
+			Map<BlockDataKey<?>, Object> map = internals.qsl$getSettingsMap();
```
BasiqueEvangelist, line 36: Isn't this map only made if `QuiltBlockSettings.extraData` was ever called?

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/data/ExtraBlockData.java
```diff
@@ -0,0 +1,110 @@
...
+ */
+public interface ExtraBlockData {
+	/**
+	 * Constructs a new builder.
+	 *
+	 * @return new builder
+	 */
+	static Builder builder() {
+		return new ExtraBlockDataImpl.BuilderImpl(new HashMap<>());
+	}
```
BasiqueEvangelist: Is this needed? The block's data won't be constructed by the user anyway, you can only add to the builder.

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/data/ExtraBlockData.java
```diff
@@ -0,0 +1,110 @@
...
+		 * @return this builder
+		 */
+		<T> Builder put(BlockDataKey<T> key, T value);
+
+		/**
+		 * Builds the collection.
+		 *
+		 * @return new collection
+		 */
+		ExtraBlockData build();
```
BasiqueEvangelist: Same as for `ExtraBlockData.builder`. Is this needed?

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/impl/QuiltBlockInternals.java
Leo40Git: Removed.

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/impl/QuiltBlockInternals.java
Leo40Git: ...yes, it indeed is. Oops.

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/impl/ExtraBlockDataImpl.java
```diff
@@ -0,0 +1,52 @@
...
+
+	public record BuilderImpl(Map<BlockDataKey<?>, Object> values) implements Builder {
+		@Override
+		public <T> Builder put(BlockDataKey<T> key, T value) {
+			values.put(key, value);
+			return this;
+		}
+
+		public ExtraBlockData build() {
+			return new ExtraBlockDataImpl(new HashMap<>(values));
```
BasiqueEvangelist: You can just pass the `values` map to the `ExtraBlockDataImpl` now, since the builder'll be consumed after build anyway.

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/QuiltBlockSettings.java
```diff
@@ -0,0 +1,308 @@
...
+		thisAccessor.setToolRequired(otherAccessor.isToolRequired());
+
+		// also copy extra data
+		var thisInternals = (BlockSettingsInternals) this;
+		var otherInternals = (BlockSettingsInternals) this;
+
+		Map<BlockDataKey<?>, Object> settingsMap = otherInternals.qsl$getSettingsMap();
+		if (settingsMap != null)
+			thisInternals.qsl$setSettingsMap(new HashMap<>(settingsMap));
+	}
```
BasiqueEvangelist: Does this copy the other settings' loot table ID?

Leo40Git: Nope. Neither Vanilla nor Fabric bother with it, I'll add it now.

BasiqueEvangelist: This also doesn't copy the `jumpVelocityMultiplier` and all the various `*Predicate`s

library/blocks/block-extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/data/BlockDataKey.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+package org.quiltmc.qsl.block.extensions.api.data;
+
+/**
+ * Represents a key that can be used to get a value from an {@link ExtraBlockData} instance.
+ *
+ */
+@SuppressWarnings("ClassCanBeRecord") // this key is identity-based, so it can't be a record!
+public final class BlockDataKey<T> {
```
LambdAurora, line 24: Bit worried about the identity-baseness of this class as you will need to get a hold of the key reference for your stuff.

Like if you had something where the type is Float, which is managed by an external library, you're forced to bring that library in your dev env to be able to use the key since it's identity-based.

Leo40Git: This entire system is being replaced (as soon as I get back to my laptop and also sleep), so this PR will be on hiatus until the new system is done, which will be another QSL library.

Leo40Git:
This PR is reverting to draft mode until #29 is completed and merged.

Leo40Git:
actually I'm just gonna rewrite this, so uh
bye PR

PR #26, opened 2021-08-16
Resource Loader, unique annotation doesn't prevent conflicts.

LambdAurora:
Turns out the `@Unique` annotation doesn't actually prevent conflicts, it only prevents overwriting the field. `@Unique` fields then still should be prefixed for good measure.

This PR adds the prefixes to the `@Unique` fields in resource loader.

### DIFF
--- a/library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/DefaultResourcePackMixin.java
+++ b/library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/DefaultResourcePackMixin.java
@@ -45,11 +45,11 @@ import org.quiltmc.qsl.resource.loader.impl.ResourceLoaderImpl;
 public abstract class DefaultResourcePackMixin {
 	// Redirects all resource access to the MC resource pack.
 	@Unique
-	final ModNioResourcePack quilt$internalPack = this.quilt$locateAndLoad();
+	final ModNioResourcePack quilt$internalPack = this.locateAndLoad();
 
 	@SuppressWarnings({"ConstantConditions", "EqualsBetweenInconvertibleTypes"})
 	@Unique
-	private ModNioResourcePack quilt$locateAndLoad() {
+	private ModNioResourcePack locateAndLoad() {
 		return ResourceLoaderImpl.locateAndLoadDefaultResourcePack(
 				this.getClass().equals(DefaultResourcePack.class) ?
 						ResourceType.SERVER_DATA : ResourceType.CLIENT_RESOURCES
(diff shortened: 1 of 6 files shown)


library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/DefaultResourcePackMixin.java
```diff
@@ -45,11 +45,11 @@
 public abstract class DefaultResourcePackMixin {
 	// Redirects all resource access to the MC resource pack.
 	@Unique
-	final ModNioResourcePack internalPack = this.locateAndLoad();
+	final ModNioResourcePack quilt$internalPack = this.quilt$locateAndLoad();
 
 	@SuppressWarnings({"ConstantConditions", "EqualsBetweenInconvertibleTypes"})
 	@Unique
-	private ModNioResourcePack locateAndLoad() {
+	private ModNioResourcePack quilt$locateAndLoad() {
```
i509VCB: I believe mixin will still scramble these private methods, enable the export environment variable to see further

i509VCB (approved):

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/DefaultResourcePackMixin.java
TheGlitch76: Did you confirm that it doesn't scramble the name in a conflict? The javadoc is very clear that it will, so having to prefix this method would be a bug.
> non-public methods are **renamed** [emphasis original] if a matching target method is
> found, this allows utility methods to be safely assigned meaningful names
 > in code, but renamed if a conflict occurs when a mixin is applied.

TheGlitch76 (changes requested):

library/core/resource-loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/DefaultResourcePackMixin.java
LambdAurora: It only renames if such method/field exists in the final class as part of the original code-base.

This PR was motivated after seeing those screenshots in the Quilt Community Discord.
image
image

LambdAurora: Ok, so, private methods do get scrambled, but fields don't.

TheGlitch76 (approved):

PR #27, opened 2021-08-17
Lifecycle events

i509VCB:
This adds the lifecycle events module to the core library. These events are widely useful for tracking the game through client startup to shutdown and the same with the logical server.

This is not complete at the moment, I've only implemented the server lifecycle so far, I will do the client side shortly. Also some more detailed documentation should be written up.

### DIFF
--- a/library/core/lifecycle-events/src/main/java/org/quiltmc/qsl/lifecycle/api/client/event/ClientLifecycleEvents.java
+++ b/library/core/lifecycle-events/src/main/java/org/quiltmc/qsl/lifecycle/api/client/event/ClientLifecycleEvents.java
@@ -1,0 +1,97 @@
+package org.quiltmc.qsl.lifecycle.api.client.event;
+
+import org.quiltmc.qsl.base.api.event.ArrayEvent;
+
+import net.minecraft.client.MinecraftClient;
+
+import net.fabricmc.api.EnvType;
+import net.fabricmc.api.Environment;
+
+/**
+ * Events indicating the lifecycle of a Minecraft client.
+ *
+ * <p>The lifecycle of a Minecraft client, starts when the client is ready. The client will tick the client and then the
+ * integrated server if in single player.
+ *
+ * @see ClientTickEvents
+ */
+@Environment(EnvType.CLIENT)
+public final class ClientLifecycleEvents {
+	// There is no CLIENT_STARTING event because there is no way to allow mods to register callbacks that early without
+	// possibly initializing the game's registries improperly in preLaunch.
+
+	/**
+	 * An event indicating that a Minecraft client is ready to tick and render.
+	 *
+	 * <p>It should be noted this event is executed while the splash screen is visible, not the main menu.
+	 */
+	public static final ArrayEvent<Ready> READY = ArrayEvent.create(Ready.class, callbacks -> client -> {
+		for (var callback : callbacks) {
+			callback.readyClient(client);
+		}
+	});
+
+	/**
+	 * An event indicating that a Minecraft client has finished it's last tick and will shut down.
+	 *
+	 * <p>If the client is connected to a server, the client will disconnect from the server. Then if the client has a
+	 * running an integrated server, the integrated server will be shutdown. Finally all client facilities are torn down.
+	 *
+	 * <h2>What should mods do when this event is executed</h2>
+	 *
+	 * Mods which maintain session data when connected to a server should should save that data here as those mods still
+	 * have access to the connected server.
+	 *
+	 * <p>If your mod has any data on the integrated server, avoid doing that here, use
+	 * {@link org.quiltmc.qsl.lifecycle.api.event.ServerLifecycleEvents#STOPPING ServerLifecycleEvents.SERVER_STOPPING}
+	 * instead to clean up any data on the integrated server.
+	 */
+	public static final ArrayEvent<Stopping> STOPPING = ArrayEvent.create(Stopping.class, callbacks -> client -> {
+		for (var callback : callbacks) {
+			callback.stoppingClient(client);
+		}
+	});
+
+	/**
+	 * An event indicating the client has finished shutdown and will exit.
+	 *
+	 * <p>The Java Virtual Machine will terminate after this event is executed.
+	 */
+	public static final ArrayEvent<Stopped> STOPPED = ArrayEvent.create(Stopped.class, callbacks -> client -> {
+		for (var callback : callbacks) {
+			callback.stoppedClient(client);
+		}
+	});
+
+	private ClientLifecycleEvents() {}
+
+	/**
+	 * Functional interface to be implemented on callbacks for {@link #READY}.
+	 * @see #READY
+	 */
+	@FunctionalInterface
+	@Environment(EnvType.CLIENT)
+	public interface Ready {
+		void readyClient(MinecraftClient client);
+	}
+
+	/**
+	 * Functional interface to be implemented on callbacks for {@link #STOPPING}.
+	 * @see #STOPPING
+	 */
+	@FunctionalInterface
+	@Environment(EnvType.CLIENT)
+	public interface Stopping {
+		void stoppingClient(MinecraftClient client);
+	}
+
+	/**
+	 * Functional interface to be implemented on callbacks for {@link #STOPPED}.
+	 * @see #STOPPED
+	 */
+	@FunctionalInterface
+	@Environment(EnvType.CLIENT)
+	public interface Stopped {
+		void stoppedClient(MinecraftClient client);
+	}
+}
--- a/library/core/lifecycle-events/src/main/java/org/quiltmc/qsl/lifecycle/api/client/event/ClientLifecycleEvents.java
+++ b/library/core/lifecycle-events/src/main/java/org/quiltmc/qsl/lifecycle/api/client/event/ClientLifecycleEvents.java
@@ -28,13 +28,11 @@ import net.fabricmc.api.Environment;
  *
  * <p>The lifecycle of a Minecraft client, starts when the client is ready. The client will tick the client and then the
  * integrated server if in single player.
- *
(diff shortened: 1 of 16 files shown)


i509VCB:
I believe this is done and ready for review, pending a few spots needing polish.

LambdAurora:
Spotless is not happy with your latest commit.

library/core/lifecycle-events/src/main/java/org/quiltmc/qsl/lifecycle/api/client/event/ClientLifecycleEvents.java
```diff
@@ -0,0 +1,97 @@
...
+	 * <p>If the client is connected to a server, the client will disconnect from the server. Then if the client has a
+	 * running an integrated server, the integrated server will be shutdown. Finally all client facilities are torn down.
+	 *
+	 * <h2>What should mods do when this event is executed</h2>
+	 *
+	 * Mods which maintain session data when connected to a server should should save that data here as those mods still
+	 * have access to the connected server.
+	 *
+	 * <p>If your mod has any data on the integrated server, avoid doing that here, use
+	 * {@link org.quiltmc.qsl.lifecycle.api.event.ServerLifecycleEvents#STOPPING ServerLifecycleEvents.SERVER_STOPPING}
```
BasiqueEvangelist: Why are you renaming `STOPPING` to `SERVER_STOPPING` here?
```suggestion
	 * {@link org.quiltmc.qsl.lifecycle.api.event.ServerLifecycleEvents#STOPPING ServerLifecycleEvents.STOPPING}
```

library/core/lifecycle-events/src/main/java/org/quiltmc/qsl/lifecycle/api/client/event/ClientTickEvents.java
```diff
@@ -0,0 +1,62 @@
...
+	 * An event indicating the client has finished an iteration of the tick loop.
+	 *
+	 * <p>Since there will be a time gap before the next tick, this is a great spot to run any asynchronous operations
+	 * for the next tick.
+	 */
+	public static final ArrayEvent<End> END = ArrayEvent.create(End.class, callbacks -> client -> {
+		for (var callback : callbacks) {
+			callback.endClientTick(client);
+		}
+	});
```
BasiqueEvangelist, line 55: Shouldn't these events also (if it's enabled) push profiler spans for callbacks?

library/core/lifecycle-events/src/main/java/org/quiltmc/qsl/lifecycle/api/event/ServerWorldTickEvents.java
```diff
@@ -0,0 +1,64 @@
...
+
+import org.quiltmc.qsl.base.api.event.ArrayEvent;
+
+import net.minecraft.server.MinecraftServer;
+import net.minecraft.server.world.ServerWorld;
+
+import net.fabricmc.api.EnvType;
+import net.fabricmc.api.Environment;
+
+@Environment(EnvType.CLIENT)
```
BasiqueEvangelist: Why is this common class only present on the client??

library/core/lifecycle-events/src/main/java/org/quiltmc/qsl/lifecycle/api/event/ServerWorldLoadEvents.java
```diff
@@ -0,0 +1,74 @@
...
+
+import org.quiltmc.qsl.base.api.event.ArrayEvent;
+
+import net.minecraft.server.MinecraftServer;
+import net.minecraft.server.world.ServerWorld;
+
+/**
+ * Events indicating whether a world has been loaded or unloaded from a server.
+ *
+ * <p>Mods which implement dynamically loading and unloading worlds from a server may call execute these events to allow
```
BasiqueEvangelist:
```suggestion
 * <p>Mods which implement dynamically loading and unloading worlds from a server may execute these events to allow
```

library/core/lifecycle-events/src/main/java/org/quiltmc/qsl/lifecycle/api/event/ServerWorldTickEvents.java
i509VCB: LOL dumb mistake

library/core/lifecycle-events/src/main/java/org/quiltmc/qsl/lifecycle/api/client/event/ClientTickEvents.java
i509VCB: I have not actually carried over the profiler spans for these events. I find the profiler to be less accurate than something like Spark or VisualVM.

library/core/lifecycle-events/src/main/java/org/quiltmc/qsl/lifecycle/api/client/event/ClientLifecycleEvents.java
```diff
@@ -0,0 +1,130 @@
...
+		for (var callback : callbacks) {
+			callback.readyClient(client);
+		}
+	});
+
+	/**
+	 * An event indicating that a Minecraft client has finished it's last tick and will shut down.
+	 *
+	 * <p>If the client is connected to a server, the client will disconnect from the server. Then if the client has a
+	 * running an integrated server, the integrated server will be shutdown. Finally all client facilities are torn down.
```
OroArmor: does this happen before or after? also the documentation sorta sounds like disconnecting would fire this event.

library/core/lifecycle-events/src/main/java/org/quiltmc/qsl/lifecycle/api/client/event/ClientTickEvents.java
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * Callbacks registered to any of these events should ensure as little time as possible is spent executing, since the tick
+ * loop is a very hot code path.
+ */
+@Environment(EnvType.CLIENT)
+public final class ClientTickEvents {
+	/**
+	 * An event indicating an iteration of the client's tick loop will start.
+	 */
+	public static final ArrayEvent<Start> START = ArrayEvent.create(Start.class, callbacks -> client -> {
```
OroArmor, line 39: Would `BEGIN` be a better name as its the beginning of the tick? also prefixing with `_TICK` might help differentiate between this even and the world start event

OroArmor (commented):
pretty good, i think names shouldnt match as it might get mixed up

library/core/lifecycle-events/src/main/java/org/quiltmc/qsl/lifecycle/api/client/event/ClientLifecycleEvents.java
LambdAurora: I think it happens before iirc

library/core/lifecycle-events/src/main/java/org/quiltmc/qsl/lifecycle/api/client/event/ClientTickEvents.java
LambdAurora: I mean, we also can say that about `START` as it's the start of the tick ;)
Not sure that there's really a need to change those.

And I don't think prefixing here is useful since the class name already does the job.

library/core/lifecycle-events/src/main/java/org/quiltmc/qsl/lifecycle/api/client/event/ClientLifecycleEvents.java
i509VCB: Fired before, disconnection occurs as a result of exiting the client.

library/core/lifecycle-events/src/main/java/org/quiltmc/qsl/lifecycle/api/client/event/ClientTickEvents.java
i509VCB: I deemed the prefix/suffix to be redundant considering the name of the class.

library/core/lifecycle-events/src/main/java/org/quiltmc/qsl/lifecycle/api/client/event/ClientLifecycleEvents.java
LambdAurora: > Fired before, disconnection occurs as a result of exiting the client.

ah, might be worth mentioning it in the docs then, as else it's confusing

library/core/lifecycle-events/src/main/java/org/quiltmc/qsl/lifecycle/api/event/ServerLifecycleEvents.java
```diff
@@ -0,0 +1,170 @@
...
+		 * @param server the server which is shutting down
+		 */
+		void stoppingServer(MinecraftServer server);
+	}
+
+	/**
+	 * Functional interface to be implemented on callbacks for {@link #STOPPED}.
+	 * @see #STOPPED
+	 */
+	@Environment(EnvType.CLIENT)
```
Leo40Git: That... doesn't seem correct, considering the event field isn't also marked as client-only.

Leo40Git (changes requested):

library/core/lifecycle-events/src/main/java/org/quiltmc/qsl/lifecycle/api/event/ServerLifecycleEvents.java
i509VCB: I thought I got these all 😢

Leo40Git (approved):

LambdAurora (approved):

OroArmor (approved):

PR #31, opened 2021-09-04
Crash report info API

### LINKED ISSUE #23: Possible crash report info apis
At the moment the only thing crash report info does is add quilt mods to the system details log. However I see it as being useful to possible allow other mods to add to the system details via a public api.

Some immediate use cases that come to my head could be:
- Mods with native libraries dumping information about said libraries.
- Mods heavily modifying the rendering system such as Sodium and Blaze4D being able to dump info about the OpenGL capabilities/Vulkan extensions that are available/enabled.
- Something like optifabric could insert info about the version of optifine in use.

williambl:
This PR adds six events for adding details to crash reports:
 - `SystemDetails` allows for adding details with no level-context, such as native library versions.
 - `WorldDetails` allows for adding details with a level-context, such as data attached to the player or level.
 - `BlockDetails` allows for adding details in the context of a block.
 - `EntityDetails` allows for adding details about a specific entity.
 - `BlockEntityDetails` allows for adding details about a specific blockentity.
 - `CrashReportCreation` allows for adding new sections to a crash report.

Closes #23.

### DIFF
--- a/library/core/crash_info/src/main/java/org/quiltmc/qsl/crash/impl/CrashInfoImpl.java
+++ b/library/core/crash_info/src/main/java/org/quiltmc/qsl/crash/impl/CrashInfoImpl.java
@@ -3,8 +3,10 @@ package org.quiltmc.qsl.crash.impl;
 import net.fabricmc.api.ModInitializer;
 import net.fabricmc.loader.api.FabricLoader;
 import net.fabricmc.loader.api.ModContainer;
+import org.jetbrains.annotations.ApiStatus;
 import org.quiltmc.qsl.crash.api.CrashReportEvents;
 
+@ApiStatus.Internal
 public final class CrashInfoImpl implements ModInitializer {
 	@Override
 	public void onInitialize() {
--- a/library/core/crash_info/src/main/java/org/quiltmc/qsl/crash/impl/CrashInfoImpl.java
+++ b/library/core/crash_info/src/main/java/org/quiltmc/qsl/crash/impl/CrashInfoImpl.java
@@ -1,0 +1,27 @@
+package org.quiltmc.qsl.crash.impl;
+
+import net.fabricmc.api.ModInitializer;
+import net.fabricmc.loader.api.FabricLoader;
+import net.fabricmc.loader.api.ModContainer;
+import org.quiltmc.qsl.crash.api.CrashReportEvents;
+
+public final class CrashInfoImpl implements ModInitializer {
+	@Override
+	public void onInitialize() {
+		CrashReportEvents.SYSTEM_DETAILS.register(details -> details.addSection("Quilt Mods", () -> {
+			StringBuilder builder = new StringBuilder();
+
+			for (ModContainer mod : FabricLoader.getInstance().getAllMods()) {
+				var metadata = mod.getMetadata();
+
+				builder.append("\n\t\t%s: %s %s".formatted(
+						metadata.getId(),
+						metadata.getName(),
+						metadata.getVersion().getFriendlyString())
+				);
+			}
+
+			return builder.toString();
+		}));
+	}
+}
--- a/library/core/crash_info/src/main/java/org/quiltmc/qsl/crash/mixin/SystemDetailsMixin.java
+++ b/library/core/crash_info/src/main/java/org/quiltmc/qsl/crash/mixin/SystemDetailsMixin.java
@@ -16,8 +16,6 @@
 
 package org.quiltmc.qsl.crash.mixin;
 
-import net.fabricmc.loader.api.FabricLoader;
-import net.fabricmc.loader.api.ModContainer;
 import net.minecraft.util.SystemDetails;
 import org.quiltmc.qsl.crash.api.CrashReportEvents;
 import org.spongepowered.asm.mixin.Mixin;
@@ -38,22 +36,6 @@ abstract class SystemDetailsMixin {
 	 */
 	@Inject(method = "<init>", at = @At("TAIL"))
 	private void addQuiltMods(CallbackInfo info) {
-		this.addSection("Quilt Mods", () -> {
-			StringBuilder builder = new StringBuilder();
-
-			for (ModContainer mod : FabricLoader.getInstance().getAllMods()) {
-				var metadata = mod.getMetadata();
-
-				builder.append("\n\t\t%s: %s %s".formatted(
-						metadata.getId(),
-						metadata.getName(),
-						metadata.getVersion().getFriendlyString())
-				);
-			}
-
-			return builder.toString();
-		});
-
 		CrashReportEvents.SYSTEM_DETAILS.invoker().addDetails((SystemDetails) (Object) this);
 	}
 }
--- a/library/core/crash_info/src/main/java/org/quiltmc/qsl/crash/mixin/SystemDetailsMixin.java
+++ b/library/core/crash_info/src/main/java/org/quiltmc/qsl/crash/mixin/SystemDetailsMixin.java
@@ -16,18 +16,17 @@
 
 package org.quiltmc.qsl.crash.mixin;
 
-import java.util.function.Supplier;
-
+import net.fabricmc.loader.api.FabricLoader;
+import net.fabricmc.loader.api.ModContainer;
+import net.minecraft.util.SystemDetails;
+import org.quiltmc.qsl.crash.api.CrashReportEvents;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 
-import net.minecraft.util.SystemDetails;
-
-import net.fabricmc.loader.api.FabricLoader;
-import net.fabricmc.loader.api.ModContainer;
+import java.util.function.Supplier;
 
 @Mixin(SystemDetails.class)
 abstract class SystemDetailsMixin {
@@ -55,6 +54,6 @@ abstract class SystemDetailsMixin {
 			return builder.toString();
 		});
 
-		// If we wish to add any additional sections, this is a great place to do so.
+		CrashReportEvents.SYSTEM_DETAILS.invoker().addDetails((SystemDetails) (Object) this);
 	}
 }
--- a/library/core/crash_info/src/main/java/org/quiltmc/qsl/crash/api/CrashReportEvents.java
+++ b/library/core/crash_info/src/main/java/org/quiltmc/qsl/crash/api/CrashReportEvents.java
@@ -1,0 +1,30 @@
+package org.quiltmc.qsl.crash.api;
+
+import net.minecraft.util.crash.CrashReportSection;
+import net.minecraft.world.World;
+import org.quiltmc.qsl.base.api.event.ArrayEvent;
+
+public final class CrashReportEvents {
+	public static final ArrayEvent<SystemDetails> SYSTEM_DETAILS = ArrayEvent.create(SystemDetails.class, callbacks -> details -> {
+		for (var callback : callbacks) {
+			callback.addDetails(details);
+		}
+	});
+
+	public static final ArrayEvent<WorldDetails> WORLD_DETAILS = ArrayEvent.create(WorldDetails.class, callbacks -> (world, details) -> {
+		for (var callback : callbacks) {
+			callback.addDetails(world, details);
+		}
+	});
+
+
+	@FunctionalInterface
+	public interface SystemDetails {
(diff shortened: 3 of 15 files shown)


library/core/crash-info/src/main/java/org/quiltmc/qsl/crash/mixin/WorldMixin.java
```diff
@@ -0,0 +1,19 @@
...
+import org.quiltmc.qsl.crash.api.CrashReportEvents;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
+
+@Mixin(World.class)
+public class WorldMixin {
+	@Inject(method = "addDetailsToCrashReport", at = @At("TAIL"), locals = LocalCapture.CAPTURE_FAILSOFT)
```
williambl: Wasn't sure about `LocalCapture.CAPTURE_FAILSOFT` - do we prefer crashes for this?

OroArmor (approved):

OroArmor:
Could there be a way to add a new section? I know that there are parts thta go more into depth about the crash, like rendering or ticking an entity

williambl:
> Could there be a way to add a new section? I know that there are parts thta go more into depth about the crash, like rendering or ticking an entity

I can see this being useful. That also makes me realise that it might be useful to add events which allow adding extra info to those sections, by injecting to `BlockEntity#populateCrashReport`, `Entity#populateCrashReport`, and `CrashReportSection#addBlockInfo`

library/core/crash-info/src/main/java/org/quiltmc/qsl/crash/mixin/WorldMixin.java
i509VCB: I believe we want to do fail hard everywhere

library/core/crash-info/src/main/java/org/quiltmc/qsl/crash/mixin/WorldMixin.java
```diff
@@ -0,0 +1,19 @@
...
+import net.minecraft.world.World;
+import org.quiltmc.qsl.crash.api.CrashReportEvents;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
+
+@Mixin(World.class)
+public class WorldMixin {
```
i509VCB: Abstract mixins please

library/core/crash-info/src/main/java/org/quiltmc/qsl/crash/mixin/WorldMixin.java
```diff
@@ -0,0 +1,19 @@
...
+	@Inject(method = "addDetailsToCrashReport", at = @At("TAIL"), locals = LocalCapture.CAPTURE_FAILSOFT)
+	void quilt$addCrashReportDetails(CrashReport report, CallbackInfoReturnable<CrashReportSection> cir, CrashReportSection crashReportSection) {
```
i509VCB: Namespace not necessary in an injector.

library/core/crash-info/src/main/java/org/quiltmc/qsl/crash/mixin/WorldMixin.java
i509VCB: I recall some odd behavior in a past mc version where this may have been initialized at some point earlier in game lifecycle?

i509VCB: If that behavior still applies, the javadoc should make it clear this is just for obtaining info and a crash may not be in progress?

i509VCB:
> Could there be a way to add a new section? I know that there are parts thta go more into depth about the crash, like rendering or ticking an entity

For entities we could do auto invoking, however I am not sure of the utility for these hooks on entities?

OroArmor:
I was more suggesting being able to add those types of sections, not add to them

OroArmor:
For example, Blaze4D could dump a bunch of state info into a dedicated section, rather than cluttering the rest of the global data

library/core/crash-info/src/main/java/org/quiltmc/qsl/crash/mixin/WorldMixin.java
williambl: Looks like it's also used to create a debug report (from /debug)

library/core/crash-info/src/main/java/org/quiltmc/qsl/crash/api/CrashReportEvents.java
```diff
@@ -0,0 +1,165 @@
...
+	@FunctionalInterface
+	public interface SystemDetails {
+		void addDetails(net.minecraft.util.SystemDetails details);
+	}
+
+	/**
+	 * Functional interface to be implemented on callbacks for {@link #SYSTEM_DETAILS}.
+	 *
+	 * @see #SYSTEM_DETAILS
+	 */
```
BasiqueEvangelist:
```suggestion
	/**
	 * Functional interface to be implemented on callbacks for {@link #WORLD_DETAILS}.
	 *
	 * @see #WORLD_DETAILS
	 */
```

library/core/crash-info/src/main/java/org/quiltmc/qsl/crash/mixin/CrashReportSectionMixin.java
```diff
@@ -0,0 +1,36 @@
...
+import org.quiltmc.qsl.crash.api.CrashReportEvents;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
+
+@Mixin(CrashReportSection.class)
+public abstract class CrashReportSectionMixin {
+	@Inject(method = "addBlockInfo", at = @At("TAIL"), locals = LocalCapture.CAPTURE_FAILHARD)
```
BasiqueEvangelist: This doesn't need to capture any locals.
```suggestion
	@Inject(method = "addBlockInfo", at = @At("TAIL"))
```

library/core/crash-info/src/main/java/org/quiltmc/qsl/crash/mixin/WorldMixin.java
```diff
@@ -0,0 +1,35 @@
...
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
+
+@Mixin(World.class)
+public abstract class WorldMixin {
+	@Inject(method = "addDetailsToCrashReport", at = @At("TAIL"), locals = LocalCapture.CAPTURE_FAILHARD)
+	void addCrashReportDetails(CrashReport report, CallbackInfoReturnable<CrashReportSection> cir, CrashReportSection crashReportSection) {
+		CrashReportEvents.WORLD_DETAILS.invoker().addDetails((World) (Object) this, crashReportSection);
+	}
```
BasiqueEvangelist: Can't you just use the to-be-returned `CrashReportSection` here?
```suggestion
	@Inject(method = "addDetailsToCrashReport", at = @At("TAIL"))
	void addCrashReportDetails(CrashReport report, CallbackInfoReturnable<CrashReportSection> cir) {
		CrashReportEvents.WORLD_DETAILS.invoker().addDetails((World) (Object) this, cir.getReturnValue());
	}
```

i509VCB:
Do we want to implement Qsl's add mods to system details via the new event?

LambdAurora:
> Do we want to implement Qsl's add mods to system details via the new event?

We can, would reduce the size of the injector, but I don't know if it's worth it.

library/core/crash-info/src/testmod/java/org/quiltmc/qsl/crash/test/mixin/EndCrystalEntityMixin.java
```diff
@@ -0,0 +1,42 @@
...
+@Mixin(EndCrystalEntity.class)
+public abstract class EndCrystalEntityMixin extends Entity {
+	public EndCrystalEntityMixin(EntityType<?> type, World world) {
+		super(type, world);
+	}
+
+	@Inject(method = "tick", at=@At("HEAD"))
+	void crashOnTick(CallbackInfo ci) {
+		if (world.getBlockState(getBlockPos().down()).getBlock() == Blocks.DIAMOND_BLOCK) {
+			kill();
```
LambdAurora: Would prefer this to have a `this.` to make it a bit more clear that it calls the Entity's kill method, makes it easier to determine where stuff is when looking outside of an IDE.

LambdAurora (approved):
Overall good, just need to determine if add mods to system details should be via the event, then for me it would be good to merge.

williambl:
@LambdAurora @i509VCB I don't think it's worth it, because it would involve adding another entry point for not all that much benefit.

i509VCB:
> @LambdAurora @i509VCB I don't think it's worth it, because it would involve adding another entry point for not all that much benefit.

is this a case of events being invoked very early on?

williambl:
I mean it'd be a normal ModInitalizer entrypoint, the event doesn't need to be done at a specific time or anything, it's just that the module currently is entrypointless and it feels cleaner to me to keep it that way. If the consensus is for adding an entrypoint for this though then I have nothing against doing that.

i509VCB:
> I mean it'd be a normal ModInitalizer entrypoint, the event doesn't need to be done at a specific time or anything, it's just that the module currently is entrypointless and it feels cleaner to me to keep it that way. If the consensus is for adding an entrypoint for this though then I have nothing against doing that.

Well modules having entrypoints is nothing new, events for tracking loading of entities using the lifecycle events to fire removals on server shutdown, I think it would make sense to eat our own dog food here with regards to the apis we provide where possible to reduce special codepaths.

williambl:
Done, mods list now uses the events. I'm wondering whether it might be best to put the mods list in its own section?

williambl:
Rebased, no conflicts now. I believe this is complete. (unless somebody agrees that it's a good idea to put the mods list in its own section)

library/core/crash_info/src/main/java/org/quiltmc/qsl/crash/impl/CrashInfoImpl.java
```diff
@@ -0,0 +1,27 @@
+package org.quiltmc.qsl.crash.impl;
+
+import net.fabricmc.api.ModInitializer;
+import net.fabricmc.loader.api.FabricLoader;
+import net.fabricmc.loader.api.ModContainer;
+import org.quiltmc.qsl.crash.api.CrashReportEvents;
+
+public final class CrashInfoImpl implements ModInitializer {
```
LambdAurora, line 10: `@ApiStatus.Internal`

LambdAurora (approved):

OroArmor (approved):

library/core/crash_info/src/main/java/org/quiltmc/qsl/crash/impl/CrashInfoImpl.java
williambl: Done. In future, should that be on all impl classes?

OroArmor: Yes.

library/core/crash_info/src/main/java/org/quiltmc/qsl/crash/mixin/SystemDetailsMixin.java
```diff
@@ -16,18 +16,15 @@
...
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 
-import net.minecraft.util.SystemDetails;
-
-import net.fabricmc.loader.api.FabricLoader;
-import net.fabricmc.loader.api.ModContainer;
+import java.util.function.Supplier;
 
 @Mixin(SystemDetails.class)
 abstract class SystemDetailsMixin {
```
IotaBread, line 30: Is the shadow of `#addSection` (which I can't comment on) used anywhere? If not I guess it can be removed

LambdAurora:
Will merge, and it will cause build to fail but that's expected, will fix manually.

PR #33, opened 2021-09-05
Command module

williambl:
# Command API

Consists of two new modules, `command` and `client-command`, in a new library, `command`.

Mostly a port of fabric's Commands API v1, with a few differences:

 - The command registration event has a `registerIntegrated` parameter, for commands which should only be registered on an integrated server, like `/publish`
 - Client command registration is done through an event
 - Client help command is different (`/qcc help`, longer alternative to `qcc` being `quilt_commands:client_commands`)
 - Client commands now have a lower priority to server ones, as was originally intended for the fabric commands API but not implemented.
 - Client commands can be run with `/qcc run <command>` if they are overridden by a server command


BasiqueEvangelist:
Why is client command registration done with an event? Mods still have access to the `DISPATCHER` field, so is there really any reason to use the event?

OroArmor:
The event makes way more sense. The field probably can become private. The only problem is when to fire the event

williambl:
For consistency with the server-side command API. The field is not necessarily needed to be exposed in API any more.

williambl:
@TehcJS why remove registerForIntegrated? It's more than just `!dedicated` - there are situations where both registerForIntegrated and registerForDedicated are true. (Admittedly, only in datagen and running checks on game boot for missing arguement types, but we should keep consistency in those cases regardless)

sylv256:
> @TehcJS why remove registerForIntegrated? It's more than just `!dedicated` - there are situations where both registerForIntegrated and registerForDedicated are true. (Admittedly, only in datagen and running checks on game boot for missing arguement types, but we should keep consistency in those cases regardless)

In what case does that apply?

sylv256:
also i'll re-add it. I didn't know it could be both integrated and dedicated.

sylv256:
they should be called "integrated" and "dedicated"

williambl:
> they should be called "integrated" and "dedicated"

they were originally, it didn't seem clear enough though. Rename if you want though, javadoc can handle clarity.

sylv256:
i'll also write javadoc for it, yeah

williambl:
I'm currently writing up some javadoc, dont worry

library/commands/build.gradle
```diff
@@ -0,0 +1,8 @@
+plugins {
+	id("qsl.library")
+}
+
+qslLibrary {
+	libraryName = "commands"
```
TheGlitch76: This needs to be `command` per the RFC

library/commands/qsl_commands/src/main/java/org/quiltmc/qsl/commands/api/client/QuiltClientCommandSource.java
```diff
@@ -0,0 +1,108 @@
...
+	 * @return the player
+	 */
+	ClientPlayerEntity getPlayer();
+
+	/**
+	 * Gets the entity that used the command.
+	 *
+	 * @return the entity
+	 */
+	default Entity getEntity() {
```
TheGlitch76: It feels like this interface is missing some `@Override` annotations here. Are these methods coming from `CommandSource`?

library/commands/qsl_commands/src/main/java/org/quiltmc/qsl/commands/api/client/QuiltClientCommandSource.java
```diff
@@ -0,0 +1,108 @@
...
+	/**
+	 * Gets the meta property under {@code key} that was assigned to this source.
+	 *
+	 * <p>This method should return the same result for every call with the same {@code key}.
+	 *
+	 * @param key the meta key
+	 * @return the meta
+	 */
+	default Object getMeta(String key) {
+		return null;
```
TheGlitch76: Perhaps document that this does nothing for client command sources? Or was this left unimplemented on accident?

TheGlitch76 (changes requested):
api overall lgtm, but the library name must be fixed before merge

library/commands/qsl_commands/src/main/java/org/quiltmc/qsl/commands/api/client/QuiltClientCommandSource.java
LambdAurora: And if it's not an `@Override` and just a context class, we can drop all `get` prefixes.

library/commands/qsl_commands/build.gradle
```diff
@@ -0,0 +1,12 @@
+plugins {
+	id("qsl.module")
+}
+
+qslModule {
+	moduleName = "qsl_commands"
+	version = "1.0.0"
+	library = "commands"
```
OroArmor:
```suggestion
	library = "command"
```

library/commands/qsl_commands/src/main/java/org/quiltmc/qsl/commands/impl/client/ClientCommandInternals.java
```diff
@@ -0,0 +1,285 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.quiltmc.qsl.commands.api.client.ClientCommandManager.*;
+
+@Environment(EnvType.CLIENT)
+public final class ClientCommandInternals {
+	private static final Logger LOGGER = LogManager.getLogger();
+	private static final char PREFIX = '/';
```
OroArmor: Any chance to support different prefixes for client commands? I know its not the largest issue, but I think that it does have its strengths, and there have been a few people that i have interacted with and said that they would reimplement the client command system for custom prefixes.

settings.gradle
```diff
@@ -19,6 +19,7 @@ includeBuild("build-logic")
 
 // Libraries
 library("core")
+library("commands")
```
OroArmor:
```suggestion
library("command")
```

OroArmor (changes requested):
The package should be renamed from `org.quiltmc.qsl.commands` to `org.quiltmc.qsl.command`

library/commands/qsl_commands/src/main/java/org/quiltmc/qsl/commands/impl/client/ClientCommandInternals.java
Juuxel: For this, feel free to take a look at my original implementation for Fabric API which had custom symbol prefixes (before [this commit](https://github.com/Juuxel/fabric/commit/362b2f4c097a3367b716ab1dde5fbe2ee8b87cd7), so at [this revision](https://github.com/Juuxel/fabric/tree/148d88b8f91450510528ba1c110ec2c5b465d5e3)). (I don't mind if you take the code for QSL)

library/commands/qsl_commands/src/main/java/org/quiltmc/qsl/commands/api/client/QuiltClientCommandSource.java
williambl: > It feels like this interface is missing some `@Override` annotations here. Are these methods coming from `CommandSource`?

These are extensions to `ClientCommandSource`, all of which (but `getClient` and `getMeta`) have equivalents on `ServerCommandSource` - nothing is being overriden here. I will make the javadoc more understandable.

> And if it's not an `@Override` and just a context class, we can drop all `get` prefixes.

The prefixes make the names the same as the methods' equivalents on `ServerCommandSource` (on QM).

library/commands/qsl_commands/src/main/java/org/quiltmc/qsl/commands/api/client/QuiltClientCommandSource.java
williambl: I'm not _entirely sure_ what the point of this is, to be honest. I understand that it's there so that custom implementations can implement it and provide extra meta, but I don't understand the motivation. I'd been meaning to ask Juuz about it, and will do on the toolchain discord to try and understand this more.

williambl:
Going to rebase this onto 1.18 and force-push.

library/commands/qsl_commands/src/main/java/org/quiltmc/qsl/commands/api/client/QuiltClientCommandSource.java
williambl: Ah, looks like it was instead added by xpple. I'll look into what motivated the addition.

library/commands/qsl_commands/src/main/java/org/quiltmc/qsl/commands/impl/client/ClientCommandInternals.java
williambl: > For this, feel free to take a look at my original implementation for Fabric API which had custom symbol prefixes (before [this commit](https://github.com/Juuxel/fabric/commit/362b2f4c097a3367b716ab1dde5fbe2ee8b87cd7), so at [this revision](https://github.com/Juuxel/fabric/tree/148d88b8f91450510528ba1c110ec2c5b465d5e3)). (I don't mind if you take the code for QSL)

Did this include completion for non-'/' commands?

library/command/qsl_command/src/main/java/org/quiltmc/qsl/command/api/CommandRegistrationCallback.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Event for registering server-side commands.
+ */
+@FunctionalInterface
+public interface CommandRegistrationCallback {
+	/**
+	 * Invoked when server-side commands are registered.
+	 */
+	ArrayEvent<CommandRegistrationCallback> EVENT = ArrayEvent.create(CommandRegistrationCallback.class, callbacks -> (dispatcher, integrated, dedicated) -> {
```
Genau6502: If a server is not dedicated it is integrated - no need for an extra parameter

williambl: Not always the case - there are debug and datagen situations where both the integrated and dedicated parameters are true.

library/commands/qsl_commands/src/main/java/org/quiltmc/qsl/commands/impl/client/ClientCommandInternals.java
williambl: > > For this, feel free to take a look at my original implementation for Fabric API which had custom symbol prefixes (before [this commit](https://github.com/Juuxel/fabric/commit/362b2f4c097a3367b716ab1dde5fbe2ee8b87cd7), so at [this revision](https://github.com/Juuxel/fabric/tree/148d88b8f91450510528ba1c110ec2c5b465d5e3)). (I don't mind if you take the code for QSL)
> 
> Did this include completion for non-'/' commands?

Checked now, it doesn't.

I'm not a fan of per-command prefixes for multiple reasons:
 - breaks user expectations of what a command _is_ - "a command starts with `/`" is what most people first learn about commands.
 - means that we have to implement command suggestions and completion for non-`/` prefixes
 - `/` will always be special-cased _anyway_, having different code run (to prevent collisions with server commands, and also its system for command suggestion will be different to that of other prefixes)
 - extra mental load on the user, to remember which commands use which prefixes.

I wrote a partial implementation (no suggestions for non-`/` commands, slightly borked help command) based off Juuxel's work in another branch: https://github.com/williambl/quilt-standard-libraries/commit/4a994cfee06bf9a740741e4dad72fb6b3264e227.

williambl:
Regarding the `qsl_command` module - it currently has two responsibilities: the client command system, and server-side command registration. Would we be better off splitting it into `qsl_client_command` and `qsl_command`? Especially since the `command` library only contains this one module.

Also should the `qsl_` be dropped? It doesn't seem to be part of any other module name other than `qsl_base`.

LambdAurora:
> Also should the `qsl_` be dropped? It doesn't seem to be part of any other module name other than `qsl_base`.

It should be dropped, and the modid/namespace should be `quilt_<module name>`.

library/commands/qsl_commands/src/main/java/org/quiltmc/qsl/commands/impl/client/ClientCommandInternals.java
Juuxel: Yeah, they complicate the code IMO with not many benefits. I imagine the main use case is for mods grouping their commands under their own prefix. That can be done much more descriptively with a common root command (`/mod_id`), so this might not be worth it to include (especially since the autocomplete might not be fully functional).

Although... I recall having the autocomplete at least partially working in my original PR:

library/command/command/src/testmod/java/org/quiltmc/qsl/command/test/CommandApiTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+			dispatcher.register(literal("overrideme")
+					.executes(ctx -> {
+						ctx.getSource().sendFeedback(new LiteralText("Server!"), false);
+						return 0;
+					})
+			);
+		});
+
+		ClientCommandRegistrationCallback.EVENT.register(dispatcher -> {
```
Juuxel: Does this crash on a dedi server?

library/command/command/src/main/java/org/quiltmc/qsl/command/impl/client/ClientCommandInternals.java
```diff
@@ -0,0 +1,289 @@
...
+			originalToCopy.put(child, result);
+			target.addChild(result);
+
+			if (!child.getChildren().isEmpty()) {
+				copyChildren(child, result, source, originalToCopy);
+			}
+		}
+	}
+
+	private static final Command<QuiltClientCommandSource> DUMMY_COMMAND = ctx -> 0;
```
Juuxel: Putting this with the other `private static final` fields would help with clarity.

Juuxel (approved):

library/commands/qsl_commands/src/main/java/org/quiltmc/qsl/commands/api/client/QuiltClientCommandSource.java
williambl: Implemented now.

library/command/command/src/testmod/java/org/quiltmc/qsl/command/test/CommandApiTest.java
williambl: Yep, thanks. Fixed,

library/command/command/src/main/java/org/quiltmc/qsl/command/impl/client/ClientCommandInternals.java
williambl: Done.

TheGlitch76 (approved):

TheGlitch76:
@williambl Please fix license headers

LambdAurora (approved):

LambdAurora:
> @williambl Please fix license headers

There's a bit more than license headers, ArrayEvent has been renamed to Event in the meantime, so that's also a thing to consider.

williambl:
Not to bikeshed on my own PR or anything but I'm still on the fence about this:

> Regarding the `qsl_command` module - it currently has two responsibilities: the client command system, and server-side command registration. Would we be better off splitting it into `qsl_client_command` and `qsl_command`? Especially since the `command` library only contains this one module.

LambdAurora:
Mmh, does splitting brings any benefit?
Is there really nothing in the code that is shared between the two theoretical modules?

TheGlitch76:
> Regarding the qsl_command module - it currently has two responsibilities: the client command system, and server-side command registration. Would we be better off splitting it into qsl_client_command and qsl_command? Especially since the command library only contains this one module.

A library only having one module isn't a justification to split a module :P

I would say that a split is warranted if, in the general case, a mod would use one part of the module but not the other, and there is no code share between the two modules.

williambl:
> A library only having one module isn't a justification to split a module :P

Fairs

> Is there really nothing in the code that is shared between the two theoretical modules?

> I would say that a split is warranted if, in the general case, a mod would use one part of the module but not the other, and there is no code share between the two modules.

There is no code share between the two would-be modules - and imo, a mod using one is unlikely to use both (I just checked three usages (via github code search) of fabric's client commands API, and all three were in client-only mods and therefore would not be using the server commands api)

williambl:
Checks passing now, split done too.

TheGlitch76:
I'm okay with this PR entering a final comment period in that case, if Aurora is too

PR #35, opened 2021-10-04
Quilt Tags API

LambdAurora:
Big PR, but very useful features, including:
 - TagType! A way to assign some behavior to tags like making them load on the client only, or have a tag that have a default on client in-case the server doesn't have a tag, or prevent the loading of a world/server if a tag isn't present to avoid breakage, etc.
 - TagRegistry! The equivalent of TagFactory from FAPI, includes a new method to accept a TagType parameter, all tags are defaulted to TagType.NEUTRAL otherwise. Also include biome tags.
 - Do not include an equivalent to `FabricDataGeneratorTagBuilder` since the class it hooks into is not public, can be ported once global access wideners are a thing perhaps.
 
 Reviews are welcome, testing is encouraged!


Leo40Git (approved):

LambdAurora:
Latest commit includes the latest tweaks I wanted to include, which tl;dr is preventing log spam of missing entries from tags due to dynamic registries shenanigans.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/TagRegistry.java
```diff
@@ -0,0 +1,86 @@
...
+	/**
+	 * Creates a new identified neutral tag.
+	 *
+	 * @param id the identifier of the tag
+	 * @return the identified neutral tag
+	 * @see #create(Identifier, TagType) create a tag with a specific type
+	 */
+	default Tag.Identified<T> create(Identifier id) {
+		return this.create(id, TagType.CLIENT_SERVER_REQUIRED);
+	}
```
BasiqueEvangelist: Wouldn't the neutral tag type be `TagType.NEUTRAL`?

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/package-info.java
```diff
@@ -0,0 +1,29 @@
...
+ * For example, Minecraft use those to identify all stairs blocks in the game, which is used for the connection mechanic.
+ * <p>
+ * By default tags are available for most of the static registries and are one of the earliest resource to be loaded on
+ * server start.
+ * <p>
+ * The goal of the API is to give to modders the ability to get (and register) tags of any registries, dynamic ones included.
+ *
+ * <p>
+ * <h3>Tag Types</h3>
+ * {@link org.quiltmc.qsl.tag.api.TagType} allows to categorize tags and define extra-behavior.
```
BasiqueEvangelist:
```suggestion
 * {@link org.quiltmc.qsl.tag.api.TagType} allows categorizing tags and defining extra behavior.
```

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/package-info.java
```diff
@@ -0,0 +1,29 @@
...
+ * Those types allows to define if a tag is required for a server to start, for a client to connect,
```
BasiqueEvangelist:
```suggestion
 * Those types allow defining if a tag is required for a server to start, for a client to connect,
```

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/package-info.java
```diff
@@ -0,0 +1,29 @@
...
+ * if a tag is only loaded on the client, etc.
+ *
+ * <p>
+ * <h3>Get and register a tag</h3>
+ * To get and/or register a tag, use {@link org.quiltmc.qsl.tag.api.TagRegistry} create methods.
```
BasiqueEvangelist:
```suggestion
 * To get and/or register a tag, use {@link org.quiltmc.qsl.tag.api.TagRegistry}'s create methods.
```

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/TagRegistry.java
LambdAurora: Most of MC's default tags falls under the category `CLIENT_SERVER_REQUIRED`, but I can default to `NEUTRAL` again like FAPI currently does.

BasiqueEvangelist: Yes, but you're saying in the javadoc that the method creates a neutral tag.

LambdAurora: *oh*

LambdAurora: Forgot to say, but should be fixed now.

LemmaEOF (approved):

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/package-info.java
```diff
@@ -0,0 +1,29 @@
+/**
```
williambl, line 1: Is it possible to `{@see}` `package-info`s? If so, it might be useful to stick that at the bottom of every class javadoc in this package, for discoverability. It's possible that not many people will look at the package-info for documentation, which is a shame as this is really useful.

LambdAurora: It is possible, latest commit adds it.

build-logic/src/main/java/qsl/internal/extension/QslModuleExtension.java
```diff
@@ -62,18 +62,38 @@ public void setVersion(String version) {
...
+	}
+
+	public void coreApiDependencies(Iterable<String> dependencies) {
 		for (String dependency : dependencies) {
-			Map<String, String> map = new LinkedHashMap<>(2);
-			map.put("path", ":core:" + dependency);
-			map.put("configuration", "dev");
+			Dependency project = this.getCoreModule(dependency);
+			this.moduleDependencies.add(project);
+			this.project.getDependencies().add(JavaPlugin.API_CONFIGURATION_NAME, project);
```
i509VCB: Ah yeah big oops from me, this should be api everywhere

library/data/tags/build.gradle
```diff
@@ -0,0 +1,20 @@
...
+	coreDependencies([
+			"resource_loader"
+	])
+	coreTestmodDependencies([
+			"qsl_base",
+			"lifecycle_events"
+	])
+}
+
+minecraft {
```
i509VCB: loom, minecraft name is deprecated

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/QuiltTag.java
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+
+package org.quiltmc.qsl.tag.api;
+
+import net.minecraft.tag.Tag;
+
+/**
+ * Interface implemented by {@link net.minecraft.tag.Tag} instances when QSL is present.
+ *
```
i509VCB, line 28: This semantic has caused issues in the past before. Such an interface should exist for other tag impls but access to functions should be done through static helper functions that may return a value if this data is not present.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/QuiltTagBuilder.java
```diff
@@ -0,0 +1,39 @@
...
+package org.quiltmc.qsl.tag.api;
+
+import net.minecraft.tag.Tag;
+
+/**
+ * Interface implemented by {@link net.minecraft.tag.Tag.Builder} instances when QSL is present.
+ *
+ * @see org.quiltmc.qsl.tag.api
+ */
+public interface QuiltTagBuilder {
```
i509VCB, line 39: This makes the builder very awkward to use, I'd prefer a helper that temporarily borrows the builder and applies the action, returning the same builder in the process

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/TagRegistry.java
```diff
@@ -0,0 +1,87 @@
...
+
+import org.quiltmc.qsl.tag.impl.TagRegistryImpl;
+
+/**
+ * Represents a tag registry.
+ *
+ * @param <T> the type of the registered tags
+ * @see org.quiltmc.qsl.tag.api
+ */
+public interface TagRegistry<T> {
```
i509VCB, line 49: Should this be non-extendable?

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/TagType.java
```diff
@@ -0,0 +1,75 @@
...
+	 * the client will disconnect when attempting to connect to said server.
+	 */
+	CLIENT_SERVER_REQUIRED(true),
+	/**
+	 * Represents a neutral tag type, tags are not required to start nor to connect,
+	 * but they will sync if present on the server.
+	 * <p>
+	 * The client does not provide a default.
+	 */
+	NEUTRAL(false),
```
i509VCB: This name confuses me. If you describe under what condition the tags of each type are made I can propose better names.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/TagType.java
```diff
@@ -0,0 +1,75 @@
...
+
+	public boolean isRequiredToStart() {
+		return this.requiredToStart;
+	}
+
+	public boolean isRequiredToConnect() {
+		return this == CLIENT_SERVER_REQUIRED;
+	}
+
+	public boolean hasSync() {
```
i509VCB, line 79: hasSync as in the tag may be synchronized to clients?

library/data/tags/src/main/java/org/quiltmc/qsl/tag/impl/QuiltRequiredTagListHooks.java
```diff
@@ -0,0 +1,29 @@
...
+import org.jetbrains.annotations.ApiStatus;
+
+import net.minecraft.tag.Tag;
+import net.minecraft.util.Identifier;
+
+import org.quiltmc.qsl.tag.api.TagType;
+
+@ApiStatus.Internal
+public interface QuiltRequiredTagListHooks<T> {
+	Tag.Identified<T> quilt$addTag(Identifier id, TagType type);
```
i509VCB: Maybe name this `qsl$addTypedTag` to make readers know this function also includes a sync type

library/data/tags/src/main/java/org/quiltmc/qsl/tag/impl/TagDelegate.java
```diff
@@ -0,0 +1,112 @@
...
+
+import net.minecraft.tag.Tag;
+import net.minecraft.tag.TagGroup;
+import net.minecraft.util.Identifier;
+
+import org.quiltmc.qsl.tag.api.QuiltTag;
+import org.quiltmc.qsl.tag.api.TagType;
+
+@ApiStatus.Internal
+public class TagDelegate<T> implements Tag.Identified<T>, QuiltTag<T>, QuiltTagHooks {
```
i509VCB, line 46: Docs should probably deacribe the thread safety semantics of this class along with the fact this tag automatically refers to the latest set of tags from a group.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/impl/TagDelegate.java
```diff
@@ -0,0 +1,112 @@
...
+	public String toString() {
+		return "TagDelegate{id=\"" + this.id
+				+ "\", type=" + this.type
+				+ ", target=" + this.target
+				+ ", clearCount=" + this.clearCount
+				+ ", values=" + this.values()
+				+ "}";
+	}
+
+	public static record Target<T>(TagGroup<T> group, Tag<T> tag) {
```
i509VCB, line 124: Could be made protected

library/data/tags/src/main/java/org/quiltmc/qsl/tag/mixin/RequiredTagListRegistryMixin.java
```diff
@@ -0,0 +1,46 @@
...
+@Mixin(RequiredTagListRegistry.class)
+public class RequiredTagListRegistryMixin {
+	@Shadow
+	@Final
+	private static List<RequiredTagList<?>> ALL;
+
+	@Inject(method = "getBuiltinTags", at = @At("TAIL"), cancellable = true)
+	private static void onGetBuiltinTags(CallbackInfoReturnable<Set<RequiredTagList<?>>> cir) {
+		// Add tag lists registered in QSL to the map.
+		var set = new ObjectOpenHashSet<>(cir.getReturnValue());
```
i509VCB: Immutable set in result I assume?

library/data/tags/src/main/java/org/quiltmc/qsl/tag/mixin/client/ClientPlayNetworkHandlerMixin.java
```diff
@@ -0,0 +1,78 @@
...
+					shift = At.Shift.AFTER
+			)
+	)
+	private void onGameJoin(GameJoinS2CPacket packet, CallbackInfo ci) {
+		ClientTagRegistryManager.applyAll(packet.getRegistryManager());
+	}
+
+	@Inject(method = "onDisconnected", at = @At("TAIL"))
+	private void onDisconnected(Text reason, CallbackInfo ci) {
+		// @TODO Replace with networking API?
```
i509VCB, line 52: Make an issue on this todo after this api is merged.

library/data/tags/src/testmod/java/org/quiltmc/qsl/tag/test/client/ClientTagsTestMod.java
```diff
@@ -0,0 +1,87 @@
...
+import org.quiltmc.qsl.tag.test.TagsTestMod;
+
+public final class ClientTagsTestMod implements ClientModInitializer {
+	/**
+	 * This tag is only registered on the client:
+	 * - worlds opened by the client must have this tag to work, only included in the quilt_tags_testmod:required_test_pack
+	 * data-pack so the testing can be exactly sure.
+	 * - can connect to servers that don't have the tag (since it's not TagType.CLIENT_SERVER_REQUIRED)
+	 */
+	public static final Tag<Block> TEST_REQUIRED_BLOCK_TAG = TagRegistry.BLOCK.create(
```
i509VCB: There a reason this tag is not identified like the others?

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/QuiltTag.java
LambdAurora: Not entirely sure, my big issue with that is static helper functions are way too disconnected on an usage standpoint.
What I could do is force this interface to be implemented with default methods on the Tag interface so no cast exceptions.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/impl/TagDelegate.java
LambdAurora: Cannot be made protected somehow because of `ClientTagDelegate`.

library/data/tags/src/testmod/java/org/quiltmc/qsl/tag/test/client/ClientTagsTestMod.java
LambdAurora: Not sure what you mean?
The tag seems to be identified like the others.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/TagType.java
LambdAurora: Yes, now documented.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/TagType.java
LambdAurora: - `SERVER_REQUIRED` means it's a tag that is only required for a server to start (or load a singleplayer world), 
- `CLIENT_SERVER_REQUIRED` means that in addition to being required for a server to start, the client also must receive that tag from the server when connecting (default behavior of most tags in MC),
- `NEUTRAL` means it's a tag that will be loaded and synced, but will not be required for server start or for client connection, if the server doesn't have it, the client won't have it either. This is the current behavior in Fabric API.
- `CLIENT_SERVER_SYNC` means it's mostly a client tag, not required to start nor connect, if the server has the tag it will sync it to the client, otherwise the client will use the defaults (loading from the data directory still)
- `CLIENT_ONLY` means the tag only exists on the client, located in the `assets` directory instead.

build-logic/src/main/java/qsl/internal/extension/QslModuleExtension.java
LambdAurora: Fixed.

library/data/tags/src/testmod/java/org/quiltmc/qsl/tag/test/client/ClientTagsTestMod.java
Juuxel: I think i5 means the field type (the others are `Tag.Identified`)

LambdAurora: Oh, that's because I don't display its ID and content at world change since this one can be observed directly with F3.

so I don't need the restricted type, and can just get the most generic type.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/package-info.java
```diff
@@ -0,0 +1,29 @@
...
+ * <h2>The Quilt Tags API.</h2>
+ *
+ * <p>
+ * <h3>What are tags?</h3>
+ * Tags are a way to have un-hardcoded list of registry entries since the content of tags are controlled by data packs..
```
Juuxel:
```suggestion
 * Tags are a way to have un-hardcoded list of registry entries since the content of tags are controlled by data packs.
```

library/data/tags/src/main/java/org/quiltmc/qsl/tag/mixin/RequiredTagListRegistryMixin.java
```diff
@@ -0,0 +1,47 @@
...
+import net.minecraft.tag.RequiredTagList;
+import net.minecraft.tag.RequiredTagListRegistry;
+
+@Mixin(RequiredTagListRegistry.class)
+public class RequiredTagListRegistryMixin {
+	@Shadow
+	@Final
+	private static List<RequiredTagList<?>> ALL;
+
+	@Inject(method = "getBuiltinTags", at = @At("TAIL"), cancellable = true)
```
OroArmor, line 39: `TAIL` or `RETURN` for this method? If it has any branch returns, this might produce inconsistent behaviour

OroArmor (approved):
Looks good. Test mod is good. Not huge, but you might want to use `RETURN` instead of `TAIL` in some mixins, but I haven't checked the Minecraft code.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/mixin/RequiredTagListRegistryMixin.java
LambdAurora: There's no branches in the original method.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/QuiltTag.java
```diff
@@ -0,0 +1,44 @@
...
+	 * Gets the type of this tag. The type will define how this tag is handled for syncing, requirement, etc.
+	 *
+	 * @return the type of this tag
+	 */
+	TagType getType();
+
+	/**
+	 * {@return {@code true} if the given tag has been "replaced" by a data pack at least once}
+	 */
+	boolean hasBeenReplaced();
```
Juuxel, line 45: What's the use case for this? Although I recall Fabric also has this weird API...

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/QuiltTagBuilder.java
```diff
@@ -0,0 +1,39 @@
...
+	/**
+	 * Clears the contained entries and mark the tag as replaced.
+	 */
+	void clearEntries();
```
Juuxel: Would returning `QuiltTagBuilder` here be better? Creating a local for a simple builder call (or using `Util.make` for it) feels a bit odd to me.

Juuxel (approved):
I really like the client tags and different tag types (including required tags especially - even though they're a vanilla feature, some APIs like FAPI and Architectury API only expose optionals).

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/QuiltTagBuilder.java
LambdAurora: Good call, will do that.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/QuiltTag.java
LambdAurora: I guess it can be for debugging?

I don't know all the details honestly, would need to find which PR added this to FAPI.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/package-info.java
```diff
@@ -0,0 +1,29 @@
+/**
+ * <h2>The Quilt Tags API.</h2>
+ *
+ * <p>
+ * <h3>What are tags?</h3>
+ * Tags are a way to have un-hardcoded list of registry entries since the content of tags are controlled by data packs.
+ * For example, Minecraft use those to identify all stairs blocks in the game, which is used for the connection mechanic.
+ * <p>
+ * By default tags are available for most of the static registries and are one of the earliest resource to be loaded on
```
BasiqueEvangelist:
```suggestion
 * By default tags are available for most of the static registries and are one of the earliest resources to be loaded on
```

library/data/tags/src/main/java/org/quiltmc/qsl/tag/impl/TagDelegate.java
```diff
@@ -0,0 +1,125 @@
...
+import net.minecraft.tag.Tag;
+import net.minecraft.tag.TagGroup;
+import net.minecraft.util.Identifier;
+
+import org.quiltmc.qsl.tag.api.QuiltTag;
+import org.quiltmc.qsl.tag.api.TagType;
+
+/**
+ * Represents a delegated tag. A delegated tag is a wrapper around a tag which is retrieved at runtime,
+ * and is kept update throughout the lifecycle of the game.
```
BasiqueEvangelist:
```suggestion
 * and is kept up to date throughout the lifecycle of the game.
```

library/data/tags/src/main/java/org/quiltmc/qsl/tag/impl/TagRegistryImpl.java
```diff
@@ -0,0 +1,188 @@
...
+	 *
+	 * @param registryKey the key of the registry
+	 * @return {@code true} if the registry is dynamic, otherwise {@code false}
+	 */
+	public static boolean isRegistryDynamic(RegistryKey<? extends Registry<?>> registryKey) {
+		return DynamicRegistryManagerAccessor.getInfos().containsKey(registryKey);
+	}
+
+	public static void forceInit() {
+		// noop. this only forces to run the static initializer.
```
BasiqueEvangelist, line 164: Why do you need to force running clinit? Isn't only the ThreadLocal and Logger being initialized?

library/data/tags/src/main/java/org/quiltmc/qsl/tag/impl/TagRegistryImpl.java
```diff
@@ -0,0 +1,188 @@
...
+	@Override
+	public Tag.Identified<T> create(Identifier id, TagType type) {
+		return switch (type) {
+			case SERVER_REQUIRED, CLIENT_SERVER_REQUIRED -> {
+				for (var requiredTagList : RequiredTagListRegistryAccessor.getAll()) {
+					if (requiredTagList.getRegistryKey() == this.registryKey) {
+						//noinspection unchecked
+						yield ((QuiltRequiredTagListHooks<T>) requiredTagList).quilt$addTypedTag(id, type);
+					}
+				}
```
BasiqueEvangelist: Why isn't the tag list stored in the tag registry?

library/data/tags/src/main/java/org/quiltmc/qsl/tag/impl/client/ClientTagRegistryManager.java
```diff
@@ -0,0 +1,217 @@
...
+	public static void applyAll(DynamicRegistryManager registryManager) {
+		TAG_GROUP_MANAGERS.forEach((registryKey, manager) -> manager.apply(registryManager));
+	}
+
+	/**
+	 * Represents a registry content fetcher.
+	 * <p>
+	 * But this fetcher will fetch the dynamic registry manager and auto-updates its references to the under-laying registry
+	 * used for content fetching.
+	 */
```
BasiqueEvangelist:
```suggestion
	/**
	 * Represents a registry content fetcher.
	 * <p>
	 * This fetcher will auto-update the reference to the underlying registry whenever the dynamic registry manager changes.
	 */
```

library/data/tags/src/main/java/org/quiltmc/qsl/tag/impl/QuiltTagHooks.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package org.quiltmc.qsl.tag.impl;
+
+import org.jetbrains.annotations.ApiStatus;
+
+@ApiStatus.Internal
+public interface QuiltTagHooks {
+	void quilt$setClearCount(int clearCount);
```
BasiqueEvangelist: Is the actual clear count used anywhere? (excluding just getting if the tag is cleared)
If it isn't, it would probably be better to replace it with just a boolean property.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/mixin/RequiredTagListMixin.java
```diff
@@ -0,0 +1,76 @@
...
+		if (TagRegistryImpl.isClientFetchingMissingTags()) {
+			return list.stream().filter(wrapper -> QuiltTag.cast(wrapper).getType().isRequiredToConnect());
+		} else {
+			return list.stream();
+		}
+	}
+
+	@Mixin(RequiredTagList.TagWrapper.class)
+	public abstract static class TagWrapperMixin<T> implements QuiltTag<T>, QuiltRequiredTagWrapperHooks {
+		private TagType tagType = TagType.CLIENT_SERVER_REQUIRED;
```
BasiqueEvangelist: Why isn't this field `@Unique`?

library/data/tags/src/main/java/org/quiltmc/qsl/tag/mixin/TagBuilderMixin.java
```diff
@@ -0,0 +1,70 @@
...
+			at = @At(
+					value = "INVOKE",
+					target = "Lcom/mojang/datafixers/util/Either;right(Ljava/lang/Object;)Lcom/mojang/datafixers/util/Either;",
+					remap = false
+			)
+	)
+	private Either<Collection<Tag.TrackedEntry>, Object> build(Object tag) {
+		((QuiltTagHooks) tag).quilt$setClearCount(this.quilt$clearCount);
+		return Either.right(tag);
+	}
```
BasiqueEvangelist: Can't you use a `@ModifyArg` here?

library/data/tags/src/testmod/java/org/quiltmc/qsl/tag/test/client/ClientTagsTestMod.java
BasiqueEvangelist: Wouldn't it be better to make it also identified just to be consistent with the other tags?

library/data/tags/src/main/java/org/quiltmc/qsl/tag/impl/TagRegistryImpl.java
LambdAurora: If this isn't ran when it forced, it will cause recursion and very bad crashes.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/package-info.java
```diff
@@ -0,0 +1,29 @@
+/**
+ * <h2>The Quilt Tags API.</h2>
+ *
+ * <p>
+ * <h3>What are tags?</h3>
+ * Tags are a way to have un-hardcoded list of registry entries since the content of tags are controlled by data packs.
+ * For example, Minecraft use those to identify all stairs blocks in the game, which is used for the connection mechanic.
```
IotaBread: Isn't the connection mechanic related to the block state? I don't think it's related to tags at all

IotaBread (approved):

library/data/tags/src/testmod/java/org/quiltmc/qsl/tag/test/client/ClientTagsTestMod.java
LambdAurora: It's identified now.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/mixin/RequiredTagListMixin.java
LambdAurora: It is now.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/mixin/TagBuilderMixin.java
LambdAurora: I can! fixed.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/package-info.java
LambdAurora: it was fences, damn it.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/QuiltTag.java
LambdAurora: Actually found why: FabricMC/fabric#63, original reason sounds sensible.

Edit: asie did explained the entire reason why and the use case: https://github.com/FabricMC/fabric/issues/579#issuecomment-615237753

I think we should keep it then.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/impl/TagRegistryImpl.java
LambdAurora: It was an artifact of me trying to design the implementation, latest commit makes it store the tag list instead.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/TagRegistry.java
```diff
@@ -0,0 +1,89 @@
...
+import net.minecraft.util.registry.RegistryKey;
+import net.minecraft.world.biome.Biome;
+import net.minecraft.world.event.GameEvent;
+
+import org.quiltmc.qsl.tag.impl.TagRegistryImpl;
+
+/**
+ * Represents a tag registry.
+ *
+ * @param <T> the type of the registered tags
```
Juuxel:
```suggestion
 * @param <T> the type of the content in the registered tags
```

Juuxel (approved):

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/TagType.java
i509VCB: `SERVER_REQUIRED` is fine.

`CLIENT_SERVER_REQUIRED` -> `REQUIRED`. The game will not function if the server and client do not have an instance of the tag. `REQUIRED` communicates this idea succinctly.

`NEUTRAL` -> `OPTIONAL`. The client or server does not require a tag. The server is the authoritative source of the tag

`CLIENT_SERVER_SYNC` -> `CLIENT_FALLBACK`. Think of it like an `OPTIONAL` that uses the client's tag as a fallback. The server is considered to be authoritative with it's own tag list.

`CLIENT_ONLY` is fine.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/QuiltTag.java
```diff
@@ -0,0 +1,46 @@
...
+
+	/**
+	 * {@return {@code true} if the given tag has been "replaced" by a data pack at least once}
+	 * <p>
+	 * The use case for such method is to be able to provide vanilla behaviour compatibility in a hook,
+	 * but also allow overriding it for "total conversion" data-packs.
+	 */
+	boolean hasBeenReplaced();
+
+	static <T> QuiltTag<T> cast(Tag<T> tag) {
```
i509VCB: We should probably name this something that indicates we are getting the tag extensions.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/QuiltTag.java
i509VCB: Okay a compromise:

1. Note in the docs that casting to `QuiltTag` is a usage error. Encourage using the `cast` function as it is currently named to get the functionality provided by this interface.

2. Implementations of `Tag` may implement this interface. A default implementation is provided.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/TagRegistry.java
```diff
@@ -0,0 +1,89 @@
...
+
+import org.quiltmc.qsl.tag.impl.TagRegistryImpl;
+
+/**
+ * Represents a tag registry.
+ *
+ * @param <T> the type of the content in the registered tags
+ */
+@ApiStatus.NonExtendable
+public interface TagRegistry<T> {
```
i509VCB, line 49: I feel it would be useful to have a way to iterate over the tags in a tag registry. Or where ever it is applicable.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/TagRegistry.java
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	/**
+	 * Creates a new identified tag.
+	 *
+	 * @param id   the identifier of the tag
+	 * @param type the type of the tag
+	 * @return the identified tag
+	 */
+	Tag.Identified<T> create(Identifier id, TagType type);
```
i509VCB: What is the expected behavior if a tag of the specified id already exists?

library/data/tags/src/main/java/org/quiltmc/qsl/tag/impl/TagRegistryImpl.java
```diff
@@ -0,0 +1,184 @@
...
+import org.quiltmc.qsl.tag.api.TagType;
+import org.quiltmc.qsl.tag.impl.client.ClientTagRegistryManager;
+import org.quiltmc.qsl.tag.mixin.DynamicRegistryManagerAccessor;
+import org.quiltmc.qsl.tag.mixin.RequiredTagListRegistryAccessor;
+
+@SuppressWarnings("ClassCanBeRecord")
+@ApiStatus.Internal
+public final class TagRegistryImpl<T> implements TagRegistry<T> {
+	private static final Logger LOGGER = LogManager.getLogger();
+	private static final ThreadLocal<Boolean> MISSING_TAGS_CLIENT_FETCH = new ThreadLocal<>();
```
i509VCB, line 55: Is a thread local an error here? My interpretation of the uses is that the integrated server would never see the client thread signal it that it is fetching tags because the signal would never signal the server's thread.

If it is an error, would `AtomicBoolean` suffice.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/impl/TagRegistryImpl.java
```diff
@@ -0,0 +1,184 @@
...
+
+	@SuppressWarnings("unchecked")
+	public static <T> TagRegistry<T> of(RegistryKey<? extends Registry<T>> registryKey, String dataType) {
+		RequiredTagList<T> tagList = null;
+
+		// Use already registered tag list for the registry if it has the same data type, in case multiple mods tried to do it.
+		for (var value : RequiredTagListRegistryAccessor.getAll()) {
+			if (value.getRegistryKey() == registryKey) {
+				tagList = (RequiredTagList<T>) value;
+				// Throw an exception if the tag list has a different data type.
```
i509VCB, line 92: This behavior is not mentioned in the `TagRegistry` docs

library/data/tags/src/main/java/org/quiltmc/qsl/tag/impl/TagRegistryImpl.java
```diff
@@ -0,0 +1,184 @@
...
+		return this.tagList.getRegistryKey();
+	}
+
+	@Override
+	public Tag.Identified<T> create(Identifier id, TagType type) {
+		return switch (type) {
+			case SERVER_REQUIRED, CLIENT_SERVER_REQUIRED -> //noinspection unchecked
+					((QuiltRequiredTagListHooks<T>) this.tagList).quilt$addTypedTag(id, type);
+			case CLIENT_SERVER_SYNC, CLIENT_ONLY -> ClientTagRegistryManager.create(id, type, this.getRegistryKey(), this.tagGroupSupplier);
+			default -> new TagDelegate<>(id, type, this.tagGroupSupplier);
```
i509VCB: this will allow `null` as a valid tag type value.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/mixin/TagMixin.java
```diff
@@ -0,0 +1,38 @@
...
+@Mixin(Tag.class)
+public interface TagMixin<T> extends QuiltTag<T> {
+	@Override
+	default TagType getType() {
+		return TagType.NEUTRAL;
+	}
+
+	@Override
+	default boolean hasBeenReplaced() {
+		return false; // The goal is to prevent hard-fails with custom Tag implementations.
```
i509VCB, line 35: I'm not sure we can prove beyond a reasonable doubt that any implementation of `Tag` we or anyone else does not explicitly handle may replace it's contents or has had it's contents replaced.

library/data/tags/src/testmod/java/org/quiltmc/qsl/tag/test/client/ClientTagsTestMod.java
```diff
@@ -0,0 +1,87 @@
...
+import org.quiltmc.qsl.tag.test.TagsTestMod;
+
+public final class ClientTagsTestMod implements ClientModInitializer {
+	/**
+	 * This tag is only registered on the client:
+	 * - worlds opened by the client must have this tag to work, only included in the quilt_tags_testmod:required_test_pack
+	 * data-pack so the testing can be exactly sure.
+	 * - can connect to servers that don't have the tag (since it's not TagType.CLIENT_SERVER_REQUIRED)
+	 */
+	public static final Tag.Identified<Block> TEST_REQUIRED_BLOCK_TAG = TagRegistry.BLOCK.create(
```
i509VCB: Probably a good idea to assert the server required tag exists in a server test mod.

i509VCB (changes requested):

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/QuiltTag.java
i509VCB: I think it would be useful to mention an example of why this would be used similar to the fabric api comment.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/TagRegistry.java
LambdAurora: It will create a new `Tag` object whatever happens, which means if you register one for the `OPTIONAL` TagType, and `SERVER_REQUIRED`, you will end up with it required for server start (all required tag types will introduce this behavior)

If you create any other one and a `CLIENT_ONLY` they will be treated as extremely different tags since `CLIENT_ONLY` loads from the assets directory.

Any of the tags created with a sync feature will sync.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/impl/TagRegistryImpl.java
LambdAurora: This boolean is used exclusively on the client to check whether there is missing tags that should prevent the connection to the server, the entire goal of this ThreadLocal is to change the behavior of `getMissingTags` to exclude any tags not required to connect in `onSynchronizeTags(SynchronizeTagsS2CPacket)`.

`AtomicBoolean` could introduce unwanted behavior on the integrated server.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/mixin/TagMixin.java
LambdAurora: We can't, but not providing a default implementation could lead to bad crashes quickly.

So either we force the implementation of `QuiltTag`, or we encourage it without forcing.

library/data/tags/src/main/java/org/quiltmc/qsl/tag/api/TagRegistry.java
LambdAurora: Since the `TagRegistry` does not keep track of registered tags, that feels a bit impossible?

library/data/tags/src/testmod/java/org/quiltmc/qsl/tag/test/client/ClientTagsTestMod.java
LambdAurora: Makes it a bit difficult to test this since it uses a datapack, could make it enabled by default on servers though.

LambdAurora:
Since we're a bit short on time, and all reviews should have been addressed, merging it.

If any issue occurs they always can be corrected in a later PR.

PR #42, opened 2021-10-17
Entity Events module

williambl:
This PR adds:
 - A new **library**, `entity` (as described in [RFC #9](https://github.com/QuiltMC/rfcs/blob/master/rfc/0009-qsl-structure.md))
 - A new **module**, `entity_events`
 - A set of events in that module.
 - A test mod for that module.

The events included are:
 - `EntityKilledCallback`, with similar purpose to Fabric's `AfterKilledOtherEntity`.
 - `TryReviveCallback` - the `AllowDeathEvent` I wrote for FAPI was merged in a state I was unhappy with. This is closer to how I imagined it.
 - `EntityWorldChangedEvents` - unchanged from FAPI's `ServerEntityWorldChangedEvents`, other than name.
 - `EntityLoadEvents`, two events for when entities are loaded and unloaded on the server.
 - `ClientEntityLoadEvents`, ditto but for the client.
 - `ServerPlayerEntityCopyCallback`, for copying data after player respawn.

There is certainly space for more events.


library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityKilledCallback.java
```diff
@@ -0,0 +1,32 @@
...
+
+
+	/**
+	 * Called when an entity is killed by another entity.
+	 *
+	 * @param world the world
+	 * @param killer the killing entity, possibly null
+	 * @param killed the entity which was killed by the {@code killer}
+	 */
+	void onKilled(World world, @Nullable Entity killer, LivingEntity killed);
```
OroArmor: maybe a damage source for this?

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityLoadEvents.java
```diff
@@ -0,0 +1,68 @@
...
+	 * An event which is called after an entity has been loaded into a world.
+	 */
+	public static final ArrayEvent<AfterEntityLoad> AFTER_ENTITY_LOAD = ArrayEvent.create(AfterEntityLoad.class, callbacks -> (entity, world) -> {
+		for (AfterEntityLoad callback : callbacks) {
+			callback.onLoad(entity, world);
+		}
+	});
+
+	/**
+	 * An event which is called after a player has been unloaded from a world.
```
OroArmor: player here, entity above

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityTickCallback.java
```diff
@@ -0,0 +1,33 @@
...
+import net.minecraft.entity.Entity;
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.entity.damage.DamageSource;
+import org.quiltmc.qsl.base.api.event.ArrayEvent;
+
+/**
+ * A callback that is invoked when an Entity is ticked (nominally every 1/20 of a second).
+ *
+ * <p>There are two types of entity tick - standalone ({@link Entity#tick()}) and riding ({@link Entity#tickRiding()}).
+ * This callback takes a parameter which
```
OroArmor: unfinished sentence

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/TryReviveCallback.java
```diff
@@ -0,0 +1,64 @@
...
+ *     <li>a totem-of-undying mod giving the player a little health + regeneration effect</li>
+ *     <li>a minigame mod teleporting the player into a 'respawn room' and setting their health to 20.0</li>
+ *     <li>a mod that changes death mechanics switching the player over to the mod's play-mode, where death doesn't
+ *     apply</li>
+ * </ul>
+ *
+ * <p>This event is short-circuiting - if the entity has already been revived, your callback will not fire.
+ */
+@FunctionalInterface
+public interface TryReviveCallback {
```
OroArmor: does this run before `onKilled`?

OroArmor (changes requested):

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityTickCallback.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+@FunctionalInterface
+public interface EntityTickCallback {
```
i509VCB: This event deserves it's own pr as it is more complex than it seems

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityLoadEvents.java
```diff
@@ -0,0 +1,68 @@
...
+package org.quiltmc.qsl.entity.api.event;
+
+import net.minecraft.entity.Entity;
+import net.minecraft.world.World;
+import org.quiltmc.qsl.base.api.event.ArrayEvent;
+
+/**
+ * Events related to an entity being loaded into or unloaded from a world.
+ */
+public final class EntityLoadEvents {
```
i509VCB: I recall upstream my addition keeps client and server loading separate to be more explicit.

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/TryReviveCallback.java
```diff
@@ -0,0 +1,64 @@
...
+
+	/**
+	 * Called before totems try to revive the player.
+	 */
+	ArrayEvent<TryReviveCallback> BEFORE_TOTEM = ArrayEvent.create(TryReviveCallback.class, callbacks -> (entity, damageSource) -> {
```
i509VCB: Make another callback interface so before and after have their own.

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityTickCallback.java
williambl: not to go all sentence fragment on you but

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityLoadEvents.java
williambl: Oh - I didn't realise there were events for this upstream, I was only looking in the entity events module (I assumed that my memory of entity load events existing was a false one since I couldn't see them in there).

Do we want to keep this explicitness, seeing as a `world.isClient` check in the API user's code can accomplish the same thing? There may also be cases in which the same code wants to be run on entity load for both sides, which this structure makes easier.

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/TryReviveCallback.java
williambl: Will do.

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityLoadEvents.java
i509VCB: The primary reason for keeping the two events seperate is to prevent cases where mod developers load some client only objects or classes.

I would say yes to being explicit there

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityKilledCallback.java
williambl: The damage source would have to be acquired via the `getRecentDamageSource` method on the killed entity, and is therefore already accessible by user code. I can see the convenience of such a parameter, though, especially if an API user does not know about the method. Perhaps a note in the javadoc could be useful instead?

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityLoadEvents.java
williambl: Whoops, thanks.

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityLoadEvents.java
williambl: Ah, that makes sense. I'll split them.

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/TryReviveCallback.java
williambl: Yes.

williambl:
marking as ready for review just for hacktoberfest, sorry :(

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityKilledCallback.java
OroArmor: Yes, a javadoc note would be nice. I did not know that existed.

BasiqueEvangelist:
Could you add events for copying player data on respawn and travel from end?

williambl:
Going to rebase onto 1.18 and force push.

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityKilledCallback.java
williambl: Added.

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityTickCallback.java
williambl: Removed.

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityLoadEvents.java
williambl: Split.

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityLoadEvents.java
```diff
@@ -0,0 +1,110 @@
...
```
i509VCB: Please declare the two sides in their own classes.

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityLoadEvents.java
```diff
@@ -0,0 +1,110 @@
...
+	public static final ArrayEvent<AfterEntityLoadServer> AFTER_ENTITY_LOAD_SERVER = ArrayEvent.create(AfterEntityLoadServer.class, callbacks -> (entity, world) -> {
+		for (AfterEntityLoadServer callback : callbacks) {
+			callback.onLoad(entity, world);
+		}
+	});
+
+	/**
+	 * An event which is called after an entity has been loaded into a client world.
+	 */
+	public static final ArrayEvent<AfterEntityLoadClient> AFTER_ENTITY_LOAD_CLIENT = ArrayEvent.create(AfterEntityLoadClient.class, callbacks -> (entity, world) -> {
```
i509VCB: Ensure the owning class is client only.

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityLoadEvents.java
```diff
@@ -0,0 +1,110 @@
...
+		 *
+		 * @param entity the entity which has been unloaded
+		 * @param world the world the entity has been unloaded from
+		 */
+		void onUnload(Entity entity, World world);
+	}
+
+
+	@FunctionalInterface
+	public interface AfterEntityLoadClient {
```
i509VCB: This interface should be marked client only.

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityLoadEvents.java
```diff
@@ -0,0 +1,110 @@
...
+
+	@FunctionalInterface
+	public interface AfterEntityLoadServer {
+		/**
+		 * Called after an entity has loaded into a server world.
+		 *
+		 * @param entity the entity which has been loaded
+		 * @param world the world the entity has been loaded into
+		 */
+		void onLoad(Entity entity, World world);
```
i509VCB: Please use the more explicit ServerWorld/ClientWorld where applicable.

Currently both interfaces define a function of the same signature, that will cause conflicts.

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityReviveEvents.java
```diff
@@ -0,0 +1,97 @@
...
+	@FunctionalInterface
+	public interface TryReviveAfterTotem {
+		/**
+		 * Whether an entity which has fatal damage and which has not been saved by a totem should be revived.
+		 *
+		 * @param entity       the entity
+		 * @param damageSource the fatal damage source
+		 * @return true if the entity should be revived, false otherwise.
+		 */
+		boolean tryReviveAfterTotem(LivingEntity entity, DamageSource damageSource);
```
i509VCB: Function names need to differ as these signatures will conflict.

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityReviveEvents.java
```diff
@@ -0,0 +1,97 @@
...
+			if (callback.tryReviveAfterTotem(entity, damageSource)) {
+				return true;
+			}
+		}
+
+		return false;
+	});
+
+	@FunctionalInterface
+	public interface TryReviveBeforeTotem {
```
i509VCB: I think we could find a more succinct name here.

i509VCB (dismissed):

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityLoadEvents.java
williambl: If the outer class is marked client-only, is that enough?

LambdAurora: If the outer class is marked client-only, the entire outer class and its content will be stripped in server envs.

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityReviveEvents.java
williambl: I'm assuming this comment was supposed to be on something else? Nothing else is named tryReviveAfterTotem.

williambl: (I'm assuming After/Before EntityLoadClient - changed now)

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityReviveEvents.java
williambl: I'm open for suggestions, if there's anything which is shorter but still gets through the same amount of information.

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityLoadEvents.java
williambl: I wasn't sure if an inner interface/static class counted as being 'in' an outer class for that purpose (since it's in a different class file when compiled)

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityReviveEvents.java
williambl: (Although, those wouldn't conflict either, after changing the world argument types)

williambl:
Not sure why checks are failing, they're passing locally

williambl:
@BasiqueEvangelist added ServerPlayerEntityCopyCallback (name subject to change because that's quite long) - does this fit what you'd need?

BasiqueEvangelist:
> @BasiqueEvangelist added ServerPlayerEntityCopyCallback (name subject to change because that's quite long) - does this fit what you'd need?

This seems to fit all common use cases, so yes.

BasiqueEvangelist (approved):

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/ServerPlayerEntityCopyCallback.java
```diff
@@ -0,0 +1,36 @@
...
+
+
+	/**
+	 * Called when a player is copied.
+	 *
+	 * @param newPlayer the new ServerPlayerEntity instance
+	 * @param original 	the original ServerPlayerEntity instance
+	 * @param wasDeath 	true if the copying is due to the player dying, false otherwise
+	 */
+	void onPlayerCopy(ServerPlayerEntity newPlayer, ServerPlayerEntity original, boolean wasDeath);
```
lenrik1589: maybe call *copied* instance of player `copy` instead of `newPlayer` as the player itself isn't new?

TheGlitch76:
is this API still WIP?

williambl:
> is this API still WIP?

There are a couple more events to be added, but I think I'd be happy to merge these existing ones as-is. (with the caveat that I'd like someone to have a look at the the tryrevive events and see if they can be done with the event phase system, which I'm having trouble wrapping my head around)

library/entity/entity-events/build.gradle
```diff
@@ -0,0 +1,12 @@
+plugins {
+	id("qsl.module")
+}
+
+qslModule {
+	moduleName = "entity-events"
```
LambdAurora:
```suggestion
	moduleName = "entity_events"
```

QSL libraries and modules are snake_case.

library/entity/entity-events/build.gradle
```diff
@@ -0,0 +1,12 @@
...
+	version = "1.0.0"
+	library = "entity"
+	coreDependencies([
```
LambdAurora: Please merge upstream, this is not a valid way to declare dependencies anymore.

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityKilledCallback.java
```diff
@@ -0,0 +1,51 @@
...
+ * <p>The killed entity's {@link LivingEntity#getRecentDamageSource()} method can be used to retrieve the damage source
+ * which may have killed it.
+ */
+@FunctionalInterface
+public interface EntityKilledCallback {
+	/**
+	 * Invoked when an entity is killed by another entity.
+	 */
+	Event<EntityKilledCallback> EVENT = Event.create(EntityKilledCallback.class, callbacks -> (world, killer, killed) -> {
+		for (EntityKilledCallback callback : callbacks) {
```
LambdAurora:
```suggestion
		for (var callback : callbacks) {
```

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityReviveEvents.java
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * <p>These events are short-circuiting - if the entity has already been revived, your callback will not fire.
+ */
+public final class EntityReviveEvents {
+
+	/**
+	 * Called before totems try to revive the player.
+	 */
+	public static Event<TryReviveBeforeTotem> BEFORE_TOTEM = Event.create(TryReviveBeforeTotem.class, callbacks -> (entity, damageSource) -> {
+		for (TryReviveBeforeTotem callback : callbacks) {
```
LambdAurora:
```suggestion
		for (var callback : callbacks) {
```

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityReviveEvents.java
```diff
@@ -0,0 +1,97 @@
...
+		}
+
+		return false;
+	});
+
+	/**
+	 * Called after there has been an attempt to revive with totems, but it has not been successful.
+	 */
+	public static Event<TryReviveAfterTotem> AFTER_TOTEM = Event.create(TryReviveAfterTotem.class, callbacks -> (entity, damageSource) -> {
+		for (TryReviveAfterTotem callback : callbacks) {
```
LambdAurora:
```suggestion
		for (var callback : callbacks) {
```

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityWorldChangeEvents.java
```diff
@@ -0,0 +1,89 @@
...
+	 *
+	 * <p>All entities are copied to the destination and the old entity removed.
+	 * This event does not apply to the {@link ServerPlayerEntity} since players are physically moved to the new world instead of being copied over.
+	 *
+	 * <p>A mod may use this event for reference cleanup if it is tracking an entity's current world.
+	 *
+	 * @see EntityWorldChangeEvents#AFTER_PLAYER_CHANGE_WORLD
+	 */
+	public static final Event<AfterEntityChange> AFTER_ENTITY_CHANGE_WORLD = Event.create(AfterEntityChange.class, callbacks -> (original, copy, origin, destination) -> {
+		for (AfterEntityChange callback : callbacks) {
```
LambdAurora:
```suggestion
		for (var callback : callbacks) {
```

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/EntityWorldChangeEvents.java
```diff
@@ -0,0 +1,89 @@
...
+	/**
+	 * An event which is invoked server-side after a player has been moved to a different world.
+	 *
+	 * <p>This is similar to {@link EntityWorldChangeEvents#AFTER_ENTITY_CHANGE_WORLD} but is only called for players.
+	 * This is because the player is physically moved to the new world instead of being recreated at the destination.
+	 *
+	 * @see EntityWorldChangeEvents#AFTER_ENTITY_CHANGE_WORLD
+	 */
+	public static final Event<AfterPlayerChange> AFTER_PLAYER_CHANGE_WORLD = Event.create(AfterPlayerChange.class, callbacks -> (player, origin, destination) -> {
+		for (AfterPlayerChange callback : callbacks) {
```
LambdAurora:
```suggestion
		for (var callback : callbacks) {
```

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/ServerEntityLoadEvents.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Events related to an entity being loaded into or unloaded from a server world.
+ */
+public final class ServerEntityLoadEvents {
+	/**
+	 * An event which is called after an entity has been loaded into a server world.
+	 */
+	public static final Event<AfterLoad> AFTER_LOAD = Event.create(AfterLoad.class, callbacks -> (entity, world) -> {
+		for (AfterLoad callback : callbacks) {
```
LambdAurora:
```suggestion
		for (var callback : callbacks) {
```

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/ServerEntityLoadEvents.java
```diff
@@ -0,0 +1,69 @@
...
+			callback.onLoad(entity, world);
+		}
+	});
+
+	/**
+	 * An event which is called after an entity has been unloaded from a server world.
+	 */
+	public static final Event<AfterUnload> AFTER_UNLOAD = Event.create(AfterUnload.class, callbacks -> (entity, world) -> {
+		for (AfterUnload callback : callbacks) {
```
LambdAurora:
```suggestion
		for (var callback : callbacks) {
```

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/ServerEntityLoadEvents.java
```diff
@@ -0,0 +1,69 @@
...
+
+	@FunctionalInterface
+	public interface AfterLoad {
+		/**
+		 * Called after an entity has loaded into a server world.
+		 *
+		 * @param entity the entity which has been loaded
+		 * @param world the world the entity has been loaded into
+		 */
+		void onLoad(Entity entity, ServerWorld world);
```
LambdAurora: Those method names might want to follow a bit more the callback name?
If a before callback is added afterwards the method names will be quite inconsistent.

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/client/ClientEntityLoadEvents.java
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Events related to an entity being loaded into or unloaded from a client world.
+ */
+@Environment(EnvType.CLIENT)
+public final class ClientEntityLoadEvents {
+	/**
+	 * An event which is called after an entity has been loaded into a client world.
+	 */
+	public static final Event<AfterLoad> AFTER_LOAD = Event.create(AfterLoad.class, callbacks -> (entity, world) -> {
+		for (AfterLoad callback : callbacks) {
```
LambdAurora:
```suggestion
		for (var callback : callbacks) {
```

library/entity/entity-events/src/main/java/org/quiltmc/qsl/entity/api/event/client/ClientEntityLoadEvents.java
```diff
@@ -0,0 +1,70 @@
...
+			callback.onLoadClient(entity, world);
+		}
+	});
+	/**
+	 * An event which is called after an entity has been unloaded from a client world.
+	 */
+	public static final Event<AfterUnload> AFTER_UNLOAD = Event.create(AfterUnload.class, callbacks -> (entity, world) -> {
+		for (AfterUnload callback : callbacks) {
```
LambdAurora:
```suggestion
		for (var callback : callbacks) {
```

library/entity/entity-events/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,24 @@
...
+  "icon": "assets/quilt_crash_info/icon.png",
+  "contact": {
+    "homepage": "https://quiltmc.org",
+    "issues": "https://github.com/QuiltMC/quilt-standard-libraries/issues",
+    "sources": "https://github.com/QuiltMC/quilt-standard-libraries"
+  },
+  "authors": [
+    "QuiltMC"
+  ],
+  "depends": {
```
LambdAurora: It depends on more than just loader: there's qsl base, etc.

library/entity/entity-events/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,24 @@
+{
+  "schemaVersion": 1,
+  "id": "quilt_entity_events",
+  "name": "Quilt Entity Events",
+  "version": "${version}",
+  "environment": "*",
+  "license": "Apache-2.0",
+  "icon": "assets/quilt_crash_info/icon.png",
```
LambdAurora:
```suggestion
  "icon": "assets/quilt_entity_events/icon.png",
```

library/entity/entity-events/src/testmod/java/org/quiltmc/qsl/entity/test/EntityEventsTestMod.java
```diff
@@ -0,0 +1,106 @@
...
+import net.minecraft.entity.passive.ChickenEntity;
+import net.minecraft.item.Items;
+import net.minecraft.text.LiteralText;
+import net.minecraft.util.registry.Registry;
+import net.minecraft.world.dimension.DimensionType;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.quiltmc.qsl.entity.api.event.*;
+
+public class EntityEventsTestMod implements ModInitializer {
```
LambdAurora: This needs to use the new Quilt entrypoints.

LambdAurora (changes requested):
Some of those events could be AwareListeners to allow them to be used in entrypoints.

Package also seems to be wrong: https://github.com/QuiltMC/quilt-standard-libraries/blob/1.18/CONTRIBUTING.md#packages

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/EntityKilledCallback.java
```diff
@@ -0,0 +1,52 @@
...
+import org.quiltmc.qsl.base.api.event.EventAwareListener;
+
+/**
+ * A callback which is invoked on the logical server when an entity is killed by another entity.
+ *
+ * <p>The killed entity's {@link LivingEntity#getRecentDamageSource()} method can be used to retrieve the damage source
+ * which may have killed it.
+ */
+@FunctionalInterface
+public interface EntityKilledCallback extends EventAwareListener {
```
TheGlitch76: Wouldn't this be more useful as a generic "Death" callback that might have an entity associated with it? The current javadoc implies this even should only be called when an entity is killed by another entity.

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/EntityReviveEvents.java
```diff
@@ -0,0 +1,98 @@
...
+			}
+		}
+
+		return false;
+	});
+
+	@FunctionalInterface
+	public interface TryReviveBeforeTotem extends EventAwareListener {
+		/**
+		 * Whether an entity which has fatal damage should be revived.
```
TheGlitch76: usually best to start w/ a verb
```suggestion
		 * Returns whether an entity which has fatal damage should be revived.
```

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/EntityReviveEvents.java
```diff
@@ -0,0 +1,98 @@
...
+		 * @return true if the entity should be revived, false otherwise.
+		 */
+		boolean tryReviveBeforeTotem(LivingEntity entity, DamageSource damageSource);
+	}
+
+
+	@FunctionalInterface
+	public interface TryReviveAfterTotem extends EventAwareListener {
+		/**
+		 * Whether an entity which has fatal damage and which has not been saved by a totem should be revived.
```
TheGlitch76:
```suggestion
		 * Returns whether an entity which has fatal damage and which has not been saved by a totem should be revived.
```

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/ServerPlayerEntityCopyCallback.java
```diff
@@ -0,0 +1,53 @@
...
+package org.quiltmc.qsl.entity_events.api;
+
+import net.minecraft.server.network.ServerPlayerEntity;
+import org.quiltmc.qsl.base.api.event.Event;
+import org.quiltmc.qsl.base.api.event.EventAwareListener;
+
+/**
+ * A callback which is called on the logical server when a player is copied.
+ *
+ * <p>Players are copied on death and when returning from the end. The {@code wasDeath} parameter can be used to
```
TheGlitch76: Does being copied from the end also count as a world change?

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/mixin/ServerPlayerEntityMixin.java
```diff
@@ -0,0 +1,52 @@
...
+
+@Mixin(ServerPlayerEntity.class)
+abstract class ServerPlayerEntityMixin extends PlayerEntity {
+	public ServerPlayerEntityMixin(World world, BlockPos blockPos, float f, GameProfile gameProfile) {
+		super(world, blockPos, f, gameProfile);
+	}
+
+	/**
+	 * This is called by both "moveToWorld" and "teleport".
+	 * So this is suitable to handle the after event from both call sites.
```
TheGlitch76:
```suggestion
	 * This is called by both "moveToWorld" and "teleport",
	 * so this is suitable to handle the after event from both call sites.
```

TheGlitch76 (dismissed):

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/ServerPlayerEntityCopyCallback.java
i509VCB: Depends. Using the portal means you get copied. If you use the teleport command in another dimension you get moved.

Reason being the player is recreated when using the portal is why this is seperate.

williambl:
rebasing onto 1.19

williambl:
unrebasing back onto 1.18, i wasn't thinking straight

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/EntityWorldChangeEvents.java
```diff
@@ -0,0 +1,91 @@
...
+		}
+	});
+
+	/**
+	 * An event which is invoked server-side after a player has been moved to a different world.
+	 *
+	 * <p>This is similar to {@link EntityWorldChangeEvents#AFTER_ENTITY_CHANGE_WORLD} but is only called for players.
+	 * This is because the player is physically moved to the new world instead of being recreated at the destination.
+	 *
+	 * @see EntityWorldChangeEvents#AFTER_ENTITY_CHANGE_WORLD
```
TropheusJ: I would add a mention of ServerPlayerEntityCopyCallback since it has a similar effect and mods may want to listen for both.

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/ServerPlayerEntityCopyCallback.java
```diff
@@ -0,0 +1,53 @@
...
+ * differentiate between the two situations.
+ *
+ * <p>The callback is called after vanilla has done its own copying logic.
+ *
+ * @see ServerPlayerEntity#copyFrom(ServerPlayerEntity, boolean)
```
TropheusJ, line 32: I would add a mention of `EntityWorldChangeEvents.AFTER_PLAYER_CHANGE_WORLD` since it has a similar effect and mods may want to listen for both.

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/mixin/LivingEntityMixin.java
```diff
@@ -0,0 +1,49 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+@Mixin(LivingEntity.class)
+public abstract class LivingEntityMixin {
+	@Inject(method = "onKilledBy", at = @At(value = "HEAD"))
+	private void onEntityKilledOther(LivingEntity adversary, CallbackInfo ci) {
+		EntityKilledCallback.EVENT.invoker().onKilled(((LivingEntity) (Object) this).world, adversary, (LivingEntity) (Object) this);
```
TropheusJ: mixin should extend Entity, would give access to `world` without a cast.

library/entity/entity_events/src/main/resources/quilt_entity_events.mixins.json
```diff
@@ -0,0 +1,18 @@
+{
+  "required": true,
+  "package": "org.quiltmc.qsl.entity_events.mixin",
+  "compatibilityLevel": "JAVA_16",
```
TropheusJ: `JAVA_17`

library/entity/entity_events/src/testmod/java/org/quiltmc/qsl/entity_events/test/EntityEventsTestMod.java
```diff
@@ -0,0 +1,129 @@
...
+
+	// Entities going to the end are named 'end traveller'
+	@Override
+	public void afterChangeWorld(Entity originalEntity, Entity newEntity, ServerWorld origin, ServerWorld destination) {
+		if (destination.getDimension() == destination.getServer().getRegistryManager().get(Registry.DIMENSION_TYPE_KEY).get(DimensionType.THE_END_REGISTRY_KEY)) {
+			newEntity.setCustomName(new LiteralText("End Traveller"));
+		}
+	}
+
+	// Players keep the glowing enchantment after respawn
```
TropheusJ: glowing effect

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/EntityKilledCallback.java
TropheusJ: agree, general death callback has more uses, and can be filtered by the listener to only listen for kills.

TropheusJ: agree, general death callback has more uses, and can be filtered by the listener to only listen for kills.

williambl: I didn't make it a death callback because I was under the false assumption that multiple entity methods did not run super.onDeath(), making it difficult to maintanably inject an event invokation. However, from having a look it seems that only ServerPlayerEntity does this, and two injection points sounds reasonable.

williambl: Done

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/mixin/LivingEntityDeathEventMixin.java
```diff
@@ -0,0 +1,27 @@
...
+import net.minecraft.entity.damage.DamageSource;
+import net.minecraft.server.network.ServerPlayerEntity;
+import net.minecraft.world.World;
+import org.quiltmc.qsl.entity_events.api.LivingEntityDeathCallback;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(value = { LivingEntity.class, ServerPlayerEntity.class })
```
williambl, line 33: I'd like someone to have a look at this - if it's not acceptable I'm fine with moving it into two separate mixins.

library/entity/entity_events/build.gradle
```diff
@@ -0,0 +1,17 @@
+plugins {
+	id("qsl.module")
+}
+
+qslModule {
+	name = "Quilt Entity Events"
+	moduleName = "entity_events"
+	id = "quilt_entity_events"
+	description = "Provides various events to do with entities"
+	version = "1.0.0"
```
TheGlitch76: Could you merge with upstream and then fix any problems with your build so that actions pass? Notably, this line will need to be removed.

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/mixin/LivingEntityDeathEventMixin.java
OroArmor: This is fine. Multiple targets are something useful

OroArmor (approved):

williambl:
Should this be rebased onto 1.19?

EnnuiL:
> Should this be rebased onto 1.19?

I *might* be a bit biased on that, but yeah, considering that Registry Sync is pretty much almost done, with it hopefully getting on the road of a possible FCP, I think we can begin bumping the remaining PRs to 1.19

williambl:
rebased onto 1.19. not yet tested on 1.19.

williambl:
not sure why it auto-requested review from worldgen gui and data teams...

Leo40Git (approved):
LGTM!

williambl:
This has now been tested on 1.19.

library/entity/entity_events/src/testmod/resources/fabric.mod.json
```diff
@@ -0,0 +1,28 @@
+{
```
TheGlitch76: Use the QMJ for testmods

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/EntityReviveEvents.java
```diff
@@ -0,0 +1,98 @@
...
+
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.entity.damage.DamageSource;
+import org.quiltmc.qsl.base.api.event.Event;
+import org.quiltmc.qsl.base.api.event.EventAwareListener;
+
+/**
+ * A pair of callbacks which are invoked when a LivingEntity takes fatal damage.
+ *
+ * <p>Mods can return true from a callback to keep the entity alive, like totems of undying do. This will also prevent
```
EnnuiL:
```suggestion
 * <p>
 * Mods can return true from a callback to keep the entity alive, like totems of undying do. This will also prevent
```
Actually, i'm not gonna do a heckton of reviews like that here; just please use checkstyle and fix all the errors

EnnuiL (changes requested):
Glancing at the PR here, the majority of my complaints can be addressed if you format them to our Checkstyle's standards; Other than that, I don't see anything obviously wrong yet

library/entity/entity_events/src/testmod/resources/fabric.mod.json
williambl: Done

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/EntityReviveEvents.java
williambl: Checkstyle has been run and everything it flagged up fixed. Interestingly it didn't flag up these javadoc issues.

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/EntityWorldChangeEvents.java
```diff
@@ -0,0 +1,95 @@
...
+	@FunctionalInterface
+	public interface AfterPlayerChange extends EventAwareListener {
+		/**
+		 * Called after a player has been moved to different world.
+		 *
+		 * @param player the player
+		 * @param origin the original world the player was in
+		 * @param destination the new world the player was moved to
+		 */
+		void afterChangeWorld(ServerPlayerEntity player, ServerWorld origin, ServerWorld destination);
```
EnnuiL: This should be called `afterWorldChange`, or maybe `afterWorldMove`

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/EntityWorldChangeEvents.java
```diff
@@ -0,0 +1,95 @@
...
+		 * @param original the original entity
+		 * @param copy the new entity at the destination
+		 * @param origin the world the original entity is in
+		 * @param destination the destination world the new entity is in
+		 */
+		void afterChangeWorld(Entity original, Entity copy, ServerWorld origin, ServerWorld destination);
+	}
+
+	@FunctionalInterface
+	public interface AfterPlayerChange extends EventAwareListener {
```
EnnuiL: I'd say "change it to `AfterPlayerWorldChange`", but leaving it open-ended will be good for future expansions

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/EntityWorldChangeEvents.java
```diff
@@ -0,0 +1,95 @@
...
+		 *
+		 * <p>Note this event is not called if the entity is a {@link ServerPlayerEntity}.
+		 * {@link AfterPlayerChange} should be used to track when a player has changed worlds.
+		 *
+		 * @param original the original entity
+		 * @param copy the new entity at the destination
+		 * @param origin the world the original entity is in
+		 * @param destination the destination world the new entity is in
+		 */
+		void afterChangeWorld(Entity original, Entity copy, ServerWorld origin, ServerWorld destination);
```
EnnuiL: Same as AfterPlayerChange's feedback

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/LivingEntityDeathCallback.java
```diff
@@ -0,0 +1,46 @@
...
+import org.quiltmc.qsl.base.api.event.Event;
+import org.quiltmc.qsl.base.api.event.EventAwareListener;
+
+/**
+ * A callback which is invoked on the logical server when a LivingEntity dies.
+ */
+@FunctionalInterface
+public interface LivingEntityDeathCallback extends EventAwareListener {
+	/**
+	 * Invoked when an entity dies serverside.
```
EnnuiL:
```suggestion
	 * Invoked when an entity dies server-side.
```

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/ServerPlayerEntityCopyCallback.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * A callback which is called on the logical server when a player is copied.
+ *
+ * <p>
+ * Players are copied on death and when returning from the end through a portal. The {@code wasDeath} parameter can
+ * be used to differentiate between the two situations.
+ *
+ * <p>
+ * The callback is called after vanilla has done its own copying logic.
+ *
```
EnnuiL:
```suggestion
/**
 * A callback which is called on the logical server when a player is copied.
 * <p>
 * Players are copied on death and when returning from the end through a portal. The {@code wasDeath} parameter can
 * be used to differentiate between the two situations.
 * <p>
 * The callback is called after vanilla has done its own copying logic.
 *
```

EnnuiL (changes requested):
We're almost there! With the Checkstyle clean-ups, I could spot few potential problems

williambl:
> We're almost there! With the Checkstyle clean-ups, I could spot few potential problems

All addressed I think - maybe this'll get merged before its birthday! :P

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/EntityWorldChangeEvents.java
```diff
@@ -0,0 +1,95 @@
...
+		}
+	});
+
+	@FunctionalInterface
+	public interface AfterEntityWorldChange extends EventAwareListener {
+		/**
+		 * Called after an entity has been recreated at the destination when being moved to a different world.
+		 *
+		 * <p>Note this event is not called if the entity is a {@link ServerPlayerEntity}.
+		 * {@link AfterPlayerWorldChange} should be used to track when a player has changed worlds.
```
EnnuiL:
```suggestion
		 * Called after an entity has been recreated at the destination when being moved to a different world.
		 * <p>
		 * Note this event is not called if the entity is a {@link ServerPlayerEntity}.
		 * {@link AfterPlayerWorldChange} should be used to track when a player has changed worlds.
```

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/EntityWorldChangeEvents.java
```diff
@@ -0,0 +1,95 @@
...
+		}
+	});
+
+	/**
+	 * An event which is invoked server-side after a player has been moved to a different world.
+	 *
+	 * <p>This is similar to {@link EntityWorldChangeEvents#AFTER_ENTITY_WORLD_CHANGE} but is only called for players.
+	 * This is because the player is physically moved to the new world instead of being recreated at the destination.
+	 *
+	 * <p>Note that returning from the end via an end portal does not count as a world-change in this way.</p>
```
EnnuiL:
```suggestion
	 * An event which is invoked server-side after a player has been moved to a different world.
	 * <p>
	 * This is similar to {@link EntityWorldChangeEvents#AFTER_ENTITY_WORLD_CHANGE} but is only called for players.
	 * This is because the player is physically moved to the new world instead of being recreated at the destination.
	 * <p>
	 * Note that returning from the end via an end portal does not count as a world-change in this way.</p>
```

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/EntityWorldChangeEvents.java
```diff
@@ -0,0 +1,95 @@
...
+ * @apiNote For a {@link ServerPlayerEntity}, please use {@link EntityWorldChangeEvents#AFTER_PLAYER_WORLD_CHANGE}.
+ */
+public final class EntityWorldChangeEvents {
+	/**
+	 * An event which is invoked server-side after an entity has been moved to a different world.
+	 *
+	 * <p>All entities are copied to the destination and the old entity removed.
+	 * This event does not apply to the {@link ServerPlayerEntity} since players are physically moved to the new world instead of being copied over.
+	 *
+	 * <p>A mod may use this event for reference cleanup if it is tracking an entity's current world.
```
EnnuiL:
```suggestion
	 * An event which is invoked server-side after an entity has been moved to a different world.
	 * <p>
	 * All entities are copied to the destination and the old entity removed.
	 * This event does not apply to the {@link ServerPlayerEntity} since players are physically moved to the new world instead of being copied over.
	 * <p>
	 * A mod may use this event for reference cleanup if it is tracking an entity's current world.
```

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/EntityReviveEvents.java
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * <p>
+ * Mods can return true from a callback to keep the entity alive, like totems of undying do. This will also prevent
+ * {@link LivingEntityDeathCallback} listeners from being called.
+ *
+ * <p>
+ * Vanilla checks for entity health {@code <= 0} each tick (with
+ * {@link net.minecraft.entity.LivingEntity#isDead()}), and kills the entity if this is true - so the entity will
+ * still die next tick if this event is cancelled. It's assumed that the listener will do something to prevent this,
+ * for example:
```
EnnuiL:
```suggestion
 * A pair of callbacks which are invoked when a LivingEntity takes fatal damage.
 * <p>
 * Mods can return true from a callback to keep the entity alive, like totems of undying do. This will also prevent
 * {@link LivingEntityDeathCallback} listeners from being called.
 * <p>
 * Vanilla checks for entity health {@code <= 0} each tick (with
 * {@link net.minecraft.entity.LivingEntity#isDead()}), and kills the entity if this is true - so the entity will
 * still die next tick if this event is cancelled. It's assumed that the listener will do something to prevent this,
 * for example:
```

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/EntityReviveEvents.java
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * <ul>
+ *     <li>a totem-of-undying mod giving the player a little health + regeneration effect</li>
+ *     <li>a minigame mod teleporting the player into a 'respawn room' and setting their health to 20.0</li>
+ *     <li>a mod that changes death mechanics switching the player over to the mod's play-mode, where death doesn't
+ *     apply</li>
+ * </ul>
+ *
+ * <p>
```
EnnuiL: Now I'm unsure about whenever changing this one will be better;
```suggestion
 * <ul>
 *     <li>a totem-of-undying mod giving the player a little health + regeneration effect</li>
 *     <li>a minigame mod teleporting the player into a 'respawn room' and setting their health to 20.0</li>
 *     <li>a mod that changes death mechanics switching the player over to the mod's play-mode, where death doesn't
 *     apply</li>
 * </ul>
 * <p>
```

EnnuiL (changes requested):
I have a final barrage of nitpicks

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/EntityReviveEvents.java
williambl: Sure, it looks cleaner.

EnnuiL (approved):
After months of work, I believe that this is finally ready!

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/EntityReviveEvents.java
```diff
@@ -0,0 +1,96 @@
...
+ *     <li>a minigame mod teleporting the player into a 'respawn room' and setting their health to 20.0</li>
+ *     <li>a mod that changes death mechanics switching the player over to the mod's play-mode, where death doesn't
+ *     apply</li>
+ * </ul>
+ * <p>
+ * These events are short-circuiting - if the entity has already been revived, your callback will not fire.
+ */
+public final class EntityReviveEvents {
+	/**
+	 * Called before totems try to revive the player.
```
EnnuiL:
```suggestion
	 * Called before totems try to revive the entity.
```

EnnuiL (changes requested):
Actually, no, I have a final objection

EnnuiL (approved):

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/ServerPlayerEntityCopyCallback.java
```diff
@@ -0,0 +1,54 @@
...
+
+import net.minecraft.server.network.ServerPlayerEntity;
+
+import org.quiltmc.qsl.base.api.event.Event;
+import org.quiltmc.qsl.base.api.event.EventAwareListener;
+
+/**
+ * A callback which is called on the logical server when a player is copied.
+ * <p>
+ * Players are copied on death and when returning from the end through a portal. The {@code wasDeath} parameter can
```
OroArmor, line 27: Small thing, maybe `wasDead` instead of `wasDeath`

OroArmor (approved):
For such a large number of events, its surprising how few mixins are needed

library/entity/entity_events/src/main/java/org/quiltmc/qsl/entity_events/api/ServerPlayerEntityCopyCallback.java
EnnuiL: Actually, `wasDead` might be a bit confusing because in reality, it's trying to see if the cause was death, not if the player was dead

williambl: > Actually, `wasDead` might be a bit confusing because in reality, it's trying to see if the cause was death, not if the player was dead

agreed.

PR #53, opened 2021-11-27
Add event phases.

LambdAurora:
Also renamed `ArrayEvent` to simply `Event`, I feel like it's an implementation detail that should not be important in the naming over such a generic class.

Most of this is Fabric API port.


library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/QuiltUtil.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package org.quiltmc.qsl.base.api.util;
+
+import java.util.function.Function;
+
+/**
+ * A set of various utilities.
+ */
+public final class QuiltUtil {
```
i509VCB: I'm not sure why this needs to be public api.

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/event/Event.java
```diff
@@ -0,0 +1,329 @@
...
+ * <h2>Example: Executing an event</h2>
+ * <p>
+ * Executing an event is done by calling a method on the event invoker. Where {@code T} is Example, executing an event
+ * is done through the following:
+ *
+ * <pre>{@code
+ * EXAMPLE.invoker().doSomething();
+ * }</pre>
+ *
+ * @param <T> the type of the invoker used to execute an event and the type of the callback
```
i509VCB, line 111: We should mention how to use the phase system here.

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/QuiltUtil.java
LambdAurora: Well, I could put the method in QuiltBaseImpl instead but it would still be public, so I thought why not just generalize it and make it available instead?

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/event/Event.java
LambdAurora: Added.

Leo40Git (approved):

library/core/lifecycle_events/src/main/java/org/quiltmc/qsl/lifecycle/api/client/event/ClientLifecycleEvents.java
```diff
@@ -52,15 +52,15 @@
 	 * was running an integrated server, the integrated server will be shut down. Finally, all client facilities are torn down.
 	 *
 	 * <h2>What should mods do when this event is executed?</h2>
-	 *
+	 * <p>
```
williambl, line 55: Do we have a convention on where to put \<p> tags in javadoc? In most of the existing code it seems to be on the first line of a paragraph, in this PR it seems to always be on its own line.

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/impl/event/EventPhaseData.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Data of an {@link org.quiltmc.qsl.base.api.event.Event} phase.
+ */
+@ApiStatus.Internal
+public final class EventPhaseData<T> {
+	final Identifier id;
+	T[] listeners;
+	final List<EventPhaseData<T>> subsequentPhases = new ArrayList<>();
+	final List<EventPhaseData<T>> previousPhases = new ArrayList<>();
+	int visitStatus = 0; // 0: not visited, 1: visiting, 2: visited
```
williambl: What's the advantage of an int over an enum here? Is it much faster?

library/core/qsl_base/src/testmod/java/org/quiltmc/qsl/base/test/EventTests.java
```diff
@@ -0,0 +1,292 @@
...
+		}
+	}
+
+	private static void testAddedPhases() {
+		var event = createEvent();
+
+		var veryEarly = new Identifier("fabric", "very_early");
+		var early = new Identifier("fabric", "early");
+		var late = new Identifier("fabric", "late");
+		var veryLate = new Identifier("fabric", "very_late");
```
williambl: fabric -> quilt?

library/core/qsl_base/src/testmod/java/org/quiltmc/qsl/base/test/EventTests.java
```diff
@@ -0,0 +1,292 @@
...
+		}
+	}
+
+	private static void testCycle() {
+		var event = createEvent();
+
+		var a = new Identifier("fabric", "a");
+		var b1 = new Identifier("fabric", "b1");
+		var b2 = new Identifier("fabric", "b2");
+		var b3 = new Identifier("fabric", "b3");
```
williambl: More fabric

williambl (approved):
Just some nitpicky things I noticed, but nothing wrong

library/core/lifecycle_events/src/main/java/org/quiltmc/qsl/lifecycle/api/client/event/ClientLifecycleEvents.java
LambdAurora: Putting those in the first line of a paragraph was always unreadable to me, it's an old checkstyle thing on Fabric.

I think they should be on their own line in Quilt.

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/impl/event/EventPhaseData.java
LambdAurora: Switched to an enum, there should not be any changes to how fast it goes, and it's code that runs only on callback registration.
Enum should be more readable.

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/QuiltUtil.java
i509VCB: Okay if this is going to be public, we should probably name it more appropriately, something like `QuiltAssertions`?

LambdAurora: If we don't intend to have more utilities methods that are not assertions in this class, then yes, would be a good idea.

i509VCB (approved):

PR #55, opened 2021-11-27
Game Rule API

IotaBread:
Port of Fabric's Game Rule API


library/content_other/game_rule/src/main/java/org/quiltmc/qsl/game_rule/impl/rule/BoundedIntRule.java
```diff
@@ -0,0 +1,87 @@
...
+import net.minecraft.world.GameRules;
+
+import org.quiltmc.qsl.game_rule.api.GameRuleRegistry;
+import org.quiltmc.qsl.game_rule.mixin.IntRuleAccessor;
+
+public final class BoundedIntRule extends GameRules.IntRule {
+	private static final Logger LOGGER = LogManager.getLogger(GameRuleRegistry.class);
+
+	private final int minimumValue;
+	private final int maximumValue;
```
EnnuiL: Both minimum and maximum values need getters; The reason for that is that it may be useful for a mod that needs to directly access a game rule's bound (example: Boring Default Game Rules and its generation of a JSON Schema based on the game rules)

library/content_other/game_rule/src/main/java/org/quiltmc/qsl/game_rule/api/rule/DoubleRule.java
```diff
@@ -0,0 +1,128 @@
...
+import net.minecraft.server.command.ServerCommandSource;
+import net.minecraft.world.GameRules;
+
+import org.quiltmc.qsl.game_rule.api.GameRuleRegistry;
+
+public final class DoubleRule extends GameRules.Rule<DoubleRule> implements ValidateableRule {
+	private static final Logger LOGGER = LogManager.getLogger(GameRuleRegistry.class);
+
+	private final double minimumValue;
+	private final double maximumValue;
```
EnnuiL: Same here

EnnuiL (changes requested):
Uuh, from my own experiences on making Boring Default Game Rules and Damage Incorporated, Fabric's Game Rule API needs some improvements
The order of the game rule categories are scrambled even if they are all registered by one mod, the name text of enum rules has an inadequate width for wrapping, and compared to Vanilla's game rule types, the Fabric ones are unnecessarily restrictive on the access of its internals (in fact, I ended needing to create Accessors, and all of them were on Fabric API's rules)
I think a revamp of the module similar to the Quilt Tags API one would be more appropriate instead of a direct port, but here's my suggestions of what could be done now:

library/content_other/game_rule/src/main/java/org/quiltmc/qsl/game_rule/api/rule/EnumRule.java
```diff
@@ -0,0 +1,154 @@
...
+import net.minecraft.server.command.ServerCommandSource;
+import net.minecraft.world.GameRules;
+
+import org.quiltmc.qsl.game_rule.api.GameRuleRegistry;
+
+public final class EnumRule<E extends Enum<E>> extends GameRules.Rule<EnumRule<E>> {
+	private static final Logger LOGGER = LogManager.getLogger(GameRuleRegistry.class);
+
+	private final Class<E> classType;
+	private final List<E> supportedValues;
```
EnnuiL, line 44: I think supportedValues also needs a getter too

EnnuiL (changes requested):
oh whoops, forgot this one

EnnuiL (changes requested):
Oh, and finally, DoubleRule needs a `void set(double, MinecraftServer)` method; The vanilla ones have it, why not DoubleRule?

i509VCB:
As the primary author of the game rules api from fabric API, I grant QuiltMC license to the code under the Apache 2.0 license.

For the sake of not causing more license issues, I wouldn't worry about the headers in the impl and mixin packages or the testmod source set.

EnnuiL (changes requested):
The lack of a `void set(double, MinecraftServer)` method for DoubleRule still needs to be addressed, otherwise, it's fine for now

TheGlitch76:
My only concern here is that now that i'm actually staring at it a lot, `content_other` is a really awkward name for a library. I think it should have been `other_content` in the first place.

cc @LambdAurora

library/content_other/game_rule/src/main/java/org/quiltmc/qsl/game_rule/api/CustomGameRuleCategory.java
```diff
@@ -0,0 +1,79 @@
...
+import net.minecraft.text.Text;
+import net.minecraft.util.Identifier;
+import net.minecraft.world.GameRules;
+
+import org.quiltmc.qsl.game_rule.impl.RuleKeyExtensions;
+
+/**
+ * Utility class for creating custom game rule categories outside of the categories {@link GameRules.Category Minecraft provides}.
+ */
+public final class CustomGameRuleCategory {
```
williambl, line 31: Could this be a record?

TheGlitch76:
Hi--this PR seems to be unmaintained right now. I'm converting it to a draft to signify that it's not in a mergable state.

Feel free to mark it as ready for review once you begin working on this again.

EnnuiL:
About a Game Rules API, I believe that a direct port of Fabric's one will not be adequate here. With that (and also the inactivity), I'm closing this PR

PR #56, opened 2021-12-02
Add resource reload events.

LambdAurora:
Those events can be used to apply/re-sync stuff after a resource reload for example.

### DIFF
--- a/library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourceLoaderEvents.java
+++ b/library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourceLoaderEvents.java
@@ -86,7 +86,7 @@ public final class ResourceLoaderEvents {
 		 *
 		 * @param server          the server, may be {@code null} for the first reload
 		 * @param resourceManager the resource manager, may be {@code null} if the data pack reload failed
-		 * @param error           present if the data pack reload failed
+		 * @param error           present if the data pack reload failed, otherwise {@code null}
 		 */
 		void onEndDataPackReload(@Nullable MinecraftServer server, ServerResourceManager resourceManager, @Nullable Throwable error);
 	}
--- a/library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourceLoaderEvents.java
+++ b/library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourceLoaderEvents.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2021 QuiltMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.quiltmc.qsl.resource.loader.api;
+
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.resource.ServerResourceManager;
+import net.minecraft.server.MinecraftServer;
+
+import org.quiltmc.qsl.base.api.event.Event;
+import org.quiltmc.qsl.resource.loader.api.reloader.IdentifiableResourceReloader;
+
+/**
+ * Events related to the resource loader.
+ */
+public final class ResourceLoaderEvents {
+	private ResourceLoaderEvents() {
+		throw new UnsupportedOperationException("ResourceLoaderEvents only contains static definitions.");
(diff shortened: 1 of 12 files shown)


LemmaEOF (approved):

Leo40Git (approved):

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/client/ClientResourceLoaderEvents.java
```diff
@@ -0,0 +1,98 @@
...
+		/**
+		 * Called after resource packs on the Minecraft client have been reloaded.
+		 * <p>
+		 * If the reload was not successful, the old resource packs will be kept.
+		 *
+		 * @param client          the client
+		 * @param resourceManager the resource manager
+		 * @param error           present if the resource pack reload failed
+		 */
+		void onEndResourcePackReload(MinecraftClient client, ResourceManager resourceManager, Optional<Throwable> error);
```
BasiqueEvangelist: Why does this take in an `Optional<Throwable>`, while the server equivalent takes in a nullable `Throwable`?

LambdAurora: This is only mirroring Minecraft's internal code. 
On the client it returns an optional at the end of the resource reloading, and on the server it only returns the bare `Throwable`.
If we absolutely want consistency, either I wrap the error into an Optional on server, or I pass a Nullable error everywhere.

BasiqueEvangelist: Couldn't you just make it return a nullable error everywhere? That would help with consistency, and wouldn't incur any extra cost on the server (as opposed to wrapping it on the server).

BasiqueEvangelist:
Is there any way to distinguish between first load and subsequent reloads on client?

LambdAurora:
> Is there any way to distinguish between first load and subsequent reloads on client?

no, but in impl, yes

BasiqueEvangelist:
> > Is there any way to distinguish between first load and subsequent reloads on client?
> 
> no, but in impl, yes

Could you add something like this, then? This could help for mods which specifically want to run something on reloads.

library/core/resource_loader/src/main/resources/quilt_resource_loader.mixins.json
```diff
@@ -16,7 +16,11 @@
     "client.CreateWorldScreenMixin",
     "client.FontManagerResourceReloadListenerMixin",
     "client.GameOptionsMixin",
-    "client.KeyedClientResourceReloaderMixin"
+    "client.KeyedClientResourceReloaderMixin",
+    "client.MinecraftClientMixin"
+  ],
+  "server": [
```
i509VCB: No reason to mark the mixin server only as the target always exists on client

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourceLoaderEvents.java
```diff
@@ -0,0 +1,93 @@
...
+		/**
+		 * Called after data packs on a Minecraft server have been reloaded.
+		 * <p>
+		 * If the reload was not successful, the old data packs will be kept.
+		 *
+		 * @param server          the server, may be {@code null} for the first reload
+		 * @param resourceManager the resource manager, may be {@code null} if the data pack reload failed
+		 * @param error           present if the data pack reload failed, otherwise {@code null}
+		 */
+		void onEndDataPackReload(@Nullable MinecraftServer server, ServerResourceManager resourceManager, @Nullable Throwable error);
```
i509VCB, line 91: One of the questions I was asked when I designed an event like this in fabric api is, is the throwable important to users of the event?

library/core/resource_loader/src/main/resources/quilt_resource_loader.mixins.json
LambdAurora: Uh? The dedicated server main class exists on the client???

LambdAurora:
> > > Is there any way to distinguish between first load and subsequent reloads on client?
> 
> Could you add something like this, then? This could help for mods which specifically want to run something on reloads.

Added

library/core/resource_loader/src/main/resources/quilt_resource_loader.mixins.json
i509VCB: Yep

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourceLoaderEvents.java
LambdAurora: imo yes, for example on client it could redirect to a temporary screen describing the error, what is to blame, etc.

And I prefer to be able to know that something fails if the end event trigger rather than being blind on that data.

BasiqueEvangelist:
~~smh merged just while i was reviewing it~~

LambdAurora:
> ~smh merged just while i was reviewing it~

eh, if there's still an issue we still can fix it, it's not set in stone

PR #59, opened 2021-12-02
Key Binds API

EnnuiL:
A built-from-scratch API that covers Fabric Key Binding API's functionality and more

The PR was planned in two stages: Blue and Pink. The implemented feature set of Blue was:
- Key binds now support being dynamically disabled, making them completely invisible to the game's eyes
- The Key Binds screen now shows a tooltip on conflicting keys showing the other key binds that use the same key, and adds a `(!)` indicator to the key text itself for accessibility reasons
- The API exposes some methods meant to help with mod intercompatibility

Currently, work is being done on the Pink feature set:
- [ ] A custom config file at `./config/qsl/key_binds.json` allowing for our own key bind extensions (WIP)
- [ ] Key binds now accepts key chords, expanding the possibilities for modded key binds (WIP)
- [ ] Key binds will contain key predicates, which essentially points all the use cases where it will be used. This will allow multiple key binds that never overlaps to share the same key

Outdated but relevant information about the stages is available in #61


TheGlitch76:
Now is the time for any last-minute bikesheds on where to put the keybindings API, since I know people are unhappy with `gui`

OroArmor:
`gui` isn't perfect, but its the closest large group in my opinion.

you could make an `input` or `client` group, but then we end up with 30 different modules

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/api/KeyBindingRegistry.java
```diff
@@ -0,0 +1,78 @@
...
+
+import net.fabricmc.api.EnvType;
+import net.fabricmc.api.Environment;
+
+import net.minecraft.client.option.KeyBinding;
+
+import org.quiltmc.qsl.key.bindings.impl.KeyBindingRegistryImpl;
+
+@Environment(EnvType.CLIENT)
+public class KeyBindingRegistry {
```
OroArmor: Javadocs for this class.

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/api/KeyBindingRegistry.java
```diff
@@ -0,0 +1,78 @@
...
+
+	/**
+	 * Sets the state of the key binding.
+	 *
+	 * @param key the key binding
+	 * @param enabled the new state
+	 */
+	public static void setEnabled(KeyBinding key, boolean enabled) {
+		KeyBindingRegistryImpl.setEnabled(key, enabled);
+	}
```
OroArmor: What exactly does `enabled` mean? Personally I know, but its a little vague here and should be clarified to be that the `KeyBinding` does not work.

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/impl/KeyBindingRegistryImpl.java
```diff
@@ -0,0 +1,133 @@
...
+
+	public static KeyBinding registerKeyBinding(KeyBinding key, boolean enabled) {
+		if (key == null) {
+			LOGGER.error("Attempted to register a null key bind!");
+			return null;
+		}
+
+		for (KeyBinding otherKey : quiltKeys.keySet()) {
+			if (key == otherKey || key.getTranslationKey() == otherKey.getTranslationKey()) {
+				LOGGER.error("Attempted to register %s, but a conflicting key has already been registered!", key.getTranslationKey());
```
OroArmor: Clarify what `conflicting` means in this context. I would assume the same keybind, but its the translation key.

OroArmor: In addition, i would have a separate error for when the same keybinding is registered twice.

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/mixin/client/KeyBindingEntryMixin.java
```diff
@@ -0,0 +1,100 @@
...
+import net.minecraft.client.util.math.MatrixStack;
+import net.minecraft.text.Text;
+import net.minecraft.text.TranslatableText;
+import net.minecraft.util.Formatting;
+
+import org.quiltmc.qsl.key.bindings.impl.ConflictTooltipOwner;
+
+@Environment(EnvType.CLIENT)
+@Mixin(KeyBindingEntry.class)
+public abstract class KeyBindingEntryMixin extends KeyBindingListWidget.Entry implements ConflictTooltipOwner {
```
OroArmor: I would add a way to show that a keybinding is disabled. I'm sure there is a texture that would work for this

OroArmor (approved):
Looks great! I know you talked about supporting some combination keybinds, not sure if that should be an extra PR or not, but its definitely something to discuss further

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/mixin/client/KeyBindingEntryMixin.java
EnnuiL: When a keybind's disabled, not only it won't work, but it'll be hidden in the game; It's that way because it would be perfect for the usecase of disabling a keybind in-game without requiring a restart (ex: Ok Zoomer's extra keybinds).

However, thinking about it, perhaps I should split the `enabled` state into `enabled` and `visible`; I'll see what I can do, since there's still room for major changes like that

OroArmor: Ah ok. I thought it would still show up, but have no effect in game. makes sense now.

EnnuiL: After some thought, having a separate state for only disabling it functionally doesn't really make sense; Mods can already do that themselves

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/api/KeyBindingRegistry.java
```diff
@@ -0,0 +1,117 @@
...
+	 */
+	public static boolean isEnabled(KeyBinding key) {
+		return KeyBindingRegistryImpl.isEnabled(key);
+	}
+
+	/**
+	 * Sets the state of the key binding.
+	 *
+	 * <p>This method allows for disabling the key binding, making it invisible
+	 * to the game. options.txt is the only expection, who still stores the disabled
```
BasiqueEvangelist:
```suggestion
	 * to the game. options.txt is the only exception, who still stores the disabled
```

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/api/KeyBindingRegistry.java
```diff
@@ -0,0 +1,117 @@
...
+		return ((KeyBindingAccessor) key).getBoundKey();
+	}
+
+	/**
+	 * Gets the state of the key binding.
+	 *
+	 * <p>The state of a key binding determines whenever it's disabled or not.
+	 *
+	 * <p>When a keybind is disabled, it is effectively hidden from the game,
+	 * being non-existent to it. options.txt is the only expection, who still
```
BasiqueEvangelist:
```suggestion
	 * being non-existent to it. options.txt is the only exception, who still
```

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/impl/KeyBindingRegistryImpl.java
```diff
@@ -0,0 +1,134 @@
...
+	public static boolean isEnabled(KeyBinding key) {
+		if (quiltKeys.containsKey(key)) {
+			return quiltKeys.get(key);
+		}
+
+		return false;
+	}
+
+	public static void setEnabled(KeyBinding key, boolean enabled) {
+		if (quiltKeys.containsKey(key)) {
```
BasiqueEvangelist: This should throw an exception if the key isn't a quilt key

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/impl/KeyBindingRegistryImpl.java
```diff
@@ -0,0 +1,134 @@
...
+				KeyBindingAccessor.getKeysById().remove(key.getTranslationKey(), key);
+			}
+
+			((KeyBindingAccessor) key).callReset();
+			KeyBinding.updateKeysByCode();
+		}
+	}
+
+	public static void registerKeyBindingManager(GameOptions options, KeyBinding[] allKeys) {
+		keyBindingManagers.add(new KeyBindingManager(options, allKeys));
```
BasiqueEvangelist: Is there any case where there is *more than one* key binding manager?

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/impl/KeyBindingRegistryImpl.java
```diff
@@ -0,0 +1,134 @@
...
+	public static final Logger LOGGER = LogManager.getLogger();
+
+	private static Map<KeyBinding, Boolean> quiltKeys = new HashMap<>();
+	private static List<KeyBindingManager> keyBindingManagers = new ArrayList<>(1);
+	private static KeyBinding[] enabledQuiltKeysArray = new KeyBinding[] {};
+	private static List<KeyBinding> disabledQuiltKeys = new ArrayList<>(0);
+
+	public static KeyBinding registerKeyBinding(KeyBinding key, boolean enabled) {
+		if (key == null) {
+			LOGGER.error("Attempted to register a null key bind!");
```
BasiqueEvangelist: Why is this just logged and not an immediate exception?

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/impl/KeyBindingRegistryImpl.java
EnnuiL: Good idea

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/impl/KeyBindingRegistryImpl.java
EnnuiL: Well, if another mod creates another GameOptions instance, then yes, there would be more than one manager, however, I'll probably ditch this system; It won't really be future-proof for a possible creation of a separate keybind options file

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/impl/KeyBindingRegistryImpl.java
EnnuiL: Like, throw an NPE? I already return it immediately after that message

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/api/KeyBindingRegistry.java
EnnuiL: I am _sooooo_ good at speaking languages

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/impl/KeyBindingRegistryImpl.java
BasiqueEvangelist: You just return `null`, not an NPE.
This would mean that an NPE is thrown the first time the mod tries to use the keybind, which is... not optimal.

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/impl/KeyBindingRegistryImpl.java
BasiqueEvangelist: > if another mod creates another GameOptions instance

Do mods ever create GameOptions instances, though? And if they do, why?

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/impl/KeyBindingRegistryImpl.java
EnnuiL: I already return `null` one line after the log message though, I don't let the rest of the code run and make bad things happen

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/impl/KeyBindingRegistryImpl.java
EnnuiL: I'm not aware of any mods that do that, but well, mods do cursed stuff all the time, so I thought that having more than one GameOptions instance would be plausible

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/impl/KeyBindingRegistryImpl.java
EnnuiL: And done; It's currently the same exception for both non-registered keys and vanilla ones, but I'll eventually split it

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/impl/KeyBindingRegistryImpl.java
EnnuiL: Oh boy, that was a confusing discussion, but the idea (use exceptions instead of returning null) has been implemented

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/api/KeyBindingRegistry.java
```diff
@@ -0,0 +1,135 @@
...
+ * <p>This class also allows for getting key binds registered by other mods.
+ */
+@Environment(EnvType.CLIENT)
+public class KeyBindingRegistry {
+	/**
+	 * Registers a key binding with it being initially enabled.
+	 *
+	 * <pre>
+	 * {@code
+	 * public static final KeyBinding EXAMPLE_KEY_BIND = KeyBinding.registerKeyBinding(
```
OroArmor:
```suggestion
	 * public static final KeyBinding EXAMPLE_KEY_BIND = KeyBindingRegistry.registerKeyBinding(
```

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/impl/ConflictTooltipOwner.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.util.List;
+
+import net.fabricmc.api.EnvType;
+import net.fabricmc.api.Environment;
+
+import net.minecraft.text.Text;
+
+@Environment(EnvType.CLIENT)
+public interface ConflictTooltipOwner {
```
OroArmor: This could either be exposed as API as it could have some functionality or make it slightly clearer that it's internal with `@ApiStatus.Internal`

OroArmor (changes requested):

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/api/KeyBindingRegistry.java
EnnuiL: Oh whoops, I'll fix that

library/gui/key_bindings/src/main/java/org/quiltmc/qsl/key/bindings/impl/ConflictTooltipOwner.java
EnnuiL: Ah, I was thinking of replacing the interface mixin with the two accessors that I've done for resource pack tooltips; I think it's much cleaner than the current one

EnnuiL: ...and nevermind, while the accessor does simply a bit of code, ConflictTooltipOwner will still be needed;
I'll mark it as internal

library/gui/key_binds/src/main/java/org/quiltmc/qsl/key/binds/api/KeyBindRegistry.java
```diff
@@ -0,0 +1,135 @@
...
+		KeyBindRegistryImpl.setEnabled(key, enabled);
+	}
+
+	/**
+	 * Returns a map containing all modded key binds (and vanilla ones if specified).
+	 *
+	 * @param includeVanilla {@code true} if vanilla entries should be included, else {@code false}
+	 * @return a map containing all modded (and optionally vanilla) key binds
+	 */
+	public static Map<KeyBind, Boolean> getAllKeyBinds(boolean includeVanilla) {
```
LambdAurora: This really should use fastutil's Object2BooleanMap.

EnnuiL: I'll probably abandon the Map<KeyBind, Boolean> approach once on the Pink stage, because having some sort of `QuiltKeyBinds` would probably be better than a map between keybinds and the extra info

i509VCB: Probably would be a good idea to encapsulate the inner map so the api can be a little more resistant to change.

LambdAurora:
> A custom config file at `./config/qsl/key_binds.json` allowing for our own key bind extensions (WIP)

Now that I'm thinking of it, probably should be `./config/quilt/key_binds.json`.

EnnuiL:
> > A custom config file at `./config/qsl/key_binds.json` allowing for our own key bind extensions (WIP)
> 
> Now that I'm thinking of it, probably should be `./config/quilt/key_binds.json`.

I thought about it (after all, FAPI used `./config/fabric/`), however, I don't know; I'd think `./config/quilt/` would instead be used by something like the Quilt Loader, and not QSL

PR #62, opened 2021-12-19
Feature:  event as entrypoints / Event.listenAll

LambdAurora:
This pull request adds the following:

 - `Event.listenAll(Object listener, Event<?> ... events)`
   This method allows to bulk-register a listener instance to multiple events, throws if the events array is empty or if the listener doesn't implement one of the event's callback interface.
 - Events as entrypoints
   - Introduce 3 entrypoints: `events`, `client_events`, `server_events`
   - Any of those entrypoints can implement event callback interfaces and if those interfaces extend `EventAwareListener`, `ClientEventAwareListener`, `DedicatedServerEventAwareListener`, then the entrypoint object will automatically be registered as a listener of the associated events. Will not work for all events since some events can have multiple event instances associated to the same callback interface.
   
## But why?

There's already multiple ways of registering a listener, and this only makes it easier:
 - why deal with a bunch of lambdas and listener registration in a mod initializer if we just need to deal with some events?
 - grouping multiple events as a bundle for a feature, then registering the listener instance in one call instead of a lot of calls


library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/impl/event/EventRegistry.java
```diff
@@ -0,0 +1,113 @@
...
+			// Search if the callback qualifies is unique to this event.
+			if (target.listenerClass().isAssignableFrom(type)) {
+				List<?> entrypoints = FabricLoader.getInstance().getEntrypoints(target.entrypointKey(), target.listenerClass());
+
+				// Search for matching entrypoint.
+				for (Object entrypoint : entrypoints) {
+					var listenedPhases = getListenedPhases(entrypoint.getClass());
+
+					// Searching if the given entrypoint is a listener of the event being registered.
+					if (type.isAssignableFrom(entrypoint.getClass())) {
```
i509VCB: Is this enough is one question I would ask. If an event callback interface has a generic this may be wrong?

i509VCB: More specifically we need tests to ensure we don't let undefined behavior in here.

LambdAurora: If there is a generic, it implies there can be multiple instances of an event for the given callback, so for this register method this is right, since it won't extend EventAwareListener.

Tho for `listenAll` it mays be wrong then, since it uses the event array, I'm not exactly sure how I can compare it?

FoundationGames:
This looks really good apart from one thing -
The registry entry added event takes an additional argument upon registration of the callback, to specify which registry to listen to events for. How would such a thing be implemented here?

Also, to address the concern of multiple events using the same callback interface - it might be good to require all events to use unique callback interfaces with unique method names

LambdAurora:
@FoundationGames I honestly don't understand what you mean.

TheGlitch76 (approved):
as i5 said I'd like to see some tests here, but I don't consider it important enough to be a blocker for this PR

LambdAurora:
> as i5 said I'd like to see some tests here, but I don't consider it important enough to be a blocker for this PR

listenAll having generics sanity checks would be nice, but I have to admit it's not the most motivating thing to do

LambdAurora:
Ok, after looking into the generics issue, the simplest and most efficient solution is to disallow registration of listeners for an event whose callback type have generics in `Event#listenAll`, otherwise it would involve very expensive reflection calls which are just not worth it.

This should address the concerns. It also now comes with some tests.

Leo40Git (approved):

TheGlitch76:
Is this PR OK for a last call?

LambdAurora:
imo, yes

williambl (approved):
LGTM

PR #65, opened 2022-01-12
Block Extensions API: add map color provider setter

Leo40Git:


LambdAurora (approved):
Nice.

Now we need to figure out why actions are failing.

TheGlitch76 (approved):

PR #72, opened 2022-01-28
Make Gradle look way better

OroArmor:
Dont touch this code, it works after bashing gradle in the head for a couple hours.


OroArmor:
This does have some side effects, but they don't seem to impact the compilation, mostly the `pom.xml` and `.module` files, but it should be fine :tm:

TheGlitch76 (changes requested):
I'd really like to see a comprehensive example of this system somewhere accessible, so that newbies can find all the things they can do with dependencies without trying to read undocumented Gradle code or sifting through a ton of module buildscripts.

i509VCB:
I know we aren't using kotlin dsl, but is there a way to express `module(testmod = true)` or something like it? The braces are fine if there is no real other option.

TheGlitch76:
It's easy if we're writing our gradle plugin in Groovy like you're intended to, but from my limited research its impossible in Gradle unless we do something like [this](https://blog.mrhaki.com/2013/01/groovy-goodness-adding-extra-methods.html)

OroArmor:
> I know we aren't using kotlin dsl, but is there a way to express `module(testmod = true)` or something like it? The braces are fine if there is no real other option. 

I do not know. I assume not, as the gradle systems I am using do not allow it. I do plan on moving to an enum for the different configurations though.

OroArmor:
> I'd really like to see a comprehensive example of this system somewhere accessible, so that newbies can find all the things they can do with dependencies without trying to read undocumented Gradle code or sifting through a ton of module buildscripts.

Documented in `CONTRIBUTING.md`

CONTRIBUTING.md
```diff
@@ -177,6 +177,32 @@ In the case of a pseudo-local variable (a field used briefly to pass around a lo
...
+    moduleDependencies {
+        library_name { // The QSL Library to depend on
+            module_name { // The QSL Module in `library_name` to depend on
+                api() // Default behavior, should not need to be called.
+            }
+            module_2 {
+                testmod() // Only depend on this module when testing the code
+            }
+            module_3 {
+                impl() // This library is only used to compile this module and is not part of the API. This should be rarely used.
```
TheGlitch76: i don't think that statement is true, it would be better to explain when to use it (which is probably pretty often)
```suggestion
                // By default, dependencies are `api` dependencies; they are put on the classpath
                // of mods that depend on this module.
                // Impl dependencies are dependencies that are only used internally in a module,
                // and classes from it are never exposed through this module's public API.
                // For example, a dependency on lifecyle events would usually be impl
                impl() 
```

CONTRIBUTING.md
```diff
@@ -177,6 +177,32 @@ In the case of a pseudo-local variable (a field used briefly to pass around a lo
...
+In the `qslModule` extension, there is a `moduleDependencies` field. Dependencies are declared in a tree like structure reflecting how QSL libraries and modules are layed out.
+This field can be configured like:
+```groovy
+qslModule {
+    // ...
+    moduleDependencies {
+        library_name { // The QSL Library to depend on
+            module_name { // The QSL Module in `library_name` to depend on
+                api() // Default behavior, should not need to be called.
+            }
```
TheGlitch76:
```suggestion
            // Simple dependencies do not need a code block.
            module_name // the QSL Module in `library_name` to depend on
```

build-logic/src/main/java/qsl/internal/dependency/QslModuleDependency.java
```diff
@@ -0,0 +1,58 @@
...
+
+		configuration = objects.property(ConfigurationType.class);
+		configuration.convention(ConfigurationType.API);
+	}
+
+	public Property<ConfigurationType> getConfiguration() {
+		return configuration;
+	}
+
+	public void testmod() {
```
TheGlitch76: this should be testmodOnly() so that it's more obvious this dependency isn't included in the main pom

TheGlitch76 (changes requested):
More reviews to clarify documentation and names.

By the way, why does this PR not use something like:
```groovy
library_name {
    api("module_1")
    impl("module_2")
    testmodOnly("module_3")
}
```
IMO the above is clearer than the current version of this PR, because it doesn't leave `api` configuration to be magic and reduces the verbosity of `impl` and `testmodOnly` dependencies

Leo40Git (approved):

build-logic/src/main/java/qsl/internal/dependency/QslLibraryDependency.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.List;
+import java.util.Map;
+
+import org.gradle.api.Named;
+import org.gradle.api.Project;
+import org.gradle.api.artifacts.Dependency;
+import org.gradle.api.plugins.JavaPlugin;
+import org.jetbrains.annotations.NotNull;
+
+public class QslLibraryDependency  implements Named {
```
NoComment1105: Is github being wierd or is there a double space after the class name?

library/data/tags/build.gradle
```diff
@@ -6,16 +6,16 @@ qslModule {
...
-			"qsl_base",
-			"lifecycle_events"
-	])
-	interLibraryTestmodDependencies("command", [
-			"command"
-	])
+	moduleDependencies {
+		core {
+			api("resource_loader")
+			testmodOnly("qsl_base")
```
LambdAurora, line 12: This feels strange, might need to test again if it launches correctly without this specified.

build-logic/src/main/java/qsl/internal/dependency/QslLibraryDependency.java
```diff
@@ -0,0 +1,67 @@
...
+	@Override
+	public @NotNull String getName() {
+		return name;
+	}
+
+	/**
+	 * The configuration type for the module dependency
+	 */
+	public enum ConfigurationType {
+		API(JavaPlugin.API_CONFIGURATION_NAME), IMPLEMENTATION(JavaPlugin.IMPLEMENTATION_CONFIGURATION_NAME), TESTMOD("testmodImplementation");
```
LambdAurora: Would make this multiline

CONTRIBUTING.md
```diff
@@ -177,6 +177,40 @@ In the case of a pseudo-local variable (a field used briefly to pass around a lo
...
 and finally the name of the local (`quilt$injectedMethod$localName`).
 
+## Gradle Conventions
+
+### Declaring dependencies between modules
+
+In the `qslModule` extension, there is a `moduleDependencies` field. Dependencies are declared in a tree like structure reflecting how QSL libraries and modules are layed out.
+This field can be configured like:
+```groovy
+qslModule {
```
LambdAurora, line 187: Definitely want to modify this code snippet since it's different now.

LambdAurora (commented):
The new version following glitch's suggestion is much better.

Apart from the comments, all good.

NoComment1105 (commented):
Looks great, just one comment

build-logic/src/main/java/qsl/internal/dependency/QslLibraryDependency.java
OroArmor: there is.

library/data/tags/build.gradle
OroArmor: Not sure what you mean by "strange". Looking at the diff it should be fine. 

And testing breaks the tags API... :ohno:

OroArmor: This behavior exists before this pr.

Juuxel:
> I know we aren't using kotlin dsl, but is there a way to express `module(testmod = true)` or something like it? The braces are fine if there is no real other option.

@i509VCB Gradle plugins for groovy dsl tend to use maps for that: `module(testmod: true)`

CONTRIBUTING.md
```diff
@@ -177,6 +177,36 @@ In the case of a pseudo-local variable (a field used briefly to pass around a lo
...
+
+            // Only depend on this module when testing the code
+            testmodOnly("module_2")
+
+            // By default, dependencies are `api` dependencies; they are put on the classpath
+            // of mods that depend on this module.
+            // Impl dependencies are dependencies that are only used internally in a module,
+            // and classes from it are never exposed through this module's public API.
+            // For example, a dependency on lifecyle events would usually be impl
+            impl("module_3")
```
TheGlitch76:
```suggestion
            // API dependencies are put on the classpath of mods that depend on this module.
            // Use an API dependency when you expose a class from this dependency in this module's
            // public API--for example, a method that returns a `FooBar`, defined in `module_name`
            // When in doubt, use an API dependency.
            api("module_name")

            // Impl dependencies are dependencies that are only used internally in a module,
            // and classes from it are never exposed through this module's public API.
            // For example, a dependency on lifecyle events would usually be impl
            impl("module_2")
            
            // This module is only depended on when testing the code
            testmodOnly("module_2")
```

OroArmor:
```suggestion
            // API dependencies are put on the classpath of mods that depend on this module.
            // Use an API dependency when you expose a class from this dependency in this module's
            // public API--for example, a method that returns a `FooBar`, defined in `module_name`
            // When in doubt, use an API dependency.
            api("module_name")
            
            // Impl dependencies are dependencies that are only used internally in a module,
            // and classes from it are never exposed through this module's public API.
            // For example, a dependency on lifecyle events would usually be impl
            impl("module_2")
            
            // This module is only depended on when testing the code
            testmodOnly("module_3")
```

CONTRIBUTING.md
```diff
@@ -177,6 +177,37 @@ In the case of a pseudo-local variable (a field used briefly to pass around a lo
...
+This field can be configured like:
+```groovy
+qslModule {
+    // ...
+    moduleDependencies {
+        // The QSL Library to depend on
+        library_name {
+            // API dependencies are put on the classpath of mods that depend on this module.
+            // Use an API dependency when you expose a class from this dependency in this module's
+            // public API--for example, a method that returns a `FooBar`, defined in `module_name`
```
TheGlitch76:
```suggestion
            // public API--for example, a method that returns a `FooBar`, defined in `module_name`.
```

CONTRIBUTING.md
```diff
@@ -177,6 +177,37 @@ In the case of a pseudo-local variable (a field used briefly to pass around a lo
...
+            // When in doubt, use an API dependency.
+            api("module_name")
+            
+            // Impl dependencies are dependencies that are only used internally in a module,
+            // and classes from it are never exposed through this module's public API.
+            // For example, a dependency on lifecyle events would usually be impl
```
TheGlitch76:
```suggestion
            // For example, a dependency on lifecyle events would usually be impl.
```

TheGlitch76 (approved):

library/data/tags/build.gradle
LambdAurora: I know, I put it there because for some reason qsl_base wasn't exposed to the tags API through transitivity using resource loader, this is why I say it's strange.
If that behavior remains, that's unfortunate.

LambdAurora (approved):

TheGlitch76:
This PR has entered its final comment period, to end on Saturday as this is not a change affecting the public API.

PR #77, opened 2022-02-19
Recipe API

LambdAurora:
*Being able to register recipes directly through code is amazing. Especially if you need it for recipes you cannot determine at compile-time.*

So here it is! Recipe API! *It's back*
It's basically a very simple API to register recipes through code.
It also allows the modification and the removal of recipes through other events.

## Why should I use this?

This does not replace good old JSON data-generation, but this API targets other use-cases:
 - addition of runtime recipes, those are recipes a mod cannot determine in advance, like in the case of Aurora's Decorations where there's a lot of wooden blocks that are automatically determined at runtime;
 - modification of recipes;
 - removal of recipes

This, overall, attempts to greatly improve mod compatibility, be more optimized than the virtual resource pack way.
The other target of this API is to allow mass-modification of recipes, so mods like KubeJS can hook into.

## What does it replace?

Data Generation? No, not at all, static recipe-generation should not use this API.
KubeJS? No, this is not targeted to users, though, it allows the implementation of KubeJS-style of mods.
ARRP? For runtime recipe stuff only. Constructing a `Recipe<?>` is much simpler than serializing to JSON.

## But won't it break datapacks?

tl;dr no

The way this works is it will only truly register a recipe if nothing provides it already.
And modification/removal happens after datapacks are loaded.

## What could be future plans for this module?

This module could hold tools to do Ingredient analysis and maybe modification.

## WAIT WAIT, WHAT ABOUT VIRTUAL RESOURCE PACKS?

No, constructing directly the Recipe object is just plain better:
 - Faster, there is no recipe construction -> serialization -> storage of the JSON in RAM -> deserialization -> Recipe object steps
 - Less RAM hungry, no need for a JSON representation to stay forever in RAM, can be argued that it's negligible
 - JSON is not the single source of truth, it never was. If you're convinced it is, go tell to all the worldgen modders that they are now forbidden to do in-code registration and now that they should only use JSON

### Ok, but what about KubeJS then?

This is a standard library for Quilt, this library contains the standard elements to build mods in the Quilt ecosystem.
While it should try to not break mods, and improve compatibility. It also should not be dictated by one mod.

KubeJS won't have a global picture of recipes anymore unless it uses the API too.

## Wait wait, what about the JSON of the recipes?

There's an extension to the recipe serializer to convert a recipe to JSON. There's also a property called `quilt.recipe.dump`, if set to true, at each reload it will dump recipe JSONs in `.minecraft/debug/quilt/recipe`

## Testmod?

The testmod tests the static recipe registering and the dynamic registering, the static one is just a Stick -> Diamond shapeless recipe, the dynamic one is [Iron,Gold,Coal,Charcoal] -> randomized item at reload 4x4 shaped recipe.

## Final note

This API is not final, it always can evolve to better meet the needs and constraints of the community, please keep this in mind when reviewing this PR.


LemmaEOF (approved):
Great to see this coming back! The modify and remove hooks should be a big help for recipe-scripting mods. I might give this a more in-depth look for a new mod or two…

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/impl/RecipeManagerImpl.java
```diff
@@ -0,0 +1,443 @@
...
+			this.recipeManager = recipeManager;
+			this.recipes = recipes;
+		}
+
+		private void add(Recipe<?> recipe) {
+			Map<Identifier, Recipe<?>> type = this.recipes.get(recipe.getType());
+
+			if (type == null) {
+				throw new IllegalStateException("The given recipe " + recipe.getId()
+						+ " do not have its recipe type " + recipe.getType() + " in the recipe manager.");
```
Leo40Git:
```suggestion
						+ " does not have its recipe type " + recipe.getType() + " in the recipe manager.");
```

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/RecipeLoadingEvents.java
```diff
@@ -0,0 +1,285 @@
...
+			 * @return the recipe if present and of the correct type, else {@code null}
+			 */
+			@Nullable <T extends Recipe<?>> T getRecipe(Identifier id, RecipeType<T> type);
+
+			/**
+			 * Returns all registered recipes.
+			 *
+			 * @return the registered recipes
+			 */
+			Map<RecipeType<?>, Map<Identifier, Recipe<?>>> getRecipes();
```
Leo40Git, line 188: Turn this into a `Table<RecipeType<?>, Identifier, Recipe<?>>`, maybe?

Leo40Git (approved):
Seems good! Just some minor nitpicks.

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/RecipeLoadingEvents.java
LambdAurora: Not sure this is a good idea, this is exactly the same representation as Mojang is using in `RecipeManager`.

Leo40Git: why doesn't Mojang use tables then >:V

library/data/recipe/build.gradle
```diff
@@ -0,0 +1,21 @@
...
+
+qslModule {
+	moduleName = "recipe"
+	version = "1.0.0"
+	library = "data"
+	moduleDependencies {
+		core {
+			api("qsl_base")
+			api("resource_loader")
+			testmodOnly("qsl_base")
```
OroArmor: shouldnt be needed. The `api` dependency is in `testmodOnly`

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/VanillaRecipeBuilders.java
```diff
@@ -0,0 +1,407 @@
...
+	 * @param pattern the pattern of the shaped crafting recipe
+	 * @param keys    the keys and ingredients of the recipe
+	 * @param width   the width of the shaped crafting recipe
+	 * @param height  the height of the shaped crafting recipe
+	 * @return the ingredients
+	 * @throws IllegalStateException if a key has no assigned ingredient or if there is an ingredient but no assigned key
+	 */
+	public static DefaultedList<Ingredient> getIngredients(String[] pattern, Char2ObjectMap<Ingredient> keys, int width, int height) {
+		DefaultedList<Ingredient> ingredients = DefaultedList.ofSize(width * height, Ingredient.EMPTY);
+		var patternSet = new CharArraySet(keys.keySet());
```
OroArmor: maybe
```suggestion
		var unusedKeys = new CharArraySet(keys.keySet());
```

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/VanillaRecipeBuilders.java
```diff
@@ -0,0 +1,407 @@
...
+				patternSet.remove(key);
+				ingredients.set(j + width * i, ingredient);
+			}
+		}
+
+		if (!patternSet.isEmpty()) {
+			throw new IllegalStateException("Key defines symbols that aren't used in pattern: " + patternSet);
+		} else {
+			return ingredients;
+		}
```
OroArmor:
```suggestion
		}
		return ingredients;
```

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/VanillaRecipeBuilders.java
```diff
@@ -0,0 +1,407 @@
...
+		return new CampfireCookingRecipe(id, group, input, output, experience, cookTime);
+	}
+
+	/**
+	 * Represents the basis of a recipe builder.
+	 *
+	 * @param <Self>   the type of the recipe builder
+	 * @param <Result> the type of the recipe
+	 */
+	public static abstract class Builder<Self extends Builder<Self, Result>, Result extends Recipe<?>> {
```
OroArmor:
```suggestion
	public static abstract class Builder<SELF extends Builder<SELF, RESULT>, RESULT extends Recipe<?>> {
```

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/VanillaRecipeBuilders.java
```diff
@@ -0,0 +1,407 @@
...
+		}
+
+		/**
+		 * Puts the specified items as the accepted ingredient at the specified key.
+		 *
+		 * @param items the items as ingredient
+		 * @return this builder
+		 * @see #ingredient(Ingredient)
+		 */
+		public ShapelessRecipeBuilder ingredient(ItemConvertible... items) {
```
OroArmor: For both this and the `ItemStack` version, do all the parameters count as one ingredient and are interchangeable, or are they all separate? I would clarify the javadocs on this

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/VanillaRecipeBuilders.java
```diff
@@ -0,0 +1,407 @@
...
+import net.minecraft.util.Identifier;
+import net.minecraft.util.collection.DefaultedList;
+
+/**
+ * Provides some recipe builders for Vanilla recipes.
+ *
+ * @version 1.0.0
+ * @since 1.0.0
+ */
+public final class VanillaRecipeBuilders {
```
OroArmor: This class has a lot of impl in it. Maybe breaking into two classes, one impl and this with interfaces for the builders. Most of the one line methods can stay, and the furnace recipes check can be in the impl too

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/VanillaRecipeBuilders.java
```diff
@@ -0,0 +1,407 @@
...
+import net.minecraft.recipe.StonecuttingRecipe;
+import net.minecraft.tag.TagKey;
+import net.minecraft.util.Identifier;
+import net.minecraft.util.collection.DefaultedList;
+
+/**
+ * Provides some recipe builders for Vanilla recipes.
+ *
+ * @version 1.0.0
+ * @since 1.0.0
```
OroArmor: i havent seen this in other prs, is this something that should be done more?

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/serializer/QuiltRecipeSerializer.java
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Represents a recipe serializer for mods to implement.
+ * <p>
+ * This will allow serialization to JSON of recipes. Useful for recipe dumping.
+ *
+ * @param <T> the recipe
+ * @version 1.0.0
+ * @since 1.0.0
+ */
+public interface QuiltRecipeSerializer<T extends Recipe<?>> extends RecipeSerializer<T> {
```
OroArmor, line 31: To ensure all the serializers have this, should 1. interface injection be used (once the generic issue is fixed) and/or 2. a runtime check to warn that the API isn't implemented?

OroArmor (changes requested):

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/RecipeLoadingEvents.java
```diff
@@ -0,0 +1,285 @@
...
+import org.quiltmc.qsl.base.api.event.Event;
+import org.quiltmc.qsl.base.api.event.EventAwareListener;
+
+/**
+ * Represents the recipe loading events.
+ * <p>
+ * Triggered when the recipes are loaded.
+ *
+ * @version 1.0.0
+ * @since 1.0.0
```
TheGlitch76: is this really necessary?
it isn't in our style guide to use this currently, and i don't see a reason to use them

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/RecipeLoadingEvents.java
```diff
@@ -0,0 +1,285 @@
...
+	@FunctionalInterface
+	public interface RecipeLoadingCallback extends EventAwareListener {
+		/**
+		 * Called when recipes are loaded.
+		 * <p>
+		 * {@code handler} is used to add recipes into the {@linkplain net.minecraft.recipe.RecipeManager recipe manager}.
+		 *
+		 * @param handler the recipe handler
+		 */
+		void onLoadRecipe(RecipeHandler handler);
```
TheGlitch76: The javadoc implies that this is ran once when all recipes are loaded, but the method name implies this is called each time a recipe is loaded. Please clarify both.

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/RecipeLoadingEvents.java
```diff
@@ -0,0 +1,285 @@
...
+	 * Callback called to modify or replace recipes after recipes are loaded.
+	 */
+	@FunctionalInterface
+	public interface RecipeModifyCallback extends EventAwareListener {
+		/**
+		 * Called after recipes are loaded to modify and replace recipes.
+		 *
+		 * @param handler the recipe handler
+		 */
+		void onModifyRecipe(RecipeHandler handler);
```
TheGlitch76: same as above, this seems to be an API for modifying recipes, but the method name implies this is used to listen for when a mod modifies a recipe.

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/RecipeLoadingEvents.java
```diff
@@ -0,0 +1,285 @@
...
+	 * Callback called to remove recipes after recipes are loaded.
+	 */
+	@FunctionalInterface
+	public interface RecipeRemoveCallback extends EventAwareListener {
+		/**
+		 * Called after recipes are loaded to remove recipes.
+		 *
+		 * @param handler the recipe handler
+		 */
+		void onRecipeRemovePhase(RecipeHandler handler);
```
TheGlitch76: `removeRecipes`?

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/RecipeManagerHelper.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+public final class RecipeManagerHelper {
+	private RecipeManagerHelper() {
+		throw new UnsupportedOperationException("RecipeManagerHelper only contains static definitions.");
+	}
+
+	/**
+	 * Registers a static recipe.
+	 * <p>
+	 * A static recipe is a recipe that is registered at mod startup (or later) and is kept during the whole lifecycle
```
TheGlitch76, line 34: Clarify when recipes can no longer be registered

TheGlitch76 (changes requested):
Some things i noticed from a quick glance at the API.

Also, you use the words `load` and `register` for creating recipes interchangably in this API, but I think you should pick one word and use it throughout the API's javadocs and method names.

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/VanillaRecipeBuilders.java
LambdAurora: Definitely should be done more imo
Was kind of forgotten.

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/VanillaRecipeBuilders.java
LambdAurora: A reason for this?
Is it just to avoid possible confusion with actual types?

OroArmor: mostly yeah. It was slightly hard to understand at first

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/serializer/QuiltRecipeSerializer.java
LambdAurora: Generic issue doesn't seem to be a high priority in loom, so might do the runtime check instead.

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/RecipeManagerHelper.java
LambdAurora: Er, there's no actual recipe registration, it's never locked at one point.

I probably should say that any recipe registered while the game is running  will only apply for the next data pack reload.

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/VanillaRecipeBuilders.java
LambdAurora: > `Adds item stacks as an ingredient.`
for the item stack version

I thought it was clear that it would be one ingredient?

library/data/recipe/build.gradle
TheGlitch76: this isn't resolved

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/RecipeLoadingEvents.java
```diff
@@ -0,0 +1,288 @@
...
+
+		/**
+		 * This interface should not be extended by users.
+		 */
+		@ApiStatus.NonExtendable
+		interface RecipeHandler {
+			/**
+			 * Registers a recipe into the {@link net.minecraft.recipe.RecipeManager}.
+			 * <p>
+			 * The recipe factory is only called if the recipe can be registered.
```
TheGlitch76: perhaps it would be good to list reasons why a recipe cannot be loaded here?

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/RecipeManagerHelper.java
TheGlitch76: that would be good

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/impl/ImmutableMapBuilderUtil.java
```diff
@@ -0,0 +1,84 @@
...
+	static {
+		try {
+			Field field = ImmutableMap.Builder.class.getDeclaredField("entries");
+			field.setAccessible(true);
+			ENTRIES_GETTER = MethodHandles.lookup().unreflectGetter(field);
+			field = ImmutableMap.Builder.class.getDeclaredField("size");
+			field.setAccessible(true);
+			SIZE_GETTER = MethodHandles.lookup().unreflectGetter(field);
+		} catch (NoSuchFieldException | IllegalAccessException e) {
+			LOGGER.error("Could not access ImmutableMap$Builder entries or size fields. Which is necessary for the Recipe API.");
```
TheGlitch76:
```suggestion
			LOGGER.error("Could not access ImmutableMap$Builder entries or size fields, which is necessary for the Recipe API.");
```

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/RecipeLoadingEvents.java
```diff
@@ -0,0 +1,288 @@
...
+	/**
+	 * Event to remove recipes while the {@link net.minecraft.recipe.RecipeManager} is being built.
+	 */
+	public static final Event<RemoveRecipesCallback> REMOVE = Event.create(RemoveRecipesCallback.class,
+			callbacks -> handler -> {
+				for (var callback : callbacks) {
+					callback.removeRecipes(handler);
+				}
+			});
+
```
TheGlitch76: Perhaps it would be good to clarify the ordering of these events in respect to each other? I can assume it's ADD, MODIFY, REMOVE, but that would be nice to be official both here and in `RegistryManagerHelper`

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/builder/RecipeBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+	 * Builds the recipe.
+	 *
+	 * @param id    the identifier of the recipe
+	 * @param group the group of the recipe
+	 * @return the shaped recipe
+	 */
+	public RESULT build(Identifier id, String group) {
+		Objects.requireNonNull(this.output, "The output stack cannot be null.");
+		return null;
+	}
```
TheGlitch76: is this intentional?

Would be a good idea to mark with `@Nullable` if the result may be null

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/builder/ShapedRecipeBuilder.java
```diff
@@ -0,0 +1,132 @@
...
+	 *                A character represents an ingredient and space is no ingredient
+	 */
+	public ShapedRecipeBuilder(String[] pattern) {
+		this.pattern = pattern;
+		this.width = pattern[0].length();
+		this.height = pattern.length;
+	}
+
+	/**
+	 * Puts the specified ingredient at the specified key.
```
TheGlitch76:
```suggestion
	 * Associates an ingredient with a specified key.
```

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/builder/ShapedRecipeBuilder.java
```diff
@@ -0,0 +1,132 @@
...
+		boolean success = false;
+
+		for (String line : this.pattern) {
+			for (int i = 0; i < line.length(); i++) {
+				char c = line.charAt(i);
+
+				if (c == key) {
+					this.ingredients.put(key, ingredient);
+
+					success = true;
```
TheGlitch76, line 69: maybe this warn/throw in dev if an ingredient is never found?

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/builder/VanillaRecipeBuilders.java
```diff
@@ -0,0 +1,161 @@
...
+	 * Returns the list of ingredients for shaped crafting recipes.
+	 *
+	 * @param pattern the pattern of the shaped crafting recipe
+	 * @param keys    the keys and ingredients of the recipe
+	 * @param width   the width of the shaped crafting recipe
+	 * @param height  the height of the shaped crafting recipe
+	 * @return the ingredients
+	 * @throws IllegalStateException if a key has no assigned ingredient or if there is an ingredient but no assigned key
+	 */
+	public static DefaultedList<Ingredient> getIngredients(String[] pattern, Char2ObjectMap<Ingredient> keys, int width, int height) {
```
TheGlitch76, line 64: Is fastutil in public api really necessary here?

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/package-info.java
```diff
@@ -0,0 +1,27 @@
+/**
+ * <h2>The Recipe API.</h2>
+ * <p>
+ * <h3>Quick note about vocabulary in the Recipe API:</h3>
+ * <ul>
+ *     <li>A static recipe is a recipe which is registered in the API once and is automatically added to
```
TheGlitch76:
```suggestion
 *     <li>A static recipe is a recipe which is registered with the API once and is automatically added to
```

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/package-info.java
```diff
@@ -0,0 +1,27 @@
...
+ * <h3>{@link org.quiltmc.qsl.recipe.api.RecipeLoadingEvents RecipeLoadingEvents}</h3>
+ * Contains some events to register, modify, and remove recipes.
+ * <p>
+ * <h3>When to use the Recipe API?</h3>
+ * This API targets specific use-cases:
+ * <ul>
+ *     <li>Dynamic recipe registration, in the case the recipes cannot be determined at compile-time.</li>
+ *     <li>Modification and removal of recipes.</li>
+ * </ul>
+ * This API is <b>NOT</b> supposed to be used for known at compile-time data-generation, please use the appropriate tools instead.
```
TheGlitch76, line 25: Perhaps "recipe API" isn't a specific enough name then.

"Runtime Registry Registration, Revision, and Removal API"
aka RRRR API /j

TheGlitch76 (changes requested):

library/data/recipe/build.gradle
LambdAurora: how in hell did it got not pushed

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/builder/VanillaRecipeBuilders.java
LambdAurora: Is it really that bad?

Given that the type exists already, and Project Valhalla still far, it kinds of feel wrong to use a Map.

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/builder/RecipeBuilder.java
LambdAurora: This is intentional as-in it's supposed to be overriden in recipe builder implementations.

LambdAurora: Made it abstract instead.

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/package-info.java
LambdAurora: Eh, recipe API will be home to more recipe-related APIs in the future.

OroArmor: Could I use the constructors here to make the recipes for a datagen API? or should the datagen API have its own thing?

LambdAurora: Since the recipe serializers are patched in this module, sure, you can.
Though, Minecraft's datagen stuff for recipes are more advanced since it also includes recipe unlock advancements.

OroArmor: Would the datagen api then use a wrapper around this method + add the unlock advancement?

LambdAurora: Could for non-Vanilla recipes, but that's a choice for the developer.

For recipe datagen there's the `RecipeJsonProvider` interface which has a method to get the recipe serializer and a toJson/serialize method.

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/builder/VanillaRecipeBuilders.java
TheGlitch76: it's technically poor practice, but eh--it's fine

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/RecipeManagerHelper.java
```diff
@@ -0,0 +1,91 @@
...
+	 * of the game.
+	 * <p>
+	 * Static recipes are automatically added to the {@linkplain net.minecraft.recipe.RecipeManager recipe manager}
+	 * when recipes are loaded, and only is added if no other recipe with the same identifier is already loaded.
+	 * <p>
+	 * Only applies after a data pack reload.
```
TheGlitch76:
```suggestion
	 * Static recipes can be added at any time, but are only applied after a data pack reload.
```

TheGlitch76 (approved):

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/serializer/QuiltRecipeSerializer.java
LambdAurora: A warning log has been added in dev environment or with the `quilt.recipe.debug` property.

library/data/recipe/src/main/java/org/quiltmc/qsl/recipe/api/RecipeManagerHelper.java
```diff
@@ -0,0 +1,91 @@
...
+	 * <p>
+	 * Triggered after {@link #addRecipes(RecipeLoadingEvents.AddRecipesCallback)}
+	 * and {@link #modifyRecipes(RecipeLoadingEvents.ModifyRecipesCallback)}.
+	 *
+	 * @param callback the callback to remove recipes
+	 * @see RecipeLoadingEvents#REMOVE
+	 */
+	public static void removeRecipes(RecipeLoadingEvents.RemoveRecipesCallback callback) {
+		RecipeLoadingEvents.REMOVE.register(callback);
+	}
```
BasiqueEvangelist, line 90: Do we really need these helpers? Isn't it simple enough just to register a handler to event?

LambdAurora: Those are mostly present because of `registerStaticRecipe`, I didn't like having one util alone and all the main events kind of separate.
And having the events definitions in this class sounded quite wrong.

OroArmor (approved):

PR #87, opened 2022-03-12
Add a way to register new resource pack providers.

LambdAurora:
As the title says.

Also contains some extra stuff like a QuiltResourcePack interface that is injected in ResourcePack, this allows to make stuff a bit more generic and expose some new stuff to API.

This is not the virtual resource pack API, but it's a step closer to it.

Use cases:
 - visible virtual resource packs, but only the registration of the resource pack profile is provided
 - global datapack directory, as it would use a resource pack provider

The testmod includes a test virtual resource pack profile, it generates a random dirt and pack.png textures.


library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/GroupResourcePack.java
```diff
@@ -60,8 +61,21 @@ public GroupResourcePack(ResourceType type, List<? extends ResourcePack> packs)
 						.add(pack)));
 	}
 
+	/**
+	 * {@return the list of resource packs contained within this group}
```
TheGlitch76: unnecessary braces?

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/QuiltResourcePack.java
```diff
@@ -0,0 +1,41 @@
...
+	/**
+	 * {@return a display name for this resource pack}
+	 */
+	Text getDisplayName();
+
+	/**
+	 * Gets the activation type of this resource pack.
+	 * <p>
+	 * This only serves as a hint as ultimately the {@link net.minecraft.resource.pack.ResourcePackProfile}
+	 * mays have the last word.
```
TheGlitch76:
```suggestion
	 * has the last word.
```

TheGlitch76: it's not immediately clear what you mean by it having "the last word". Do you mean that if you set an ResourcePackProfile activation type, this value is ignored?

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/QuiltResourcePack.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package org.quiltmc.qsl.resource.loader.api;
+
+import net.minecraft.text.Text;
+
+/**
+ * Represents a resource pack with extended metadata, injected into {@link net.minecraft.resource.pack.ResourcePack}.
+ */
+public interface QuiltResourcePack {
```
TheGlitch76, line 29: Is this meant to be ApiStatus.NonExtendable then?

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/GroupResourcePack.java
LambdAurora: Necessary braces, this is the inline return notation, the inline return notation has specific semantics, in addition of filling the Returns field, it also generates a summary/description as `Returns <content>.`.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/QuiltResourcePack.java
LambdAurora: Yeah, if your ResourcePackProfile says "always enabled", the value will be ignored.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/QuiltResourcePack.java
LambdAurora: Meh, ResourcePack extends it via interface injection, so it's technically extendable.

OroArmor (approved):

TheGlitch76:
merge conflicts

LambdAurora:
Merge conflict fixed.

TropheusJ (approved):

TheGlitch76 (commented):
I don't really have time to review this again, so in the interest of not holding this PR up i'm dismissing my review.

Per the PR policy, any member of the QSL team can approve this PR (since Oro has already approved), and this is good to go for FCP.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/QuiltResourcePack.java
TheGlitch76: Use the annotation please :tiny_potato:

LambdAurora: Done!

Leo40Git (approved):

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/QuiltResourcePack.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Represents a resource pack with extended metadata, injected into {@link net.minecraft.resource.pack.ResourcePack}.
+ */
+@InjectedInterface(ResourcePack.class)
+public interface QuiltResourcePack {
+	/**
+	 * {@return a display name for this resource pack}
+	 */
+	Text getDisplayName();
```
BasiqueEvangelist: Why is this not `default`?

LambdAurora: This is an injected interface, the default implementation is in the mixin.
If a `default` was provided it wouldn't work well.

BasiqueEvangelist: Yes, but this means that all mods that have classes that implement ResourcePack will need to also implement QuiltResourcePack, otherwise javac will complain.

LambdAurora: Yes.

Not sure what the proper path should be.
Should I add a simple default dummy return value?

OroArmor (approved):

PR #88, opened 2022-03-18
Fix Command API registration method in 1.19 for registry-dependent arguments

LambdAurora:
1.19 introduced a breaking change to how registry-dependent arguments are fetched, it requires passing a new class which isn't passed through the command API currently.

This needs more work to fix it on client commands too.


LemmaEOF (approved):

LemmaEOF (approved):

LemmaEOF:
...good job for approving twice gh lmao

LambdAurora:
Client commands have been fixed too now, change includes:
 - breaking the DISPATCHER field into a method as it has to be rebuilt every time the client connects to a server/world,
 - have a similar fix as the server commands.

Leo40Git (approved):

library/command/client_command/src/main/java/org/quiltmc/qsl/command/api/client/ClientCommandManager.java
```diff
@@ -58,20 +60,23 @@
...
 	/**
-	 * The command dispatcher that handles client command registration and execution.
+	 * {@return the command dispatcher that handles client command registration and execution}
+	 *
+	 * @see ClientCommandRegistrationCallback client command registration
 	 */
-	public static final CommandDispatcher<QuiltClientCommandSource> DISPATCHER = new CommandDispatcher<>();
-
-	private ClientCommandManager() {
+	public static CommandDispatcher<QuiltClientCommandSource> getDispatcher() {
```
williambl: Should have been like this from the start, honestly.

library/command/client_command/src/main/java/org/quiltmc/qsl/command/impl/client/ClientCommandInternals.java
```diff
@@ -90,7 +97,7 @@ public static boolean executeCommand(String message) {
 
 		// The interface is implemented on ClientCommandSource with a mixin.
 		// noinspection ConstantConditions
-		QuiltClientCommandSource commandSource = (QuiltClientCommandSource) client.getNetworkHandler().getCommandSource();
+		var commandSource = (QuiltClientCommandSource) client.getNetworkHandler().getCommandSource();
```
williambl: Not pertinent but what's the stance on interface injection in QSL, to eliminate things like this? I saw that it was used once in some other module and then removed after there were issues, are we avoiding it for now?

williambl (approved):
Looks good.

library/command/client_command/src/main/java/org/quiltmc/qsl/command/impl/client/ClientCommandInternals.java
LambdAurora: Those issues were fixed, so we can definitely use injected interfaces now

OroArmor: Let's get interface injection here and then I'll approve

OroArmor (commented):
Add some interface injection

OroArmor (approved):

PR #89, opened 2022-03-21
Fix registry monitor with registration in 1.18.2, fix Block.asItem() …

LambdAurora:
…with modded.

This oversight has not been caught when porting due to a missing test.

This PR introduces a proxy object for the registry which delays registration.

### DIFF
--- a/library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/event/DelayedRegistry.java
+++ b/library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/event/DelayedRegistry.java
@@ -191,12 +191,12 @@ public final class DelayedRegistry<T> extends MutableRegistry<T> {
 
 	@Override
 	public void resetTags() {
-		// noop
+		throw new UnsupportedOperationException("DelayedRegistry does not support resetTags.");
 	}
 
 	@Override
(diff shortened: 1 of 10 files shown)


OroArmor (approved):

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/event/DelayedRegistry.java
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	@Override
+	public boolean isKnownTag(TagKey<T> tag) {
+		return this.wrapped.isKnownTag(tag);
+	}
+
+	@Override
+	public void resetTags() {
+		// noop
```
LemmaEOF: Should this and `bindTags` throw a UOE like `set` and `replace` do?

LemmaEOF (approved):

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/event/DelayedRegistry.java
LambdAurora: Probably

Leo40Git (approved):

TheGlitch76 (approved):

PR #90, opened 2022-03-23
Add a method for quick registration in RegistryEntryContext.

LambdAurora:
Rationale: is more visible as it's directly in the context object, may avoid issues.

### DIFF
--- a/library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/event/RegistryEntryContext.java
+++ b/library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/event/RegistryEntryContext.java
@@ -50,7 +50,7 @@ public interface RegistryEntryContext<V> {
 	/**
 	 * Safely registers a new entry in the registry of this context.
 	 * <p>
-	 * Registration may be delayed with {@link RegistryMonitor#forAll(RegistryEvents.EntryAdded)}.
+	 * Registration may be delayed when called from {@link RegistryMonitor#forAll(RegistryEvents.EntryAdded)}.
 	 *
 	 * @param id    the identifier of the entry
 	 * @param value the value to register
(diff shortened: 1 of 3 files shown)


library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/event/RegistryEntryContext.java
```diff
@@ -46,4 +46,18 @@
 	 * {@return the entry's raw int identifier}
 	 */
 	int rawId();
+
+	/**
+	 * Safely registers a new entry in the registry of this context.
+	 * <p>
+	 * Registration may be delayed with {@link RegistryMonitor#forAll(RegistryEvents.EntryAdded)}.
```
TheGlitch76:
```suggestion
	 * Registration may be delayed when called from {@link RegistryMonitor#forAll(RegistryEvents.EntryAdded)}.
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/event/RegistryMonitor.java
```diff
@@ -47,7 +48,8 @@
 	 * Entries must also match the monitor's filters.
 	 * <p>
 	 * Registration inside the callback must use the {@link RegistryEntryContext#registry()} method to get the registry instance,
```
TheGlitch76:
```suggestion
 	 * Registration to the registry being iterated must use the {@link RegistryEntryContext#registry()} method to get the registry instance inside the callback,
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/event/RegistryMonitor.java
```diff
@@ -47,7 +48,8 @@
...
-	 * for example: {@code Registry.register(context.registry(), id, block);}.
+	 * or alternatively use the {@link RegistryEntryContext#register(Identifier, Object)} method,
```
TheGlitch76: I would actually suggest this as the primary method, and recommend the non-context method as an alternative

OroArmor (approved):

Leo40Git (approved):

TheGlitch76 (approved):

PR #91, opened 2022-03-25
Dimension API

NoahvdAa:
Mostly just a port of fabric's dimension API for now, with some minor changes to javadocs and the implementation.

_Sidenote:_ I'm not 100% sure whether I added the copyright headers correctly, please let me know if anything needs changing.


library/worldgen/dimension/build.gradle
```diff
@@ -0,0 +1,19 @@
...
+}
+
+qslModule {
+	moduleName = "dimension"
+	version = "1.0.0"
+	library = "worldgen"
+	moduleDependencies {
+		core {
+			api("qsl_base")
+			api("resource_loader")
```
TheGlitch76: resource loader can probably be impl

peaceheis (approved):
Looks good to me for an initial PR to work off of! Thank you Noah, especially for tests and useful comments.

library/worldgen/dimension/src/main/java/org/quiltmc/qsl/worldgen/dimension/mixin/server/EntityMixin.java
```diff
@@ -0,0 +1,45 @@
...
+import org.quiltmc.qsl.worldgen.dimension.access.EntityAccess;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+@Mixin(Entity.class)
+public class EntityMixin implements EntityAccess {
+
+	public TeleportTarget overriddenTeleportTarget = null;
```
TheGlitch76: Mark `@Unique` and add a `quilt$` prefix

TheGlitch76 (changes requested):
Thanks for contributing! Please fix these small issues and this will be good to go.

library/worldgen/dimension/build.gradle
NoahvdAa: Looks like it actually isn't needed at all, API and testmod both work fine without it.

library/worldgen/dimension/src/testmod/resources/data/quilt_dimension/dimension_type/void_type.json
```diff
@@ -0,0 +1,17 @@
...
+  "has_ceiling": false,
+  "infiniburn": "#minecraft:infiniburn_overworld",
+  "logical_height": 256,
+  "has_raids": false,
+  "respawn_anchor_works": false,
+  "bed_works": false,
+  "piglin_safe": false,
+  "height": 256,
+  "min_y": 0,
+  "effects": "fabric_dimension:void"
```
Leo40Git: This should be `quilt_dimension:void`.

Leo40Git (approved):
Looks good!

library/worldgen/dimension/build.gradle
```diff
@@ -0,0 +1,18 @@
...
+}
+
+qslModule {
+	moduleName = "dimension"
+	version = "1.0.0"
+	library = "worldgen"
+	moduleDependencies {
+		core {
+			api("qsl_base")
+			testmodOnly("lifecycle_events")
```
TheGlitch76, line 12: Does the testmod really not need resource loader?

NoahvdAa: I wasn't expecting it to work without the resource loader either, it compiles and runs fine though (and there seem to be no resource loader being used).

TheGlitch76: Did you run the testmod for the specific module, or did you run all the testmods?

NoahvdAa: Looks like I was wrong. After running just the testmod, it crashed without the resource loader. The implementation doesn't seem to need it though, so `testmodOnly`?

TheGlitch76: Yes

TheGlitch76 (approved):

library/worldgen/dimension/src/main/java/org/quiltmc/qsl/worldgen/dimension/mixin/server/ServerPlayerEntityMixin.java
```diff
@@ -0,0 +1,39 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(ServerPlayerEntity.class)
+public class ServerPlayerEntityMixin {
+
+	@Inject(method = "createEndSpawnPlatform", at = @At("HEAD"), cancellable = true, allow = 1)
+	public void createEndSpawnPlatform(ServerWorld world, BlockPos centerPos, CallbackInfo ci) {
+		// Don't generate the platform below the players' feet if they are being teleported to a custom location
+		if (((EntityAccess) this).getTeleportTarget() != null)
+			ci.cancel();
```
LambdAurora: Codestyle

```suggestion
		if (((EntityAccess) this).getTeleportTarget() != null) {
			ci.cancel();
		}
```

library/worldgen/dimension/src/main/java/org/quiltmc/qsl/worldgen/dimension/impl/QuiltDimensionsImpl.java
```diff
@@ -0,0 +1,49 @@
...
+@ApiStatus.Internal
+public class QuiltDimensionsImpl {
+
+	// Static only-class, no instantiation necessary!
+	private QuiltDimensionsImpl() {
+	}
+
+	@SuppressWarnings("unchecked")
+	public static <E extends Entity> E teleport(Entity entity, ServerWorld destinationWorld, TeleportTarget location) {
+		Preconditions.checkArgument(Thread.currentThread() == ((ServerWorld) entity.getWorld()).getServer().getThread(), "This method may only be called from the main server thread");
```
LambdAurora: Nit-picky, but this would improve crash reports in some cases!

```suggestion
		Preconditions.checkArgument(Thread.currentThread() == entity.getServer().getThread(), "This method may only be called from the main server thread");
```

LambdAurora (approved):
Just some very nit-picky suggestions.

library/worldgen/dimension/src/main/java/org/quiltmc/qsl/worldgen/dimension/impl/QuiltDimensionsImpl.java
NoahvdAa: A similar check is already done from the API method, should that one be replaced by this and moved into here (the impl)?

LambdAurora: Edited the review, turns out there's a getServer method in entity already, so no need to use fragile casts.

library/worldgen/dimension/src/main/java/org/quiltmc/qsl/worldgen/dimension/api/QuiltDimensions.java
```diff
@@ -0,0 +1,63 @@
...
+	 * @param <E> the type of the entity that is being teleported
+	 * @return the teleported entity in the destination dimension, which will either be a new entity or teleported,
+	 * 		depending on the type of entity
+	 * @apiNote this method may only be called on the main server thread
+	 */
+	@Nullable
+	public static <E extends Entity> E teleport(@NotNull Entity entity, @NotNull ServerWorld destinationWorld, @Nullable TeleportTarget location) {
+		Preconditions.checkNotNull(entity, "entity may not be null");
+		Preconditions.checkNotNull(destinationWorld, "destinationWorld may not be null");
+		Preconditions.checkArgument(!destinationWorld.isClient, "This method may only be called from the server side");
```
LambdAurora:
```suggestion
		Preconditions.checkArgument(!destinationWorld.isClient(), "This method may only be called from the server side");
```

Last nit-pick, the isClient field should never be used as it could be made private at any update, while the method is kind of more safe.

NoahvdAa:
Thanks, fixed.

PR #95, opened 2022-04-10
Initial registry sync implementation

### LINKED ISSUE #28: The new registry sync format.
This issue pertains to discussion about the new registry sync format.

There have been a few issues with the registry sync fabric uses that causes it to reach the maximum packet size quickly.

A few of these issues include:
- The overhead of the NBT format in serialization for the packet
- The inclusion of the namespace inside all Minecraft namespaced identifiers.
- The map structure of the packet using key -> value for every namespace and string takes up extra room.
- Where a sequence of all identifiers inside the packet are of the same namespace, the namespace is included in the identifier.
- Some registries get many more registrations than others, meaning just a few registries in the same packet fill up the entire payload of the packet.
- Overloaded registries cannot be split if they hit the max payload size.

Given the above issues, I feel we can greatly reduce the size of the packet with some optimizations.

Patbox:
Initial implementation of registry sync for qsl. 

While it is functional, I'm still not sure about api structure.
Other thing to consider, is proper blockstate sync, which isn't implemented yet.

Additionally I though about usage of small hack for having sync before play stage on server, but on it on client for nicer experience while keeping support for proxy software like velocity.


LambdAurora:
You might want to execute `./gradlew :core:registry:applyLicenses`.

library/core/registry/src/main/java/org/quiltmc/qsl/registry/mixin/client/MinecraftClientMixin.java
```diff
@@ -0,0 +1,24 @@
...
+import org.quiltmc.qsl.registry.impl.sync.SynchronizedRegistry;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(MinecraftClient.class)
+public class MinecraftClientMixin {
+	@Inject(method = "disconnect(Lnet/minecraft/client/gui/screen/Screen;)V", at = @At("TAIL"))
+	private void polymer_onDisconnect(CallbackInfo ci) {
```
LambdAurora: This looks suspicious :p

library/core/registry/src/main/java/org/quiltmc/qsl/registry/mixin/client/MinecraftClientMixin.java
```diff
@@ -0,0 +1,24 @@
...
+		for (var reg : Registry.REGISTRIES) {
+			if (reg instanceof SynchronizedRegistry registry && registry.quilt$requiresSyncing()) {
+				System.out.println("Restored snapshot of " + ((Registry<Registry<?>>) Registry.REGISTRIES).getId(reg));
```
LambdAurora: I don't think this is wanted.

library/core/registry/src/main/java/org/quiltmc/qsl/registry/mixin/client/ConnectScreenMixin.java
```diff
@@ -0,0 +1,24 @@
...
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(ConnectScreen.class)
+public class ConnectScreenMixin {
+	@Inject(method = "connect(Lnet/minecraft/client/MinecraftClient;Lnet/minecraft/client/network/ServerAddress;)V", at = @At("HEAD"))
+	private void quilt$snapshotRegistry(MinecraftClient client, ServerAddress address, CallbackInfo ci) {
+		for (var reg : Registry.REGISTRIES) {
+			if (reg instanceof SynchronizedRegistry registry && registry.quilt$requiresSyncing()) {
+				System.out.println("Created snapshot of " + ((Registry<Registry<?>>) Registry.REGISTRIES).getId(reg));
```
LambdAurora: Mmh, leftover debug print.

library/core/registry/src/main/java/org/quiltmc/qsl/registry/mixin/SimpleRegistryMixin.java
```diff
@@ -51,8 +90,121 @@ protected SimpleRegistryMixin(RegistryKey<? extends Registry<V>> key, Lifecycle
 			at = @At("RETURN")
 	)
 	private void quilt$invokeEntryAddEvent(int rawId, RegistryKey<V> key, V entry, Lifecycle lifecycle, boolean checkDuplicateKeys,
-	                                       CallbackInfoReturnable<Holder<V>> cir) {
+										   CallbackInfoReturnable<Holder<V>> cir) {
```
LambdAurora: Extremely nit-picky but I would keep the spaces from smart-tabs as it keeps alignment.

library/core/registry/src/main/java/org/quiltmc/qsl/registry/mixin/BlocksMixin.java
```diff
@@ -0,0 +1,35 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(Blocks.class)
+public abstract class BlocksMixin {
+	@Inject(method = "<clinit>", at = @At("RETURN"))
+	private static void onInit(CallbackInfo ci) {
+		RegistryMonitor.create(Registry.BLOCK)
+				.forUpcoming(context -> context.value().getStateManager().getStates().forEach(Block.STATE_IDS::add));
```
LambdAurora: If you use forUpcoming you can use the event directly?

Not sure which way should be the default.

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/ServerPackets.java
```diff
@@ -0,0 +1,46 @@
...
+import net.minecraft.util.Identifier;
+import org.jetbrains.annotations.ApiStatus;
+
+@ApiStatus.Internal
+public final class ServerPackets {
+	public static final Identifier HELLO = id("registry/hello");
+
+	/**
+	 * Sets registry next data packets belong to
+	 * {
```
LambdAurora, line 50: Code stuff like this probably should be wrapped around `<pre><code>` blocks.

TheEpicBlock (approved):

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/RegistryLibInitializer.java
```diff
@@ -0,0 +1,31 @@
...
+package org.quiltmc.qsl.registry.impl;
+
+import net.fabricmc.api.EnvType;
+import net.fabricmc.api.Environment;
+import net.fabricmc.loader.api.ModContainer;
+import org.quiltmc.qsl.base.api.entrypoint.client.ClientModInitializer;
+import org.quiltmc.qsl.registry.impl.sync.ClientRegistrySync;
+
+@Environment(EnvType.CLIENT)
+public class RegistryLibInitializer implements ClientModInitializer {
```
OroArmor:
```suggestion
public class RegistrySyncClientInitializer implements ClientModInitializer {
```
Dont forget the file name too

library/core/registry/src/testmod/java/org/quiltmc/qsl/registry/test/RegistryLibSyncTest.java
```diff
@@ -0,0 +1,100 @@
...
+import net.minecraft.util.registry.Registry;
+import org.quiltmc.qsl.base.api.entrypoint.ModInitializer;
+import org.quiltmc.qsl.lifecycle.api.client.event.ClientLifecycleEvents;
+import org.quiltmc.qsl.lifecycle.api.event.ServerLifecycleEvents;
+
+import java.nio.file.Files;
+import java.nio.file.StandardOpenOption;
+
+
+public class RegistryLibSyncTest implements ModInitializer {
```
OroArmor, line 49: I would make a comment or two about what this test does. It seems to make sure there is a client server mismatch, but I'm not 100% sure

OroArmor (changes requested):

library/core/registry/src/testmod/java/org/quiltmc/qsl/registry/test/RegistryLibSyncTest.java
LambdAurora: I also would add `@ApiStatus.Internal` to initializers.

Patbox: It forces rawId mismatch to make sure it works correctly (with identifiers being the same, since it's base for sync)

Patbox: > @ApiStatus.Internal

Not sure if it's needed there, since it's part of test mod

OroArmor (approved):

LambdAurora (approved):

TheGlitch76:
Hi, I will probably not get to this until after the Beta launch, but I do intend to review this ASAP.

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/sync/RegistrySynchronization.java
```diff
@@ -0,0 +1,101 @@
...
+
+	/**
+	 * Checks if registry is optional
+	 */
+	public static <T> boolean isEntryOptional(SimpleRegistry<T> registry, T entry) {
+		return RegistryFlag.isOptional(SynchronizedRegistry.as(registry).quilt$getEntryFlag(entry));
+	}
+
+	/**
+	 * Blocks entry from being synced
```
TropheusJ: took me a good 30 seconds to realize this means `Blocks` as in `prevents` and not a Minecraft block - maybe change to `Prevents`?

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/sync/RegistrySynchronization.java
```diff
@@ -0,0 +1,101 @@
...
+	 */
+	public static <T> void setEntrySkipped(SimpleRegistry<T> registry, T entry) {
+		RegistryFlag.setEntry(registry, entry, RegistryFlag.SKIP);
+	}
+
+	/**
+	 * Blocks entry from being synced
```
TropheusJ: here too

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/client/ClientRegistrySync.java
```diff
@@ -0,0 +1,186 @@
...
+		if (registry instanceof SynchronizedRegistry synchronizedRegistry) {
+			currentRegistry = synchronizedRegistry;
+			currentCount = count;
+			currentFlags = flags;
+			currentRegistryId = identifier;
+			syncMap = new HashMap<>();
+		} else if (RegistryFlag.isOptional(flags)) {
+			optionalRegistry = true;
+		} else {
+			LOGGER.warn("Trying to sync registry " + identifier + " which doesn't " + (registry == null ? "support it!" : "exist!"));
```
TropheusJ: swapped logic on this ternary?

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/sync/RegistrySynchronization.java
```diff
@@ -0,0 +1,101 @@
...
+import org.quiltmc.qsl.registry.impl.sync.SynchronizedRegistry;
+
+/**
+ * Methods for manipulation of registry sync
+ */
+@ApiStatus.Experimental
+public final class RegistrySynchronization {
+
+	/**
+	 * Marks registry status as dirty, rebuilding its sync data
```
TheGlitch76: The javadoc says that this method rebuilds sync data, but the implementation does no such thing!
```suggestion
	 * Marks registry status as dirty, requiring its synchronization data to be rebuilt.
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/ServerPackets.java
```diff
@@ -0,0 +1,87 @@
...
+ */
+
+package org.quiltmc.qsl.registry.impl.sync;
+
+import it.unimi.dsi.fastutil.ints.IntSet;
+import net.minecraft.util.Identifier;
+import org.jetbrains.annotations.ApiStatus;
+
+/**
+ * Identifiers of packets send by server
```
TheGlitch76:
```suggestion
 * Identifiers of packets sent by server
```

library/core/registry/src/testmod/java/org/quiltmc/qsl/registry/test/RegistryLibSyncTest.java
```diff
@@ -0,0 +1,112 @@
...
+import org.quiltmc.qsl.registry.impl.sync.SynchronizedRegistry;
+
+import java.nio.file.Files;
+import java.nio.file.StandardOpenOption;
+
+/**
+ * Items/Blocks are registered in different order on client/server to make sure sync works correctly
+ * Server also gets its own entry that shouldn't block client from joining
+ */
+public class RegistryLibSyncTest implements ModInitializer {
```
TheGlitch76, line 49: There is a lot of stuff in this API that needs covered by tests:
- Registry sync for registries added by mods
- Optional entries and registries
- Skipped entries and registries

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/ClientPackets.java
```diff
@@ -0,0 +1,53 @@
...
+ * limitations under the License.
+ */
+
+package org.quiltmc.qsl.registry.impl.sync;
+
+import net.minecraft.util.Identifier;
+import org.jetbrains.annotations.ApiStatus;
+
+/**
+ * Identifiers of packets send by server
```
TheGlitch76:
```suggestion
 * Identifiers of packets sent by server
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/sync/RegistrySynchronization.java
```diff
@@ -0,0 +1,101 @@
...
+	 */
+	public static void markDirty(SimpleRegistry registry) {
+		SynchronizedRegistry.as(registry).quilt$markDirty();
+	}
+
+	/**
+	 * Makes registry synchronize with client
+	 */
+	public static void synchronize(SimpleRegistry<?> registry) {
+		SynchronizedRegistry.as(registry).quilt$requiresSyncing();
```
BasiqueEvangelist: Doesn't `quilt$requiresSyncing` just return a boolean?
```suggestion
		SynchronizedRegistry.as(registry).quilt$markForSync();
```

BasiqueEvangelist:
Could you add a method to the API which resyncs registries with a player? I'm using an FAPI impl method in DynReg right now, but that won't work with QSL...

Also, are you going to put saving registries to disk in a different PR?

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/sync/RegistrySynchronization.java
Patbox: Ah right, somehow missed that

Patbox:
> Could you add a method to the API which resyncs registries with a player? I'm using an FAPI impl method in DynReg right now, but that won't work with QSL...
> 
> Also, are you going to put saving registries to disk in a different PR?

Sounds like something I could add. My long term I plan on moving registry sync to it's own "virtual" networking stage (kinda like what polymer does currently) to allow nicer interaction with vanilla clients (and possibly fabric ones). However I need to make api for that first

Patbox:
For saving on disk, is it really needed? Outside of Status Effect (which I might pr alternative/proper fix later), I don't think there is real need for it

BasiqueEvangelist:
> For saving on disk, is it really needed? Outside of Status Effect (which I might pr alternative/proper fix later), I don't think there is real need for it

Ah right, they removed the need for biomes to be persisted, I forgot about that.

> Sounds like something I could add. My long term I plan on moving registry sync to it's own "virtual" networking stage (kinda like what polymer does currently) to allow nicer interaction with vanilla clients (and possibly fabric ones). However I need to make api for that first

Nice!

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/sync/RegistrySynchronization.java
```diff
@@ -0,0 +1,101 @@
...
+	 * Marks registry status as dirty, rebuilding its sync data
+	 */
+	public static void markDirty(SimpleRegistry registry) {
+		SynchronizedRegistry.as(registry).quilt$markDirty();
+	}
+
+	/**
+	 * Makes registry synchronize with client
+	 */
+	public static void synchronize(SimpleRegistry<?> registry) {
```
TheGlitch76: i think markForSync makes more sense as a name, and that seems to be what you use in the rest of the codebase.

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/RegistryFlag.java
```diff
@@ -0,0 +1,51 @@
...
+package org.quiltmc.qsl.registry.impl.sync;
+
+import net.minecraft.util.Identifier;
+import net.minecraft.util.registry.SimpleRegistry;
+import org.jetbrains.annotations.ApiStatus;
+
+/**
+ * Registry flags used to determine how registry (entries) should be handled by sync
+ */
+@ApiStatus.Experimental
```
TheGlitch76: No need to mark something that isn't API as experimental API

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/ServerPackets.java
```diff
@@ -0,0 +1,87 @@
...
+ */
+@ApiStatus.Internal
+public final class ServerPackets {
```
TheGlitch76, line 27: IMO these channels could all use more descriptive names

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/ServerPackets.java
```diff
@@ -0,0 +1,87 @@
...
+	 * <pre><code>
+	 * {
+	 *   Count: VarInt
+	 *   [
+	 *     Supported Version: VarInt
+	 *   ]
+	 * }
+	 * </code></pre>
+	 */
+	public static final Identifier HELLO = id("registry/hello");
```
TheGlitch76: registry_sync/handshake
since this is where the client + server negotiate the version of registry sync to use

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/ServerPackets.java
```diff
@@ -0,0 +1,87 @@
...
+
+	/**
+	 * Ends registry sync. No data
+	 */
+	public static final Identifier GOODBYE = id("registry/goodbye");
```
TheGlitch76: registry_sync/finish?

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/ServerPackets.java
```diff
@@ -0,0 +1,87 @@
...
+	 * Sets registry next data packets belong to
+	 * <pre><code>
+	 * {
+	 *   Registry identifier: Identifier
+	 *   Count: VarInt
+	 *   Flags: byte
+	 * }
+	 * </code></pre>
+	 */
+	public static final Identifier START = id("registry/start");
```
TheGlitch76: registry_sync/begin_registry?

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/ServerPackets.java
```diff
@@ -0,0 +1,87 @@
...
+	 * Ends registry sync. No data
+	 */
+	public static final Identifier GOODBYE = id("registry/goodbye");
+
+	/**
+	 * Sets registry next data packets belong to
+	 * <pre><code>
+	 * {
+	 *   Registry identifier: Identifier
+	 *   Count: VarInt
```
TheGlitch76: count of what?

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/ServerPackets.java
```diff
@@ -0,0 +1,87 @@
...
+	 *     [
+	 *       Path: String
+	 *       Id: VarInt
+	 *       Flags: byte
+	 *     ]
+	 *   ]
+	 * }
+	 * </code></pre>
+	 */
+	public static final Identifier DATA = id("registry/data");
```
TheGlitch76: registry_sync/registry_data

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/ServerPackets.java
```diff
@@ -0,0 +1,87 @@
...
+
+	/**
+	 * Applies changes, doesn't have any data
+	 */
+	public static final Identifier APPLY = id("registry/apply");
```
TheGlitch76: registry_sync/apply

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/ServerPackets.java
```diff
@@ -0,0 +1,87 @@
...
+	 *       Flags: byte
+	 *     ]
+	 *   ]
+	 * }
+	 * </code></pre>
+	 */
+	public static final Identifier DATA = id("registry/data");
+
+	/**
+	 * Applies changes, doesn't have any data
```
TheGlitch76: Does this mark the end of one START cycle, or is it used to apply all registry changes at once? Could use clarification in javadoc.

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/ClientPackets.java
```diff
@@ -0,0 +1,53 @@
...
+	 * </code></pre>
+	 */
+	public static final Identifier HELLO = id("registry/hello");
+
+
+	/**
+	 * Sent after registry sync failure before client disconnect
+	 * <pre><code>
+	 * {
+	 *     Registry: Identifier
```
TheGlitch76, line 43: any reason to include this if its seemingly always ignored?

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/ServerRegistrySync.java
```diff
@@ -0,0 +1,162 @@
...
+		ServerPlayNetworking.registerGlobalReceiver(ClientPackets.SYNC_FAILED, ServerRegistrySync::handleSyncFailedPacket);
+	}
+
+	// This is currently unused, as it requires "preplay" networking api to be useful in any way
+	// Why can't it use Minecraft's Login stage you might ask? That's simple, it would limit compatibility with
+	// any proxy software like Velocity (this is issue with forge for example).
+	// Ideally we should introduce new stage/api for it by creating custom ServerPlayPacketListener
+	// that just handles all of custom logic/emulation of login stage, which is fairly possible
+	// to do while keeping compatibility with vanilla clients (as long as other mods on server allow it).
+	// It's thanks to ping packets send by server and tpc/Minecraft preserving packet order!
```
TheGlitch76: what's tpc?

TheGlitch76:
```suggestion
	// It's thanks to ping packets sent by server and tpc/Minecraft preserving packet order!
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/ClientPackets.java
```diff
@@ -0,0 +1,53 @@
...
+import net.minecraft.util.Identifier;
+import org.jetbrains.annotations.ApiStatus;
+
+/**
+ * Identifiers of packets send by server
+ */
+@ApiStatus.Internal
+public final class ClientPackets {
+	/**
+	 * Response for {@link ServerPackets#HELLO}
```
TheGlitch76:
```suggestion
	 * Response for {@link ServerPackets#HELLO}. Selects the registry sync version to be used from the server's supported options.
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/ServerPackets.java
TheGlitch76, line 27: Are we sure we want to keep the registry sync packet structure an impl detail, instead of considering it part of the public API?

library/core/registry/src/main/java/org/quiltmc/qsl/registry/mixin/RegistryMixin.java
```diff
@@ -62,4 +63,25 @@ private static void onFreezeBuiltins(CallbackInfo ci) {
 		DebugChunkGeneratorAccessor.setZSideLength(zLength);
 		//endregion
 	}
+
+	@Inject(method = "<clinit>", at = @At("TAIL"))
+	private static void quilt$markRegistriesForSync(CallbackInfo ci) {
+		SynchronizedRegistry.markForSync(
```
TheGlitch76: Isn't it kind of unexpected behavior for new registries to not be synced by default?
Maybe a better solution is to sync all by default, and use RegistrySynchronization as an opt-out instead?

TheGlitch76 (changes requested):

library/core/registry/src/main/java/org/quiltmc/qsl/registry/mixin/RegistryMixin.java
Patbox: Vanilla ones are also opt in, so it made sense to me to apply that globally, but I can change that

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/ServerPackets.java
Patbox: Well structure could be part of public, but ideally mods shouldn't need it anyway. I guess documenting it somewhere is something we could look

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/ServerRegistrySync.java
Patbox: TCP, just got letters switched

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/ClientPackets.java
Patbox: In future (aka when I get around virtual networking stages api), kicking will be moved to server. Through I guess for now it can be removed and readded in later sync version

OroArmor (approved):

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/ClientPackets.java
TheGlitch76: It's fine to future-proof and keep it in.

TheGlitch76 (approved):
I feel pretty comfortable with this code now. Nice work!

Patbox:
Ok updated this pr with:
- Moved registry sync to run earlier to prevent clients from connecting when they are missing required entries/use vanilla
- Added detection of vanilla client with customisable kick message
- Added support for Fabric Registry Sync directly into this module

OroArmor (approved):

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/sync/RegistrySynchronization.java
```diff
@@ -0,0 +1,100 @@
...
+package org.quiltmc.qsl.registry.api.sync;
+
+import net.minecraft.util.Identifier;
+import net.minecraft.util.registry.SimpleRegistry;
+import org.jetbrains.annotations.ApiStatus;
+import org.quiltmc.qsl.registry.impl.sync.RegistryFlag;
+import org.quiltmc.qsl.registry.impl.sync.SynchronizedRegistry;
+
+/**
+ * Methods for manipulation of registry sync
```
LambdAurora:
```suggestion
 * Methods for manipulation of registry synchronization.
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/sync/RegistrySynchronization.java
```diff
@@ -0,0 +1,100 @@
...
+
+	/**
+	 * Marks registry status as dirty, requiring its synchronization data to be rebuilt.
+	 */
+	public static void markDirty(SimpleRegistry registry) {
+		SynchronizedRegistry.as(registry).quilt$markDirty();
+	}
+
+	/**
+	 * Makes registry synchronize with client
```
LambdAurora:
```suggestion
	 * Makes registry synchronize with clients.
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/sync/RegistrySynchronization.java
```diff
@@ -0,0 +1,100 @@
...
+	 */
+	public static void markForSync(SimpleRegistry<?> registry) {
+		SynchronizedRegistry.as(registry).quilt$markForSync();
+	}
+
+	/**
+	 * Makes registry optional for sync
```
LambdAurora:
```suggestion
	 * Marks a registry as optional for synchronization.
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/sync/RegistrySynchronization.java
```diff
@@ -0,0 +1,100 @@
...
+	 */
+	public static void setRegistryOptional(SimpleRegistry<?> registry) {
+		RegistryFlag.setRegistry(registry, RegistryFlag.OPTIONAL);
+	}
+
+	/**
+	 * Checks if registry is optional
```
LambdAurora:
```suggestion
	 * {@return {@code true} if the given registry is marked as optional, or {@code false} otherwise}
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/sync/RegistrySynchronization.java
```diff
@@ -0,0 +1,100 @@
...
+	 */
+	public static boolean isRegistryOptional(SimpleRegistry<?> registry) {
+		return RegistryFlag.isOptional(SynchronizedRegistry.as(registry).quilt$getRegistryFlag());
+	}
+
+	/**
+	 * Makes registry entry optional for sync
```
LambdAurora:
```suggestion
	 * Marks the given registry entry as optional of synchronization.
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/sync/RegistrySynchronization.java
```diff
@@ -0,0 +1,100 @@
...
+	 */
+	public static <T> void setEntryOptional(SimpleRegistry<T> registry, T entry) {
+		RegistryFlag.setEntry(registry, entry, RegistryFlag.OPTIONAL);
+	}
+
+	/**
+	 * Makes registry entry optional for sync
```
LambdAurora:
```suggestion
	 * Marks the specified registry entry as optional of synchronization.
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/sync/RegistrySynchronization.java
```diff
@@ -0,0 +1,100 @@
...
+	 */
+	public static <T> void setEntryOptional(SimpleRegistry<T> registry, Identifier identifier) {
+		RegistryFlag.setEntry(registry, identifier, RegistryFlag.OPTIONAL);
+	}
+
+	/**
+	 * Checks if registry is optional
```
LambdAurora:
```suggestion
	 * {@return {@code true} if the registry entry is marked as optional, or {@code false} otherwise}
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/sync/RegistrySynchronization.java
```diff
@@ -0,0 +1,100 @@
...
+	 */
+	public static <T> boolean isEntryOptional(SimpleRegistry<T> registry, T entry) {
+		return RegistryFlag.isOptional(SynchronizedRegistry.as(registry).quilt$getEntryFlag(entry));
+	}
+
+	/**
+	 * Prevents entry from being synced
```
LambdAurora:
```suggestion
	 * Prevents the given registry entry from being synced.
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/sync/RegistrySynchronization.java
```diff
@@ -0,0 +1,100 @@
...
+	 */
+	public static <T> void setEntrySkipped(SimpleRegistry<T> registry, T entry) {
+		RegistryFlag.setEntry(registry, entry, RegistryFlag.SKIP);
+	}
+
+	/**
+	 * Prevents entry from being synced
```
LambdAurora:
```suggestion
	 * Prevents the specified registry entry from being synced.
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/sync/RegistrySynchronization.java
```diff
@@ -0,0 +1,100 @@
...
+	 */
+	public static <T> void setEntrySkipped(SimpleRegistry<T> registry, Identifier identifier) {
+		RegistryFlag.setEntry(registry, identifier, RegistryFlag.SKIP);
+	}
+
+	/**
+	 * Checks if registry is skipped from sync
```
LambdAurora:
```suggestion
	 * {@return {@code true} if the given registry entry is skipped in synchronization, or {@code false} otherwise}
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/RegistryConfig.java
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+
+package org.quiltmc.qsl.registry.impl;
+
+import org.quiltmc.config.api.Config;
+import org.quiltmc.config.api.values.TrackedValue;
+import org.quiltmc.loader.api.config.QuiltConfig;
+
+public class RegistryConfig {
```
LambdAurora:
```suggestion
@ApiStatus.Internal
public class RegistryConfig {
```

I also would appreciate more blank lines in this class, everything is so close together right now.

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/ClientPackets.java
```diff
@@ -0,0 +1,53 @@
...
+ * limitations under the License.
+ */
+
+package org.quiltmc.qsl.registry.impl.sync;
+
+import net.minecraft.util.Identifier;
+import org.jetbrains.annotations.ApiStatus;
+
+/**
+ * Identifiers of packets sent by server
```
LambdAurora:
```suggestion
 * Identifiers of packets sent by server.
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/RegistryFlag.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+package org.quiltmc.qsl.registry.impl.sync;
+
+import net.minecraft.util.Identifier;
+import net.minecraft.util.registry.SimpleRegistry;
+import org.jetbrains.annotations.ApiStatus;
+
+/**
+ * Registry flags used to determine how registry (entries) should be handled by sync
```
LambdAurora:
```suggestion
 * Registry flags used to determine how registry (entries) should be handled by sync.
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/ServerRegistrySyncNetworkHandler.java
```diff
@@ -0,0 +1,349 @@
...
+ *
+ * This is special PacketListener for handling registry sync.
+ * Why does it exist? Wouldn't usage of login packets be better?
+ *
+ * And well, yes it would, but sadly these aren't compatible with proxy
+ * software like Velocity (see Forge). Thankfully emulating them on PLAY
+ * protocol isn't too hard and gives equal results. And doing them on PLAY
+ * is needed for Fabric compatibility anyway
+ */
+public final class ServerRegistrySyncNetworkHandler implements ServerPlayPacketListener {
```
LambdAurora, line 47:
```suggestion
@ApiStatus.Internal
public final class ServerRegistrySyncNetworkHandler implements ServerPlayPacketListener {
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/client/ClientRegistrySync.java
```diff
@@ -0,0 +1,260 @@
...
+
+		handler.getConnection().send(ClientPlayNetworking.createC2SPacket(ClientPackets.SYNC_FAILED, buf));
+	}
+
+	private static void handleGoodbyePacket(MinecraftClient client, ClientPlayNetworkHandler handler, PacketByteBuf buf, PacketSender sender) {
+	}
+
+	private static void handleStartPacket(MinecraftClient client, ClientPlayNetworkHandler handler, PacketByteBuf buf, PacketSender sender) {
+		var identifier = buf.readIdentifier();
+		var count = buf.readVarInt();
```
LambdAurora: Please don't use `var` for primitives.

LambdAurora (changes requested):

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/RegistryConfig.java
```diff
@@ -0,0 +1,41 @@
...
+import org.quiltmc.config.api.values.TrackedValue;
+import org.quiltmc.loader.api.config.QuiltConfig;
+
+@ApiStatus.Internal
+public class RegistryConfig {
+	public static final String REGISTRY_SYNC = "registry_sync";
+	public static final String NO_REGISTRY_SYNC_MESSAGE = "no_registry_sync_message";
+	public static final String SUPPORT_FABRIC_API_SYNC = "support_fabric_api_sync";
+	private static Config config = null;
+	public static Config getConfig() {
```
EnnuiL: The way that the config system is implemented is way too cursed; I'd recommend basing it on Ok Zoomer's [Config](https://github.com/EnnuiL/OkZoomer/blob/1.19/src/main/java/io/github/ennuil/ok_zoomer/config/OkZoomerConfig.java)/[ConfigManager](https://github.com/EnnuiL/OkZoomer/blob/1.19/src/main/java/io/github/ennuil/ok_zoomer/config/OkZoomerConfigManager.java) system, or even Haven's [basic Config example](https://github.com/Haven-King/developer-wiki/blob/main/wiki/configs/basic/markdown/basic.md), considering that the only thing being done with the config entries is reading it
For keeping snake_case on these examples (which I do recommend, please don't use camelCase on TOML), there's still the cursedness of using it on few fields, but we can fix it later once naming conventions are here

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/ServerRegistrySyncNetworkHandler.java
```diff
@@ -0,0 +1,351 @@
...
+				this.connection.disconnect(reason);
+			});
+		} catch (Exception var3) {
+			LOGGER.error("Error whilst disconnecting player", var3);
+		}
+	}
+
+
+	@Override
+	public void onHandSwing(HandSwingC2SPacket packet) {
```
EnnuiL: There are *way* too many blank lines there; Change all of those to `{ }`

library/core/registry/src/main/java/org/quiltmc/qsl/registry/mixin/IdListMixin.java
```diff
@@ -0,0 +1,48 @@
...
+import org.quiltmc.qsl.registry.impl.sync.SynchronizedIdList;
+import org.spongepowered.asm.mixin.Final;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+
+import java.util.List;
+
+@Mixin(IdList.class)
+public class IdListMixin<T> implements SynchronizedIdList {
+
```
EnnuiL: Make this go poof...?
```suggestion
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/client/ClientRegistrySync.java
```diff
@@ -0,0 +1,260 @@
...
+		}
+	}
+
+	public static void restoreSnapshot() {
+		for (var reg : Registry.REGISTRIES) {
+			if (reg instanceof SynchronizedRegistry registry && registry.quilt$requiresSyncing()) {
+				registry.quilt$restoreIdSnapshot();
+			}
+		}
+		ClientRegistrySync.rebuildBlockStates();
```
EnnuiL: I noticed that there are bits of static calls that has the class bit in (like this one) and many others without it; I'd recommend going through them all and add them back when it makes sense

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/client/ClientRegistrySync.java
```diff
@@ -0,0 +1,260 @@
...
+			currentCount = count;
+			currentFlags = flags;
+			currentRegistryId = identifier;
+			syncMap = new HashMap<>();
+		} else if (RegistryFlag.isOptional(flags)) {
+			optionalRegistry = true;
+		} else {
+			LOGGER.warn("Trying to sync registry " + identifier + " which doesn't " + (registry != null ? "support it!" : "exist!"));
+			sendSyncFailedPacket(handler, identifier);
+			handler.getConnection().disconnect(getMessage("missing_registry","Client is missing required registry! Mismatched mods?"));
```
EnnuiL:
```suggestion
			handler.getConnection().disconnect(getMessage("missing_registry", "Client is missing required registry! Mismatched mods?"));
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/client/ClientRegistrySync.java
```diff
@@ -0,0 +1,260 @@
...
+	private static boolean optionalRegistry;
+
+	public static void registerHandlers() {
+		ClientPlayNetworking.registerGlobalReceiver(ServerPackets.HANDSHAKE, ClientRegistrySync::handleHelloPacket);
+		ClientPlayNetworking.registerGlobalReceiver(ServerPackets.REGISTRY_START, ClientRegistrySync::handleStartPacket);
+		ClientPlayNetworking.registerGlobalReceiver(ServerPackets.REGISTRY_DATA, ClientRegistrySync::handleDataPacket);
+		ClientPlayNetworking.registerGlobalReceiver(ServerPackets.REGISTRY_RESTORE, ClientRegistrySync::handleApplyPacket);
+		ClientPlayNetworking.registerGlobalReceiver(ServerPackets.END, ClientRegistrySync::handleGoodbyePacket);
+		ClientPlayNetworking.registerGlobalReceiver(ServerPackets.REGISTRY_RESTORE, ClientRegistrySync::handleRestorePacket);
+
```
EnnuiL: Line go poof
```suggestion
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/client/ClientFabricRegistrySync.java
```diff
@@ -0,0 +1,185 @@
...
+				}
+
+				var missingEntries = currentRegistry.quilt$applySyncMap(syncMap);
+
+				if (ClientRegistrySync.checkMissing(handler, registry.getKey().getValue(), missingEntries)) {
+					break;
+				}
+			}
+		}
+
```
EnnuiL, line 179: Yet another excess of lines
```suggestion
```

EnnuiL (changes requested):
Just address these, and all should be fine

EnnuiL (approved):
That looks much better now! I'd say only stylistic nitpicks are left, however, the logic itself seems fine; PR approved

LambdAurora (approved):

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/RegistryConfig.java
```diff
@@ -0,0 +1,36 @@
...
+package org.quiltmc.qsl.registry.impl;
+
+import org.jetbrains.annotations.ApiStatus;
+import org.quiltmc.config.api.WrappedConfig;
+import org.quiltmc.config.api.annotations.Comment;
+import org.quiltmc.loader.api.config.QuiltConfig;
+
+@ApiStatus.Internal
+public class RegistryConfig extends WrappedConfig {
+	public static final RegistryConfig INSTANCE = QuiltConfig.create("quilt", "registry", RegistryConfig.class);
```
EnnuiL: Let's define a standard here already: `qsl` as the containing folder, module name for the relevant config;
This is what Key Binds API was using, and it would be better in the long-run
```suggestion
	public static final RegistryConfig INSTANCE = QuiltConfig.create("qsl", "registry_sync", RegistryConfig.class);
```

EnnuiL (changes requested):
wait, there's still a card on my sleeves

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/RegistryConfig.java
LambdAurora: I would keep the containing folder as quilt.

EnnuiL: Once again, `quilt` is Loader (or perhaps even `quilt_loader`), `qsl` is QSL; We shouldn't repeat the whole "Use `fabric` as Fabric API's mod ID" mistake

LambdAurora: This is configuration though, very different from a mod id.

My issue with this is if someone who doesn't know what QSL is (which could happen once there's dependency downloading yada yada) it would confuse people.

EnnuiL: It's the same fundamental issue: it tries to simplify too much, and in the result, everyone loses;
There was the whole "We'll call QFAPI "QSL" in order to confuse less people" thing, but in the end, everyone was pretty confused when "QSL" turned out to be QFAPI bundling QSL; We shouldn't lie to the user here, let them know about `qsl`

Patbox: Moved it under `quilt/qsl/registry.toml`. Used `registry` just in case of config being extended in future for non-regsync related options.

EnnuiL (approved):
The only thing left is some good ol' testing, other than that, PR approved!

LambdAurora (changes requested):
Ok so, little update for everyone!

1. Registry Syncing has been tested with a known desynced modpack, it works-ish.
2. While testing with a Fabric server and a Quilt client we noticed something missing, the module forgot to take care of some id maps, notably the item model one. Causing quite the visual desync: 
image

Thus the module is currently `s: not working`, meanwhile that going into FCP is blocked until this issue is fixed.
Once fixed testing can resume and hopefully get this into FCP ;)

Leo40Git (approved):
LGTM!

LambdAurora (approved):
Testing has been done with:
 - Quilt servers
 - Fabric servers
 - Quilt clients
 - Fabric clients

and all combination ended up working which means testing is done.

Leo40Git:
is it finally time to merge registry sync into qsl
Or, well, have it enter FCP.

PR #96, opened 2022-04-16
QSL Versioning, initial draft

TheGlitch76:
This is just enough to be able to get out a release for `1.18`, and nothing more.

Modules and libraries use the same version as QSL as a whole, because module versioning is Really Hard to do properly, and we couldn't identify a major benefit for it.

This PR will be merged once both Oro and Aurora approve, and Quilted Fabric API's QL port is merged.

### DIFF
--- a/build-logic/src/main/java/qsl/internal/Versions.java
+++ b/build-logic/src/main/java/qsl/internal/Versions.java
@@ -17,7 +17,7 @@ public final class Versions {
 	/**
 	 * The QSL version
 	 */
-	public static final String QSL_VERSION = "0.1.0";
+	public static final String QSL_VERSION = "1.1.0-beta.1";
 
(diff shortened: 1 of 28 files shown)


OroArmor (approved):

build-logic/src/main/java/qsl/internal/Versions.java
```diff
@@ -17,7 +17,7 @@ public final class Versions {
 	/**
 	 * The QSL version
 	 */
-	public static final String QSL_VERSION = "0.1.0";
+	public static final String QSL_VERSION = "1.1.0-beta.1";
```
LambdAurora, line 20: Why 1.1.0?
Should be 1.0.0

LambdAurora (approved):

PR #97, opened 2022-04-17
Improve the handling of QSL icons

EnnuiL:
This makes mod icons effectively per-library, while Mod Menu's parent now has an icon, which is stored in the mandatory Quilt Base API
I'd test this, but due to a lack of fat jars (which will be addressed by QFAPI), I couldn't be able to do that


LambdAurora (changes requested):
This won't actually work:
1. it would require an explicit dependency on resource loader from every modules
2. ModMenu **does not use resource loader**, it loads from the mod resources directly! Mod icons cannot be shared between mods!

EnnuiL:
I really should have tested it before PRing

PR #100, opened 2022-04-19
Block Extension QSL

Azagwen:
This PR contains the base for a QuiltBlock Utility, which currently only has Property proxying (i.e creating blocks that "extend" multiple blocks, thus inheriting their blockstate properties)

To create a block proxy, you simply call `QuiltBlock.createProxy()` in its constructor, as `TestBlock` for the example mod shows.
`QuiltBlock.createProxy()` takes the block's settings, and an array of **already registered blocks**, the block settings are only used to be able to call the method itself, as this needs to happen at the `super()` call
```java
public class TestBlock extends Block {

    public TestBlock(Settings settings, Block... proxies) {
        super(QuiltBlock.createProxy(settings, proxies));
    }

}
```


library/block/block_extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/QuiltBlock.java
```diff
@@ -0,0 +1,77 @@
...
+
+public class QuiltBlock {
+
+    /**
+     * Creates a proxy from one or multiple blocks (will be temporarily stocked in a {@link ThreadLocal} and used in {@link org.quiltmc.qsl.block.extensions.mixin.MixinBlock}).
+     * @param settings  Block settings of the block you want to add Proxies for (dummy usage, to be able to insert logic at {@code  super()} call.
+     * @param proxies   Array of proxy blocks to be "merged" together as one (property merging).
+     * @return          Same Block settings as above.
+     */
+    public static AbstractBlock.Settings createProxy(AbstractBlock.Settings settings, Block... proxies) {
```
LambdAurora:
```suggestion
    public static <S extends AbstractBlock.Settings> S createProxy(S settings, Block... proxies) {
```

library/block/block_extensions/src/main/java/org/quiltmc/qsl/block/extensions/mixin/MixinBlock.java
```diff
@@ -0,0 +1,88 @@
...
+    @Shadow protected abstract Block asBlock();
+    @Shadow @Final private static Logger LOGGER;
+    private final @Unique LinkedList<Block> proxies = new LinkedList<>();
+
+    @Inject(method = "<init>",
+            at = @At(value = "INVOKE", target = "Lnet/minecraft/state/StateManager$Builder;build(Ljava/util/function/Function;Lnet/minecraft/state/StateManager$Factory;)Lnet/minecraft/state/StateManager;"),
+            locals = LocalCapture.CAPTURE_FAILHARD)
+    private void injectProxyProperties(AbstractBlock.Settings settings, CallbackInfo ci, StateManager.Builder<Block, BlockState> builder) {
+        if (QuiltBlockImpl.PROXY_BLOCKS_TEMP_CONTAINER.get() != null && QuiltBlockImpl.PROXY_BLOCKS_TEMP_CONTAINER.get().length > 0){
+            var properties = Lists.newArrayList();
```
LambdAurora:
```suggestion
            var properties = new ArrayList<Property>();
```

Please don't use stuff reminiscent from the before diamond generic Java versions.

library/block/block_extensions/src/testmod/java/org/quiltmc/qsl/block/extensions/test/TestBlock.java
```diff
@@ -0,0 +1,31 @@
...
+public class TestBlock extends Block {
+
+    public TestBlock(Settings settings, Block... proxies) {
+        super(QuiltBlock.createProxy(settings, proxies));
+        this.setDefaultState(QuiltBlock.getProxyDefaultState(this));
+    }
+
+    @Override
+    public ActionResult onUse(BlockState state, World world, BlockPos pos, PlayerEntity player, Hand hand, BlockHitResult hit) {
+        var prop = QuiltBlock.getProperty(state, StairsBlock.HALF);
```
LambdAurora, line 40: This usage of getProperty doesn't sound useful?

If the stuff that was merged were merged properly, it should already include the property.

LambdAurora (commented):
Some notes: 
 - we use `<Class>Mixin` and not `Mixin<Class>`
 - I feel like copyProperty should do checks to avoid copying something that cannot be copied
  - getProxyDefaultState probably should be replaced with a more generic method called `mergeStates(BlockState source, BlockState target)`.
  
Also don't forget to run the `applyLicenses` task!

Azagwen:
> Some notes:
> 
> * we use `<Class>Mixin` and not `Mixin<Class>`
> * I feel like copyProperty should do checks to avoid copying something that cannot be copied
> * getProxyDefaultState probably should be replaced with a more generic method called `mergeStates(BlockState source, BlockState target)`.
> 
> Also don't forget to run the `applyLicenses` task!

How would you generify `getProxyDefaultState`? 

I struggle to see any ways to 100% generify this method in particular, as it's currently not possible to target a single Proxy and it needs to get all of them in a loop to gather teh default states. 

Unless I create a new method to `mergeStates` and use it in a refactor of `getProxyDefaultState` which does loop through proxies?

PR #102, opened 2022-04-20
Javadoc Improvements

### LINKED ISSUE #101: Fat Javadoc Jar
A javadoc jar that has javadocs for all the modules in QSL. Will increase API discovery and easier to link to for basic docs

OroArmor:
Fixes #101


build-logic/src/main/groovy/qsl.module.gradle
```diff
@@ -160,11 +160,10 @@ javadoc {
 				"https://asm.ow2.io/javadoc/",
 				"https://docs.oracle.com/en/java/javase/16/docs/api/",
 				"https://jenkins.liteloader.com/job/Mixin/javadoc/",
-				"https://logging.apache.org/log4j/2.x/log4j-api/apidocs/"
-				// TODO: Javadoc for mappings
+				"https://logging.apache.org/log4j/2.x/log4j-api/apidocs/",
+				"https://maven.quiltmc.org/repository/release/org/quiltmc/quilt-mappings/${Versions.MINECRAFT_VERSION}+build.${Versions.MAPPINGS_BUILD}/quilt-mappings-${Versions.MINECRAFT_VERSION}+build.${Versions.MAPPINGS_BUILD}-javadoc.jar/"
```
LambdAurora, line 182: Does this really work?

Javadoc task is executed before publishing.

OroArmor: Yes. These are the Quilt Mappings javadocs, not QSL

LambdAurora: Misread oops, nice then.

LambdAurora (approved):

build.gradle
```diff
@@ -39,9 +39,58 @@ jar {
 	enabled = false
 }
 
+task fatJavadoc(type: Javadoc) {
```
TheGlitch76, line 42: Why make this separate from normal javadoc?

OroArmor: i was having some issues with it being the normal javadoc, probably could move it back

TheGlitch76: Let me know if it doesn't work.

OroArmor: After testing, all that I can change are the names to match the standard. However, I feel like the names should stay the same as they are now, since it is not a normal javadoc generation.

TheGlitch76 (approved):

PR #104, opened 2022-04-22
Block access widening.

LambdAurora:
Access widening a bunch of block-related stuff transitively.

Based on fabric-transitive-access-wideners.


OroArmor (approved):
Works. However, I would leave the current file as is, and have the auto generated content be added to a second file that is gitignore d.

LambdAurora:
> Works. However, I would leave the current file as is, and have the auto generated content be added to a second file that is gitignore d.

Pretty sure we can't have multiple access widener files per mod.

OroArmor:
QMJ spec says otherwise

TheGlitch76:
Loom does not currently support multiple AWs.

library/block/block_extensions/src/main/resources/quilt_block_extensions.accesswidener
```diff
@@ -1,3 +1,104 @@
-accessWidener	v1	named
+accessWidener v2 named
+
 extendable  method    net/minecraft/block/AbstractBlock$Settings    <init>  (Lnet/minecraft/block/Material;Ljava/util/function/Function;)V
 extendable  method    net/minecraft/block/AbstractBlock$Settings    <init>  (Lnet/minecraft/block/Material;Lnet/minecraft/block/MapColor;)V
+
+### Generated access wideners below
```
TheGlitch76, line 6: We intentionally avoid committing generated QMJs, but i suppose this is fine.

library/block/block_extensions/template.accesswidener
```diff
@@ -0,0 +1,6 @@
+accessWidener v2 named
```
TheGlitch76, line 1: could we have a README.txt or something in this module explaining how to add things to the generated access widener?

TheGlitch76 (commented):
Everything Access Widener in QSL
Everything Access Widener in QSL
Everything Access Widener in QSL

library/block/block_extensions/src/main/resources/quilt_block_extensions.accesswidener
LambdAurora: I tried to move the entire accesswidener file to the generated resources source set like for the QMJ
but loom was failing to read it sadly.

So I gave up on that one.

library/block/block_extensions/template.accesswidener
LambdAurora: Added!

TheGlitch76 (approved):

LemmaEOF:
Would it be possible to generate wideners for only `init` methods that *aren't* public?

LambdAurora:
> Would it be possible to generate wideners for only `init` methods that _aren't_ public?

@LemmaEOF that's already what it does:

```java

					// Block constructor...
					if (method.name == "<init>" && Type.getArgumentTypes(method.desc).any { it.internalName == 'net/minecraft/block/AbstractBlock$Settings' }) {
						// ...and non-public
						if ((method.access & Opcodes.ACC_PUBLIC) == 0) {
							lines.add("transitive-accessible method $node.name <init> $method.desc")
						}
					}
```

LemmaEOF:
Oh, huh. I didn't realize that many block ctors were non-public.

Leo40Git (approved):

PR #105, opened 2022-04-22
Modify storing of status effect to make it more mod friendly

Patbox:
Minecraft by default serializes status effects as raw registry value limited to a byte! Which isn't great mod compatibility wise (raw ids shouldn't be considered stable) and limits to supporting only 256 status effects globally!

I'm not sure if it should be part of registry package through, but it is simple to move


Kroppeb (approved):

OroArmor (approved):

library/core/registry/src/main/java/org/quiltmc/qsl/registry/mixin/patch/StatusEffectInstanceMixin.java
```diff
@@ -0,0 +1,64 @@
...
+				return Registry.STATUS_EFFECT.get(identifier);
+			}
+		}
+
+		return effect;
+	}
+
+	@Inject(method = "writeNbt", at = @At("TAIL"))
+	private void quilt$storeIdentifier(NbtCompound nbt, CallbackInfoReturnable<NbtCompound> cir) {
+		nbt.putString("quilt:id", Registry.STATUS_EFFECT.getId(this.type).toString());
```
TheGlitch76: These NBT constants should be fields in an impl class somewhere.

TheGlitch76 (changes requested):
A couple questions for you:
- How does this PR handle byte overflow when there are more than 256 status effects? I don't see any code to disable or otherwise affect the Vanilla method of (de)serializing effects.
- How does this PR affect situations where a modded world is moved back to vanilla? What happens?

Patbox:
> A couple questions for you:
> 
> * How does this PR handle byte overflow when there are more than 256 status effects? I don't see any code to disable or otherwise affect the Vanilla method of (de)serializing effects.
> * How does this PR affect situations where a modded world is moved back to vanilla? What happens?

For simplicity it doesn't change how vanilla stores this, as id ones have priority. MC already has null checks so it should be safe to move world back to vanilla. It will just ignore additional nbt and use byte value

TheGlitch76:
> For simplicity it doesn't change how vanilla stores this, as id ones have priority. MC already has null checks so it should be safe to move world back to vanilla. It will just ignore additional nbt and use byte value

I don't see any code here to guard for cases where the byte value is an outright lie (like integer overflow)?

OroArmor (approved):
Just Glitch's suggestion for constants being better documented.

LambdAurora (approved):
This is necessary to ensure that status effects don't get broken on world save and provides the missing replacement for registry sync.

Now to address glitch's comments

> A couple questions for you:
>
>    How does this PR handle byte overflow when there are more than 256 status effects? I don't see any code to disable or otherwise affect the Vanilla method of (de)serializing effects.

All this code takes precedence over Vanilla, meaning if QSL is installed, QSL's way of deserialization will have priority making status effect serialization/deserialization much more reliable.

>    How does this PR affect situations where a modded world is moved back to vanilla? What happens?

If the Vanilla status effects IDs have not been jumbled, it should work just fine since the original Vanilla values are still stored as neighbors.

Leo40Git (approved):
The Flattening 2: Mojank's Leftovers
LGTM!

library/core/registry/src/main/java/org/quiltmc/qsl/registry/mixin/patch/BeaconBlockEntityMixin.java
```diff
@@ -0,0 +1,87 @@
...
+@Mixin(BeaconBlockEntity.class)
+public class BeaconBlockEntityMixin {
+
+	@Shadow
+	@Final
+	private static Set<StatusEffect> EFFECTS;
+	@Shadow
+	@Nullable StatusEffect primary;
+	@Shadow
+	@Nullable StatusEffect secondary;
```
EnnuiL:
```suggestion
	@Shadow
	@Final
	private static Set<StatusEffect> EFFECTS;

	@Shadow
	@Nullable StatusEffect primary;

	@Shadow
	@Nullable StatusEffect secondary;
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/mixin/patch/SuspiciousStewItemMixin.java
```diff
@@ -0,0 +1,90 @@
...
+				this.quilt$effect.set(Registry.STATUS_EFFECT.get(identifier));
+			} else {
+				this.quilt$effect.remove();
+			}
+		} else {
+			this.quilt$effect.remove();
+		}
+	}
+
+	@SuppressWarnings({"InvalidInjectorMethodSignature"})
```
EnnuiL: oh wow, it's quite a long name for a warning
although maybe the {}s aren't needed?
```suggestion
	@SuppressWarnings("InvalidInjectorMethodSignature")
```

EnnuiL (approved):
Nitpick nitpicked, just take a look at 'em first;
Anyway, if nothing breaks on migration from raw IDs to Identifiers, then all is fine!

LambdAurora:
> Anyway, if nothing breaks on migration from raw IDs to Identifiers, then all is fine!

Nothing would since if no quilt Identifier is present it automatically use the raw IDs.

PR #113, opened 2022-05-22
Fix crash involving Resource Loader's Overwrites

EnnuiL:
Those `@Overwrite`s were meant to target methods from ResourcePack, however, there were no points signifying that, meaning that it ends crashing on production. This fixes that

This fix should be merged as soon as possible, since it's currently blocking a Quilted FAPI update


LambdAurora (approved):

LambdAurora:
This is weird and leads me to think something else broke it since the same mixin worked before, oh well, if this fixes it.

PR #114, opened 2022-05-23
Create a fat, remapped testmod jar

TheGlitch76:
Remapped testmod jars that you can throw into your favorite launcher are created on the module, library, and root levels. They are not published to maven, but can be downloaded through Github Actions


OroArmor (approved):
The javadoc stuff should be merged at the same time as this

LambdAurora (approved):

PR #117, opened 2022-05-26
"Fix" and polish the Quilt Screen API

EnnuiL:
This PR attempts to clean up and do a tiny refactor on where it made sense on the Screen API (can't have breaking changes if the last few QSL releases are broken anyway), and also finally fixes a crash involving its injected interface; Unfortunately, the "fix" was to temporarily disable the injection until the refmap issues are no more

### DIFF
--- a/library/gui/screen/src/main/java/org/quiltmc/qsl/screen/api/client/QuiltScreen.java
+++ b/library/gui/screen/src/main/java/org/quiltmc/qsl/screen/api/client/QuiltScreen.java
@@ -46,22 +46,16 @@ public interface QuiltScreen {
 	List<ClickableWidget> getButtons();
 
 	/**
-	 * Gets the screen's internal item renderer.
-	 *
 	 * {@return the screen's item renderer}
 	 */
 	ItemRenderer getItemRenderer();
 
 	/**
-	 * Gets the screen's internal text renderer.
-	 *
 	 * {@return the screen's text renderer}
 	 */
 	TextRenderer getTextRenderer();
 
 	/**
-	 * Gets the screen's internal Minecraft client instance.
-	 *
 	 * {@return the Minecraft client instance}
 	 */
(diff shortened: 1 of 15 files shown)


maximumpower55 (approved):
LGTM

PepperCode1 (approved):

library/gui/screen/src/main/java/org/quiltmc/qsl/screen/api/client/QuiltScreen.java
```diff
@@ -28,37 +28,40 @@
...
-@InjectedInterface(Screen.class)
-public interface QuiltScreenExtensions {
+public interface QuiltScreen {
 	/**
-	 * Gets all of a screen's button widgets.
-	 * <p>
-	 * The provided list allows for addition and removal of buttons from the screen.
+	 * Gets all of the screen's button widgets.
+	 *
+	 * <p>The provided list allows for addition and removal of buttons from the screen.
```
LambdAurora: Please don't move those `<p>` tags to a start of a paragraph in an inline way.

LambdAurora (changes requested):

library/gui/screen/src/main/java/org/quiltmc/qsl/screen/api/client/QuiltScreen.java
EnnuiL: The Checkstyle config should be changed in order to reflect that then

EnnuiL: I've changed it back so it followed CONTRIBUTING.md; This resulted into some of the FAPI copy-paste being quiltified in the process

LambdAurora (approved):

TheGlitch76 (approved):

PR #120, opened 2022-05-30
Port Dimensions fast path + bug fix from upstream

TheGlitch76:
Just some ports from upstream Fabric.

The bugfix mixin isn't strictly necessary, because most people will have FAPI anyway, but I added it in case that two-year-old bug never gets fixed

### DIFF
--- a/library/worldgen/dimension/src/main/java/org/quiltmc/qsl/worldgen/dimension/impl/QuiltDimensionsImpl.java
+++ b/library/worldgen/dimension/src/main/java/org/quiltmc/qsl/worldgen/dimension/impl/QuiltDimensionsImpl.java
@@ -18,6 +18,7 @@
 package org.quiltmc.qsl.worldgen.dimension.impl;
 
 import com.google.common.base.Preconditions;
+import net.minecraft.server.network.ServerPlayerEntity;
 import org.jetbrains.annotations.ApiStatus;
 
 import net.minecraft.entity.Entity;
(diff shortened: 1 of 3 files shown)


library/worldgen/dimension/src/main/java/org/quiltmc/qsl/worldgen/dimension/impl/QuiltDimensionsImpl.java
```diff
@@ -18,6 +18,7 @@
 package org.quiltmc.qsl.worldgen.dimension.impl;
 
 import com.google.common.base.Preconditions;
+import net.minecraft.server.network.ServerPlayerEntity;
```
triphora: This is going to fail checkstyle

TheGlitch76: oops. will fix in a bit

peaceheis (commented):
everything looks good to me!if you can just fix that one style issue, i'll approve.

EnnuiL:
This PR will be necessary in order to quiltify the Fabric Dimensions API on QFAPI, so yeah, we'll actually need it

LambdAurora (approved):

EnnuiL (approved):

peaceheis (approved):

PR #136, opened 2022-06-16
1.19.1 Tracking PR

NoComment1105:
Tracking PR for 1.19.1 updates


TheGlitch76:
If we're going on the draft versioning scheme I wrote forever ago, then we won't update to 1.19.1 unless there's a specific need to, and that will become QSL 3.0.0

LambdAurora:
I have added whitelisting for the 22w24a version in the QMJs, now we can close this (for now I guess) ;)

PR #138, opened 2022-06-20
Add mod resource packs caching for faster reloads.

LambdAurora:
Also use the java.io.File API when possible instead of NIO exists.

#### First Reload

| Method | Time |
|:---------:|-------:| 
| No Cache, no File API | 8372 ms |
| No Cache, File API | 5664 ms |
| Cache, no File API | 8252 ms |
| Cache, File API | 8195 ms |
| Tree Cache, no File API | 6263 ms |
| Tree Cache, File API | 6178 ms |

#### World Reload

| Method | Time |
|:---------:|-------:| 
| No Cache, no File API | 1409 ms |
| No Cache, File API | 1260 ms |
| Cache, no File API | 1304 ms |
| Cache, File API | 1487 ms |
| Tree Cache, no File API | 1275 ms |
| Tree Cache, File API | 1187 ms |

#### First Reload

| Method | Time |
|:---------:|-------:| 
| No Cache | 27025 ms |
| Cache | 21641 ms |
| Tree Cache | 20378 ms |

#### World Reload

| Method | Time |
|:---------:|-------:| 
| No Cache | 24566ms |
| Cache | 20721ms - 21781ms |
| Tree Cache | 19679ms - 19917ms |

In the blanketcon modpack this manages to remove almost 7 seconds of startup time!

Cache corresponds to a much simpler cache implementation with a simple ConcurrentHashMap, Tree Cache is the cache implemented in this PR.

For the dev env, only the Minecraft resource pack itself is cached by this system since the QSL dev env only has this "mod" that is eligible for caching.
Mods in dev envs that may have mutable resources via debugging will not use the cache.

### DIFF
--- a/library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/cache/CacheTree.java
+++ b/library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/cache/CacheTree.java
@@ -1,0 +1,179 @@
+/*
+ * Copyright 2022 QuiltMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.quiltmc.qsl.resource.loader.impl.cache;
+
+import java.util.ArrayList;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.Nullable;
+
+import org.quiltmc.qsl.resource.loader.impl.ModIoOps;
+
+/**
+ * Contains the definition of a tree structure for caching, each node represents either a directory, a file, or a missing entry.
+ * <p>
+ * This may be dangerous if someone does a lot of one-time access.
+ *
+ * @author LambdAurora
+ */
+@ApiStatus.Internal
+public final class CacheTree {
+	/**
+	 * {@return a new tree, represented as a root branch node}
+	 */
+	static Branch newTree() {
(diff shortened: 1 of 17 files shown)


Leo40Git (approved):

OroArmor (approved):

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/cache/CacheTree.java
```diff
@@ -0,0 +1,179 @@
...
+			if (node == null) {
+				String absolutePath = childName;
+
+				if (!this.isRoot()) {
+					absolutePath = this.getFullPath() + '/' + absolutePath;
+				}
+
+				var type = io.getEntryType(absolutePath);
+
+				switch (type) {
```
EnnuiL, line 136: Hmm, I wonder if using directly using the switch as the returned value and replacing its internal returns with yields would be better or not, but yeah, I see a potential for that

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/cache/ResourceAccess.java
```diff
@@ -0,0 +1,139 @@
...
+	/**
+	 * Gets the resource entry type of the given path.
+	 *
+	 * @param pathName the path of the resource
+	 * @return the resource entry type
+	 */
+	public EntryType getEntryType(String pathName) {
+		var entry = this.getEntry(pathName);
+
+		if (entry == null) {
```
EnnuiL: Perhaps this could be a `entry != null ? entry.type() : EntryType.EMPTY`?

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/cache/ResourceTreeCache.java
```diff
@@ -0,0 +1,67 @@
...
+	private final Map<ResourceType, Set<String>> namespaces = new EnumMap<>(ResourceType.class);
+
+	public ResourceTreeCache(ModIoOps io) {
+		super(io);
+	}
+
+	@Override
+	public @Nullable Entry getEntry(String pathName) {
+		var node = this.tree.resolveOrCompute(this.io, pathName);
+		if (node != null) {
```
EnnuiL: Same here

EnnuiL (approved):
All that I have are style nitpicks that *might* not be better than their current state, so, yeah, you can ignore those

PR #146, opened 2022-06-29
Component API Implementation

0xJoeMama:
After 2 weeks of unhealthy head-banging I have finally finished this!

This API adds:
* A new module to the data library, called 'component'.
* A new class on the qsl_base module, called Maybe. It represents a monad that either holds a value or doesn't.

# A bit of an explanation: 
This module allows you attach data, functions, ticking functions to game objects.
Right now the following objects are implemented:
* Block Entities
* Entities
* Chunks
* ~~ItemStacks~~(Removed from the Initial Release)
* LevelProperties

The data can be serialized using NBT and can also be network synced(although that feature is a bit unstable at the moment).

## Component Injection
During the mod loading stage, mods are able to inject into difference game objects, provided those objects implement the ComponentProvider interface. Those injections are cached using the target class/classes and are then able to be quickly looked up.  

*But what about the case where cached injections don't suffice?* one may ask.
Well that's also covered using **dynamic injection**.  
Along with the class that you are targeting, you can also provide a predicate that will match the type of the class.
That predicate is then used when a provider is initialized to make sure it matches and/or can be injected into the target.
Proceed with caution: **The predicate is checked, at the end of the constructor of the highest class in the hierarchy. This means that only the properties initialized up to that point will be available for querying!.**

While that may not seem like a problem for things like ItemStacks, which generally aren't extended, Entities for example have a wide variety of classes with different fields. However, when injecting dynamically, one will only have access to the Entity class fields as well as the type information of the object.

## Custom implementations
I have made the API really flexible, so that anyone can just create custom implementations of most provided objects.
Component containers, component providers, components, sync headers, all of those can be extended and tweaked as per need.

## Network sync and data saving
Data is currently saved using NBT and a save call that is provided from the provider to its container.
Data is currently synced using a system I have implemented. Needless to say, this system needs some tweaking as well as a lot of testing to make sure that: 
1. No data is lost through the network.
2. ~~There are no runtime crashes caused by the system.~~
3. ~~Components are properly synced even in edge cases.~~
4. ~~Components are synced when the object is first loaded.~~

I think that after all that is done, this will be able to be released.

# Finally
Feel free to ask any questions about the implementation and provide feedback wherever possible.
This project was a massive undertaking for one person and that is why I decided to PR it now that it's reaching its final stages.
The codebase is big and documentation is needed, I understand that. However I still think I needed to share it with everyone.
Thank you for your time!


library/data/component/src/main/java/org/quiltmc/qsl/component/api/wrapper/Wrapper.java
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.quiltmc.qsl.component.api.wrapper;
+
+public class Wrapper<T> {
```
Leo40Git: What's the point of this API class? It only seems to be used in your testmod, and there it can easily be replaced by doing this:
```java
final ActionResult[] res = new ActionResult[] { ActionResult.PASS };

OptionalProducingMethod.invoke().ifPresent(thing -> {
    if (thing == ONE)
        res[0] = ActionResult.CONSUME;
})

return res[0];
```

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/ComponentInjectionCache.java
```diff
@@ -0,0 +1,60 @@
...
+package org.quiltmc.qsl.component.impl;
+
+import net.minecraft.util.Util;
+import org.quiltmc.qsl.component.api.ComponentType;
+
+import java.util.*;
+
+public class ComponentInjectionCache {
+
+	private static ComponentInjectionCache INSTANCE;
```
Leo40Git: Unnecessary singleton, just make the fields and methods static.

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/ComponentsImpl.java
```diff
@@ -0,0 +1,110 @@
...
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+@ApiStatus.Internal
+public class ComponentsImpl {
+	private static final Logger LOGGER = LoggerFactory.getLogger("QSL/Components");
+	public static final String MODID = "quilt_component";
+    private static final Map<ComponentInjectionPredicate, Set<Identifier>> INJECTION_REGISTRY = new HashMap<>();
+	private static final RegistryKey<Registry<ComponentType<?>>> REGISTRY_KEY =
```
Leo40Git: This registry key should be exposed.

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/ComponentsImpl.java
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	public static ComponentType<?> getEntry(Identifier id) {
+		return REGISTRY.getOrEmpty(id).orElseThrow(
+				ErrorUtil.illegalArgument("Cannot access element with id %s in the component registry!".formatted(id))
+		);
+	}
+
+	@NotNull
+	public static Logger getLogger() {
```
Leo40Git: ...why not just make the logger `public`?

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/container/LazifiedComponentContainer.java
```diff
@@ -0,0 +1,239 @@
...
+import org.quiltmc.qsl.component.impl.sync.packet.PacketIds;
+import org.quiltmc.qsl.component.impl.util.ErrorUtil;
+import org.quiltmc.qsl.component.impl.util.Lazy;
+import org.quiltmc.qsl.component.impl.util.StringConstants;
+import org.quiltmc.qsl.networking.api.ServerPlayNetworking;
+
+import java.util.*;
+import java.util.function.Supplier;
+
+public class LazifiedComponentContainer implements ComponentContainer {
```
Leo40Git: Possibly rename this to `LazyComponentContainer`? "Lazified" makes no sense to me.

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/predicates/FilteredInheritedInjectionPredicate.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+
+public class FilteredInheritedInjectionPredicate extends InheritedInjectionPredicate {
+	private final List<Class<?>> exceptions;
+
+	public FilteredInheritedInjectionPredicate(Class<?> clazz, Class<?>[] exceptions) {
+		super(clazz);
+		this.exceptions = new ArrayList<>(Arrays.asList(exceptions));
```
Leo40Git:
```suggestion
		this.exceptions = Arrays.asList(exceptions);
```
`Arrays.asList` returns an immutable list. I'm not sure why you'd want this predicate to be mutable, especially since all the other predicates seem to be immutable.

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/predicates/RedirectedInjectionPredicate.java
```diff
@@ -0,0 +1,35 @@
...
+import org.quiltmc.qsl.component.api.ComponentProvider;
+
+import java.util.Set;
+
+public class RedirectedInjectionPredicate extends ClassInjectionPredicate {
+	private final Set<Class<?>> redirections;
+
+	public RedirectedInjectionPredicate(Class<?> clazz, Set<Class<?>> redirections) {
+		super(clazz);
+		this.redirections = redirections;
```
Leo40Git:
```suggestion
		this.redirections = Set.copyOf(redirections);
```
Immutability is fun (no it isn't)!

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/util/Lazy.java
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	@NotNull
+	public static <T> Lazy<T> filled(@NotNull T value) {
+		return new Lazy<>(value);
+	}
+
+	@NotNull
+	public <U> Lazy<U> compose(Function<T, U> transformer) {
+		return Lazy.of(() -> transformer.apply(this.sup.get()));
```
Leo40Git: Shouldn't this be this?
```suggestion
		return Lazy.of(() -> transformer.apply(this.get()));
```
Unless bypassing the lazy behavior was intentional.

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/util/Lazy.java
```diff
@@ -0,0 +1,109 @@
...
+	public boolean isEmpty() {
+		return this.value == null;
+	}
+
+	public void compute(@NotNull Consumer<T> action) {
+		action.accept(this.get());
+	}
+
+	@Override
+	public int hashCode() {
```
Leo40Git: ...why do these methods exist? I highly doubt someone's gonna use `Lazy`s as map keys.

library/data/component/src/testmod/java/org/quiltmc/qsl/component/test/TestBlock.java
```diff
@@ -0,0 +1,96 @@
...
+		if (world.isClient || be == null) {
+			return super.onUse(state, world, pos, player, hand, hit);
+		}
+
+		ItemStack handStack = player.getStackInHand(hand);
+
+		if (!handStack.isEmpty()) {
+			final Wrapper<ActionResult> ret = new Wrapper<>(ActionResult.PASS);
+
+			be.expose(ComponentTestMod.CHUNK_INVENTORY).ifPresent(inventoryComponent -> {
```
Leo40Git: Replace this with `final ActionResult[] ret = new ActionResult[] { ActionResult.PASS }`, or use `Optional.map`.

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/ComponentsImpl.java
```diff
@@ -0,0 +1,110 @@
...
+
+@ApiStatus.Internal
+public class ComponentsImpl {
+	private static final Logger LOGGER = LoggerFactory.getLogger("QSL/Components");
+	public static final String MODID = "quilt_component";
+    private static final Map<ComponentInjectionPredicate, Set<Identifier>> INJECTION_REGISTRY = new HashMap<>();
+	private static final RegistryKey<Registry<ComponentType<?>>> REGISTRY_KEY =
+			RegistryKey.ofRegistry(new Identifier("quilt", "components"));
+	public static final Registry<ComponentType<?>> REGISTRY =
+			new SimpleRegistry<>(REGISTRY_KEY, Lifecycle.experimental(), null);
```
Leo40Git: Give these declarations some breathing space!
```suggestion
	private static final Logger LOGGER = LoggerFactory.getLogger("QSL/Components");
	
	public static final String MODID = "quilt_component";
	
    private static final Map<ComponentInjectionPredicate, Set<Identifier>> INJECTION_REGISTRY = new HashMap<>();
    
	private static final RegistryKey<Registry<ComponentType<?>>> REGISTRY_KEY =
			RegistryKey.ofRegistry(new Identifier("quilt", "components"));
	public static final Registry<ComponentType<?>> REGISTRY =
			new SimpleRegistry<>(REGISTRY_KEY, Lifecycle.experimental(), null);
```

Leo40Git (commented):
...your entire network stack confuses me greatly.

General nitpick: newlines! A lot of these files have a newline after the class declaration for some reason.

Anyway, here's my specific nitpicks.

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/ComponentsImpl.java
0xJoeMama: Okay for this I blame the new intellij version. The new usage fields make it seems like there is space where there actually isn't.

library/data/component/src/main/java/org/quiltmc/qsl/component/api/wrapper/Wrapper.java
0xJoeMama: Agreed. I had actually forgotten that class even existed to be honest.
The array thing is a bit interesting seeing how I never thought of it. Very nice, very nice!

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/ComponentsImpl.java
Leo40Git: ...I legitimately forgot those existed, I disabled those like day one :P (right-click on them)

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/ComponentsImpl.java
0xJoeMama: Just a thing I have been doing for all my mods. It looks better in code to me. If you insist though, I have no issue just making the LOGGER field public.

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/util/Lazy.java
0xJoeMama: This slipped through the cracks. I don't think that method is even used to be honest. 

*If I am to be truly honest this class was literally copied from this gist of mine: https://gist.github.com/0xJoeMama/314607455836e6bd66e74a3715f23154*

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/util/Lazy.java
0xJoeMama: Same as above this was just copied over from my gist without much thought.

0xJoeMama:
> newlines!  

Is there an issue with leaving a newline after the class declaration?

Leo40Git:
Consistency, mostly - none of the current QSL classes have a newline after the class declaration (and heck, even some of the classes in this PR don't have a newline).

0xJoeMama:
Okay will do then. Time to mess with IJ settings!

lenrik1589:
I think there is a way to include code style settings in the repo? well, at least for IDEA users

Leo40Git:
We do have an [`.editorconfig`](https://github.com/QuiltMC/quilt-standard-libraries/blob/1.19/.editorconfig) file, though it doesn't contain any IJ-specific settings.

lenrik1589:
I see

quat1024 (commented):
I'm concerned about firing the event for attaching ItemStack components from the ItemStack constructor.

Minecraft Forge does something similar with its capabilities, and it leads to `new ItemStack` and things that call it such as `copy` and `split` being giant performance hazards, as suddenly this fires off an event.

0xJoeMama:
> I'm concerned about firing the event for attaching ItemStack components from the ItemStack constructor.
> 
> Minecraft Forge does something similar with its capabilities, and it leads to `new ItemStack` and things that call it such as `copy` and `split` being giant performance hazards, as suddenly this fires off an event.

How would you support dynamic injections in this case though? Technically speaking, you could do a cached injection and only expose the component if the required criteria is met.

0xJoeMama:
We could redirect the calls in copy and split though. Hm interesting...

LemmaEOF:
Quilt component API 👀 👀 👀 

Right now, there's not really anything to get components off of non-BE blocks/blockstates. This is, of course, a bit tough because blockstates aren't quite instanced in the same way BEs or item stacks are, but I feel like it is important to provide, given that stuff like composters or redstone in vanilla set a convention that you should be able to make something simple without having to use a full block entity. Using a `BlockComponentProvider` interface with access to a full world/pos/state would work, but that makes injection either impossible or *very* unperformant. I might look into making a lookup-like system for this at some point, unless someone else has a better idea.

0xJoeMama:
@LemmaEOF That is an interesting idea. However, I'd love to hear how you are planning storing the components. Or are they gonna be static components? (In case you haven't seen the code, static components are basically component singletons)

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/container/LazifiedComponentContainer.java
0xJoeMama: This was also a bad use of ArrayList. I should've used a Set since the only method called on this collection is 'contains'.

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/ComponentsImpl.java
```diff
@@ -38,60 +38,65 @@
 
 @ApiStatus.Internal
 public class ComponentsImpl {
-	private static final Logger LOGGER = LoggerFactory.getLogger("QSL/Components");
-	public static final String MODID = "quilt_component";
-    private static final Map<ComponentInjectionPredicate, Set<Identifier>> INJECTION_REGISTRY = new HashMap<>();
-	private static final RegistryKey<Registry<ComponentType<?>>> REGISTRY_KEY =
-			RegistryKey.ofRegistry(new Identifier("quilt", "components"));
+	public static final RegistryKey<Registry<ComponentType<?>>> REGISTRY_KEY =
```
Leo40Git, line 43: When I said "exposed", I meant in the API (in `Components`, preferably)! Sorry for the confusion.

0xJoeMama: Oh, very well then!

0xJoeMama: image
However I wish to let you know that, *this^^* is a thing

Leo40Git: Vanilla has separate key fields for all its registries, so whatever

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/ComponentsImpl.java
LambdAurora: I would just make the logger field public.

The static method just looks like additional noise to me.

library/data/component/src/main/java/org/quiltmc/qsl/component/api/ComponentProvider.java
```diff
@@ -0,0 +1,54 @@
...
+		Entity.class,
+		BlockEntity.class,
+		Chunk.class,
+		LevelProperties.class,
+		ItemStack.class,
+		MinecraftClient.class
+})
+public interface ComponentProvider {
+	@NotNull
+	ComponentContainer getContainer();
```
LambdAurora: Given that this is an injected interface, we need to make sure that those names don't have too many risks of conflict, so first of all
```suggestion
	ComponentContainer getComponentContainer();
```

call and expose might be fine since they take QSL-specific arguments.

And this method might require a default implementation, as BasiqueEvangelist pointed out multiple times when we dealt with those interfaces, if there's no default implementation it will disrupt development environments and javac when trying to compile a basic Entity, since the issue is those implementations are hidden away via mixin.

library/data/component/src/main/java/org/quiltmc/qsl/component/api/ComponentProvider.java
```diff
@@ -0,0 +1,54 @@
...
+
+	default <T, U, C extends FunctionComponent<T, U>> Optional<U> call(ComponentType<C> type, T t) {
+		return this.expose(type).map(func -> func.call(this, t));
+	}
+
+	default <C extends Component> Optional<C> expose(ComponentType<C> id) {
+		return Components.expose(id, this);
+	}
+
+	default void sync() {
```
LambdAurora:
```suggestion
	default void syncComponent() {
```

BlockEntity API already injects a sync method for BEs.

LambdAurora (changes requested):
Another global remark: do not use `@NotNull`, we consider that by default everything is not null, and if something might be null we use `@Nullable`.

0xJoeMama:
I actually mark stuff with NotNull for the kotlin devs out there. 
The kotlin compiler is able to parse NotNull and allow the kotlin side to just use the normal call operator(.) and not the safe-call one(?.).

LambdAurora:
> I actually mark stuff with NotNull for the kotlin devs out there. The kotlin compiler is able to parse NotNull and allow the kotlin side to just use the normal call operator(.) and not the safe-call one(?.).

Might need to talk of this with the Kotlin team then.

Pyrofab:
I have not looked at this in too much detail yet, but I would highly discourage attaching a ComponentProvider to item stacks. The ItemStack module was removed in Cardinal Components API because it caused a lot of issues with other mods dealing with ItemStack serialization, had poor performance, required a lot of special casing, and was generally unneeded after the API Lookup API became a thing.

0xJoeMama:
> I would highly discourage attaching a ComponentProvider to item stacks

I can understand why that would be an issue. Honestly I run into multiple problems while creating the itemstack implementation of the ComponentProvider interface. ItemStacks are a weird bunch and they are instantiated quite often. Might have to do some intense thinking about this.

0xJoeMama:
However I must also say that I'd like to personally experiment with ItemStacks, *because* they are weird.
Who knows lazy instantiation might be a life saver. Or some other performance improvement down the line.
We can call the itemstack implementation experimental even when network sync is fully tested.
I want ItemStacks to work and to be honest even if they don't I'd like to see them fail myself when testing with them.

0xJoeMama:
Just not in production.

LambdAurora:
Gonna put this here, if something starts blocking out this PR we always can delay said implementation to a later PR so we can do more thinking on it ;)

Sometimes incremental implementations are easier to manage.

0xJoeMama:
'said implementation' aka ItemStack, am I correct in that?

LambdAurora:
> 'said implementation' aka ItemStack, am I correct in that?

yup, you don't have to get everything at once (and sometimes incremental implementation is easier to review too ahah :p)

0xJoeMama:
If we are talking safety, then Entity and BlockEntity could literally be released right now.
Chunks needs a bit work, to do with initial syncing.
LevelProperties aka 'Level' needs a way to encode itself safely.
ItemStacks are technically not synced through network on our part. That is because we handle ItemStack component nbt under the normal 'nbt' fields in their class. Since that field is synced through PacketByteBuf#writeItemStack we needn't worry about network sync. Issues like frequent instantiation are what are holding ItemStacks back.

0xJoeMama:
> yup, you don't have to get everything at once (and sometimes incremental implementation is easier to review too ahah :p)

'We can call the itemstack implementation experimental even when network sync is fully tested.'
Literally my words. I couldn't agree more.

0xJoeMama:
> easier to review too

By the way, I did think about how much hell reviewing this PR would be. 
Sorry everyone for probably taking a good chunk of time from your day.

LambdAurora:
> > easier to review too
> 
> By the way, I did think about how much hell reviewing this PR would be. Sorry everyone for probably taking a good chunk of time from your day.

It's fine, the API looks cool, I haven't yet jumped too much into reviewing but from afar it's cool stuff!

0xJoeMama:
> from afar it's cool stuff!

It is. And I think I've made it quite powerful. You see, I'm the simple by default, powerful when needed kind of person when it comes to APIs(and other things but that's not relevant here).

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/sync/header/SyncPacketHeader.java
```diff
@@ -17,33 +17,38 @@
...
 		SyncHeaderRegistry.register(CommonInitializer.id("chunk"), CHUNK);
+		SyncHeaderRegistry.register(CommonInitializer.id("level"), LEVEL);
 	}
 
 	public @NotNull PacketByteBuf start(@NotNull ComponentProvider provider) {
 		var buf = PacketByteBufs.create();
 		buf.writeInt(SyncHeaderRegistry.HEADERS.getRawId(this));
-		//noinspection unchecked the person calling is responsible to make sure we get a valid provider instance!
+		// the person calling is responsible to make sure we get a valid provider instance!
+		//noinspection unchecked
```
Leo40Git: `//noinspection` comments only work in IntelliJ! Annotate this method with `@SuppressWarnings("unchecked")` instead.

Leo40Git:
I feel as if `Lazy` should be moved to `qsl_base` as a public API class... but that's out of scope for this PR.

0xJoeMama:
> I feel as if `Lazy` should be moved to `qsl_base` as a public API class... but that's out of scope for this PR.

Well if we are talking moving things out, then I'd suggest NetworkCodec be moved to the networking API. It's actually really useful.

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/sync/codec/NetworkCodec.java
```diff
@@ -0,0 +1,122 @@
...
+			(buf, entity) -> buf.writeInt(entity.getId()),
+			buf -> MinecraftClient.getInstance().world.getEntityById(buf.readInt())
+	);
+	public static final NetworkCodec<Chunk> CHUNK = new NetworkCodec<>(
+			(buf, chunk) -> buf.writeChunkPos(chunk.getPos()),
+			buf -> CHUNK_POS.decode(buf)
+					.map(chunkPos -> MinecraftClient.getInstance().world.getChunk(chunkPos.x, chunkPos.z))
+					.orElseThrow()
+	);
+	public static final NetworkCodec<ItemStack> ITEMSTACK = new NetworkCodec<>(
```
Leo40Git:
```suggestion
	public static final NetworkCodec<ItemStack> ITEM_STACK = new NetworkCodec<>(
```
Correct conversion between class name (`PascalCase`) and `static final` field name (`SCREAMING_SNAKE_CASE`).

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/sync/codec/NetworkCodec.java
```diff
@@ -0,0 +1,122 @@
...
+	);
+	public static final NetworkCodec<UUID> UUID = new NetworkCodec<>(
+			PacketByteBuf::writeUuid, PacketByteBuf::readUuid
+	);
+	public static final NetworkCodec<ComponentType<?>> COMPONENT_TYPE = new NetworkCodec<>(
+			(buf, componentType) -> buf.writeInt(Components.REGISTRY.getRawId(componentType)),
+			buf -> ClientSyncHandler.getInstance().getType(buf.readInt())
+	);
+
+	public static <O, L extends List<O>> NetworkCodec<L> list(NetworkCodec<O> entryCodec, Function<Integer, L> function) {
```
Leo40Git:
```suggestion
	public static <O, L extends List<O>> NetworkCodec<L> list(NetworkCodec<O> entryCodec, Function<Integer, L> listFactory) {
```
Or alternatively introduce a custom interface for this:
```java
@FunctionalInterface
public interface ListFactory<E> {
    List<E> createList(int size);
}
```

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Maybe.java
```diff
@@ -0,0 +1,517 @@
...
+ * and quality of life changes.<br/>
+ *
+ * Can either be {@link Just} or {@link Nothing}.<br/>
+ * {@link Just} instances contain a value of type {@link T}.<br/>
+ * {@link Nothing} instance all reference the {@link Nothing} singleton and contain no data.<br/>
+ *
+ * @param <T> The type of the wrapped object
+ * @author 0xJoeMama
+ */
+public sealed abstract class Maybe<T> permits Maybe.Just, Maybe.Nothing {
```
LemmaEOF: oh right, sealed classes got finalized! very nice use of them ^.^

0xJoeMama: Wait, are they not finalized? Is there any issue with this?
I could resolve them by making the Maybe constructor private if there is some issue.

LemmaEOF: they are finalized now! I just forgot that they were because I hadn't seen them used yet lol, this is more than fine

0xJoeMama: _Not gonna lie, I thought you were about to yeet out my monad_

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Maybe.java
```diff
@@ -0,0 +1,517 @@
...
+	 *
+	 * @param <T> In this case this type parameter is unused.
+	 */
+	public static final class Nothing<T> extends Maybe<T> {
+		/**
+		 * A singleton instance is used for Nothing since it doesn't contain data and can therefore be cast to any <code>Maybe</code> type.
+		 */
+		private static final Nothing<?> INSTANCE = new Nothing<>();
+
+		private Nothing() {
```
EnnuiL: Nitpick: Drop the line break, turn this into `{ }`

library/data/component/src/main/java/org/quiltmc/qsl/component/api/ComponentType.java
```diff
@@ -0,0 +1,68 @@
...
+		if (this.isStatic) {
+			return STATIC_CACHE.getOrCreate(this, saveOperation, syncOperation);
+		}
+		return this.defaultFactory.create(saveOperation, syncOperation);
+	}
+
+	public static class Static {
+		private final Map<ComponentType<?>, Component> staticInstances = new IdentityHashMap<>();
+
+		private Static() {
```
EnnuiL: Nitpick: Turn this into a `{ }`, economize lines

library/data/component/src/main/java/org/quiltmc/qsl/component/api/ComponentType.java
```diff
@@ -0,0 +1,68 @@
...
+
+		}
+
+		@SuppressWarnings("unchecked")
+		<C extends Component> C getOrCreate(ComponentType<C> type, Runnable saveOperation, Runnable syncOperation) {
```
EnnuiL: The visibility marker (...? I don't remember the name) should be explicit on here

library/data/component/src/main/java/org/quiltmc/qsl/component/api/components/FloatComponent.java
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.quiltmc.qsl.component.api.components;
+
+import org.quiltmc.qsl.component.api.Component;
+
```
EnnuiL: Too many lines
```suggestion
```

library/data/component/src/main/java/org/quiltmc/qsl/component/api/components/NbtComponent.java
```diff
@@ -0,0 +1,69 @@
...
+	static void readFrom(NbtComponent<?> nbtComponent, Identifier id, NbtCompound root) {
+		NbtElement nbtTarget = root.get(id.toString());
+		switch (nbtComponent.nbtType()) {
+			case NbtElement.BYTE_TYPE -> ((NbtComponent<NbtByte>) nbtComponent).read((NbtByte) nbtTarget);
+			case NbtElement.SHORT_TYPE -> ((NbtComponent<NbtShort>) nbtComponent).read((NbtShort) nbtTarget);
+			case NbtElement.INT_TYPE -> ((NbtComponent<NbtInt>) nbtComponent).read((NbtInt) nbtTarget);
+			case NbtElement.LONG_TYPE -> ((NbtComponent<NbtLong>) nbtComponent).read((NbtLong) nbtTarget);
+			case NbtElement.FLOAT_TYPE -> ((NbtComponent<NbtFloat>) nbtComponent).read((NbtFloat) nbtTarget);
+			case NbtElement.DOUBLE_TYPE -> ((NbtComponent<NbtDouble>) nbtComponent).read((NbtDouble) nbtTarget);
+			case NbtElement.BYTE_ARRAY_TYPE ->
```
EnnuiL: The line breaks here seems way too arbitrary on this wall of text (oh, and on a kinda unrelated note: I love switch expressions);
Yes yes, there's a char limit per line and such; Ignore 'em here, remove the line breaks on BYTE_ARRAY_TYPE and LONG_ARRAY_TYPE; These limits do make sense as a style guide and you should follow them when it makes sense, but strictly enforcing them just because it passed few characters from it doesn't make sense (heck, iiirc, Linus Torvalds complained about the strictness of that)

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/CommonInitializer.java
```diff
@@ -0,0 +1,47 @@
...
+import org.quiltmc.loader.api.ModContainer;
+import org.quiltmc.qsl.base.api.entrypoint.ModInitializer;
+import org.quiltmc.qsl.component.impl.sync.ServerSyncHandler;
+import org.quiltmc.qsl.component.impl.sync.header.SyncPacketHeader;
+import org.quiltmc.qsl.lifecycle.api.event.ServerLifecycleEvents;
+import org.quiltmc.qsl.networking.api.ServerLoginConnectionEvents;
+
+@ApiStatus.Internal
+public final class CommonInitializer implements ModInitializer {
+	public static final String MODID = "quilt_component";
```
EnnuiL: Rename MODID to MOD_ID
```suggestion
	public static final String MOD_ID = "quilt_component";
```

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/ComponentsImpl.java
```diff
@@ -0,0 +1,74 @@
...
+public class ComponentsImpl {
+	public static final RegistryKey<Registry<ComponentType<?>>> REGISTRY_KEY =
+			RegistryKey.ofRegistry(CommonInitializer.id("component_types"));
+	public static final Registry<ComponentType<?>> REGISTRY = // TODO: Register this, maybe?!
+			new SimpleRegistry<>(REGISTRY_KEY, Lifecycle.experimental(), null);
+
+	public static final CachedInjectionManager CACHED_MANAGER = new CachedInjectionManager();
+	public static final DynamicInjectionManager DYNAMIC_MANAGER = new DynamicInjectionManager();
+
+	public static final Logger LOGGER = LoggerFactory.getLogger("QSL/Components");
```
EnnuiL: While there is no standard on the logger names, we can at least begin to standardize to one
```suggestion
	public static final Logger LOGGER = LoggerFactory.getLogger("Quilt Component");
```
This reminds me, i should get everything else to this format

library/data/component/build.gradle
```diff
@@ -0,0 +1,70 @@
+plugins {
+	id("qsl.module")
+}
+
+qslModule {
+	name = "Quilt Component Attachment API"
```
EnnuiL: Hmm, the divergence between the mod name (Quilt Component Attachment API) and the mod ID (quilt_component) is a bit weird here, and thinking about it, "Component Attachment" doesn't really make sense compared to "Registry Entry Attachments", since for the latter, registry attachments basically attach values to their registry entries, while components aren't the thing being attached on

library/data/component/build.gradle
```diff
@@ -0,0 +1,70 @@
...
+	moduleName = "component"
+	id = "quilt_component"
+	description = "API allowing attachment and querying of an ECS-like interface for entities, block-entities, chunks, etc."
```
EnnuiL: Hmm, this description needs a bit of improvement, although I'm leaving this for my second look later

library/data/component/src/main/java/org/quiltmc/qsl/component/api/provider/ComponentProvider.java
```diff
@@ -0,0 +1,49 @@
...
+import net.minecraft.world.chunk.Chunk;
+import org.quiltmc.qsl.base.api.util.InjectedInterface;
+import org.quiltmc.qsl.base.api.util.Maybe;
+import org.quiltmc.qsl.component.api.Component;
+import org.quiltmc.qsl.component.api.container.ComponentContainer;
+import org.quiltmc.qsl.component.api.ComponentType;
+import org.quiltmc.qsl.component.api.Components;
+import org.quiltmc.qsl.component.impl.container.AbstractComponentContainer;
+
+@InjectedInterface({
```
EnnuiL: The double tabs should be just one here

Future Me here: oh god, there's a heckton of double tabs everywhere; They all (or at least the majority of them) should be reduced to a single one

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/container/AbstractComponentContainer.java
```diff
@@ -0,0 +1,119 @@
...
+			}
+		});
+
+		this.addComponent(type, component);
+
+		return component;
+	}
+
+	public record ContainerOperations(Runnable saveOperation,
+									  Function<ComponentType<?>, Runnable> syncOperationFactory) {
```
EnnuiL: eugh, this thing should be in a single line

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/container/EmptyComponentContainer.java
```diff
@@ -0,0 +1,60 @@
...
+import org.quiltmc.qsl.component.api.provider.ComponentProvider;
+import org.quiltmc.qsl.component.api.ComponentType;
+
+public final class EmptyComponentContainer implements ComponentContainer {
+	public static final EmptyComponentContainer INSTANCE = new EmptyComponentContainer();
+
+	public static final ComponentContainer.Factory<EmptyComponentContainer> FACTORY =
+			(provider, injections, saveOperation, ticking, syncContext) -> INSTANCE;
+
+	private EmptyComponentContainer() {
```
EnnuiL: Turn this (and the below empty methods) into a `{ }`

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/container/LazyComponentContainer.java
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	@Override
+	public Maybe<Component> expose(ComponentType<?> id) {
+		return Maybe.wrap(this.components.get(id)).map(Lazy::get);
+	}
+
+	@Override
+	protected <COMP extends Component> void addComponent(ComponentType<COMP> type, Component component) {
+
```
EnnuiL: `{ }`

library/data/component/src/main/java/org/quiltmc/qsl/component/impl/util/StringConstants.java
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.quiltmc.qsl.component.impl.util;
+
+public final class StringConstants {
+	public static final String COMPONENT_ROOT = "QslComponentRoot";
```
EnnuiL:
```suggestion
	public static final String COMPONENT_ROOT = "QuiltComponentRoot";
```

library/data/component/src/testmod/resources/quilt_component_test.mixins.json
```diff
@@ -0,0 +1,15 @@
+{
+  "required": true,
+  "package": "org.quiltmc.qsl.component.test.mixin",
+  "compatibilityLevel": "JAVA_17",
+  "minVersion": "0.8",
+  "mixins": [
```
EnnuiL, line 5: This field can be removed

library/data/component/src/testmod/java/org/quiltmc/qsl/component/test/client/TestBeRenderer.java
```diff
@@ -0,0 +1,54 @@
...
+import net.minecraft.client.render.model.json.ModelTransformation;
+import net.minecraft.client.util.math.MatrixStack;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.math.Vec3f;
+import org.quiltmc.qsl.component.test.ComponentTestMod;
+import org.quiltmc.qsl.component.test.TestBlockEntity;
+
+import java.util.function.Predicate;
+
+public class TestBeRenderer implements BlockEntityRenderer<TestBlockEntity> {
```
EnnuiL: This should be named `TestBlockEntityRenderer`

EnnuiL (changes requested):
I'll do my review into two passes, since yeah, there's a lot to see and to be done;

On this first pass, yeah, it's pretty much style and standarization issues; I recommend using the project's Checkstyle config in order to clean up the rest of them, 'cause yeah, this isn't an exaustive one
As for the logic itself? On a glance, all seems great! Although I'm curious about whenever there are holes to poke on or not, and yeah, I'll have to actually mess around with code and such for that

0xJoeMama:
Oh, by the way. Dynamic injections don't use the event system anymore.

library/data/component/src/main/java/org/quiltmc/qsl/component/api/ComponentType.java
Leo40Git: Visibility modifier, IIRC.

And not having one _is_ valid - it means the method is _package-private_, and can only be accessed by classes in the same package.

library/data/component/src/main/java/org/quiltmc/qsl/component/api/components/NbtComponent.java
0xJoeMama: IntelliJ autoformatting got me on those. I actually just had them all be one liners.

library/data/component/src/testmod/resources/quilt_component_test.mixins.json
0xJoeMama: I suppose you mean minVersion, right?

EnnuiL: No, I mean the empty mixins field; minVersion is fine

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Maybe.java
```diff
@@ -0,0 +1,516 @@
...
+	 */
+	public static <T> Maybe<T> just(T value) {
+		return new Just<>(value);
+	}
+
+	/**
+	 * Similar function to {@link Optional#empty()}.
+	 *
+	 * @param <T> The type the {@link Nothing} instance will take the form of.
+	 * @return the singleton {@link Nothing} instance after casting it to the proper type.
```
Leo40Git:
```suggestion
	 * @return A {@link Nothing} instance.
```
Exposing this implementation details is just gonna make people do `maybe == Maybe.nothing()`.

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Maybe.java
```diff
@@ -0,0 +1,516 @@
...
+	/**
+	 * Works like {@link Maybe#unwrapOr} except that it lazily evaluated the return value using the provided {@link Supplier}.
+	 *
+	 * @param supplier A {@link Supplier} that will create the {@link T} value to be returns if the current instance is {@link Nothing}.
+	 * @return The current {@link T} value if the current instance if {@link Just},
+	 * otherwise the value produces by calling {@link Supplier#get()} on the provided {@link Supplier}.
+	 * @see Optional#orElseGet
+	 */
+	public abstract T unwrapOrGet(Supplier<? extends T> supplier);
+
```
Leo40Git: Maybe add an `unwrapOrThrow(Supplier<? extends Throwable> exceptionSupplier)` method?

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Maybe.java
```diff
@@ -0,0 +1,516 @@
...
+	 * In that case it returns the provided {@link T} value.
+	 *
+	 * @param defaultValue The value to get if the current instance if {@link Nothing}.
+	 * @return The current {@link T} value if the current instance is {@link Just}, otherwise the provided {@link T} value.
+	 * @see Optional#orElse
+	 */
+	public abstract T unwrapOr(T defaultValue);
+
+	/**
+	 * Works like {@link Maybe#unwrapOr} except that it lazily evaluated the return value using the provided {@link Supplier}.
```
Leo40Git:
```suggestion
	 * Works like {@link Maybe#unwrapOr} except that it lazily evaluated the return value using the provided {@link Supplier}.
```
```suggestion
	 * Works like {@link Maybe#unwrapOr} except that it lazily evaluates the return value using the provided {@link Supplier}.
```

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Maybe.java
```diff
@@ -267,6 +267,15 @@ public static <T> Maybe<T> fromOptional(Optional<T> opt) {
...
 	public abstract T unwrapOrGet(Supplier<? extends T> supplier);
 
+	/**
+	 * Works like {@link Maybe#unwrap} except it throws the provided {@link Throwable} and not the default one.
+	 *
+	 * @param throwableSupplier The {@link Throwable} to throw.
+	 * @return The contained value, if it exists.
+	 * @throws Throwable A customizable throwable.
+	 */
+	public abstract T unwrapOrThrow(Supplier<? extends Throwable> throwableSupplier) throws Throwable;
```
Leo40Git:
```suggestion
	public abstract <E extends Throwable> T unwrapOrThrow(Supplier<? extends E> throwableSupplier) throws E;
```
This prevents checked exceptions from being thrown without javac realizing (or at least, makes it harder to do this accidentally).

0xJoeMama:
Consider the initial feature-set done! Feel free to start breaking this!

library/data/component/src/main/resources/quilt_component.mixins.json
```diff
@@ -0,0 +1,25 @@
+{
+  "required": true,
+  "package": "org.quiltmc.qsl.component.mixin",
+  "compatibilityLevel": "JAVA_17",
+  "mixins": [
+    "RegistryAccessor",
```
oliviathevampire: This errors? Does this mixin even exist?

0xJoeMama: Woops my bad!

library/data/component/src/main/java/org/quiltmc/qsl/component/api/provider/ComponentProvider.java
```diff
@@ -0,0 +1,92 @@
...
+import org.quiltmc.qsl.component.api.ComponentType;
+import org.quiltmc.qsl.component.api.Components;
+import org.quiltmc.qsl.component.api.container.ComponentContainer;
+
+/**
+ * Any object that wishes to allow components to be attached to and queried from it,
+ * needs to implement this interface.<br/>
+ *
+ * <p>
+ * Currectly it consists of the {@link ComponentProvider#getComponentContainer()} method, which is the abstract one,
```
oliviathevampire, line 40: ``Currently,`` instead of ``Currectly``

library/data/component/src/main/java/org/quiltmc/qsl/component/api/provider/ComponentProvider.java
```diff
@@ -0,0 +1,92 @@
...
+ * Furthermore, all of them support component saving and syncing.<br/>
+ *
+ * <b>Note: You shouldn't have to use the {@link ComponentProvider#getComponentContainer()} method that often,
+ * if ever, because it's needed for implementations not for using the API.</b>
+ *
+ * @author 0xJoeMama
+ * @see ComponentContainer
+ * @see Components
+ */
+@InjectedInterface({ // We inject this inteface, so that modders don't need to use the methods in Components directly with our default implementations.
```
oliviathevampire, line 66: ``interface`` instead of ``inteface``

library/data/component/src/main/java/org/quiltmc/qsl/component/api/provider/ComponentProvider.java
```diff
@@ -0,0 +1,92 @@
...
+		BlockEntity.class,
+		Chunk.class,
+		MinecraftServer.class, // MinecraftServer and MinecraftClient contain Level components
+		MinecraftClient.class,
+		World.class
+})
+public interface ComponentProvider {
+	/**
+	 * Every {@linkplain ComponentProvider provider} must provide a {@link ComponentContainer},
+	 * so that it can store the components targetting or that are manually added to it.
```
oliviathevampire, line 77: ``targeting`` instead of ``targetting``

library/data/component/src/main/java/org/quiltmc/qsl/component/api/provider/ComponentProvider.java
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	/**
+	 * Utility method to call {@link Components#expose(ComponentType, Object)} on the actual provider.
+	 *
+	 * @param type The {@linkplain ComponentType type} we want to query.
+	 * @param <C>  The type of the held component.
+	 * @return A {@link Maybe} instance following the rules defined in {@link Components#expose(ComponentType, Object)}.
+	 */
+	default <C> Maybe<C> expose(ComponentType<C> type) {
```
lukebemish: Is there a particular benefit to exposing a `Maybe`/`Optional` here as opposed to simply returning a `@Nullable` value? My initial reaction, especially given that this method could be run quite often and constructing lots of objects could add some cost (though I'm not actually sure how bad this would be, since I haven't benchmarked it or anything, so this could be premature optimization) is that it's better to return a nullable value. However, I may be missing something regarding this.

0xJoeMama: In the default implementations of ComponentContainer, the optional object is cached and so no object is created except the one at the initialization of the container.

lukebemish: Ah. I meant to leave this comment on `ComponentContainer#expose`, which your javadocs note is called *really* often

0xJoeMama: It's okay, this method also propagates to that method. In other words the same applies here as well.

lukebemish: Yes. Personally, I'd propose switching both to just be nullable and cut out Maybe entirely.

0xJoeMama: Well, thing is the reason it's Optional in the first place is for API's sake. Optional code is much cleaner and understandable to look at, than nullable code in my opinion. However, there is the solution of adding utility methods in the ComponentProvider interface, that will make the code cleaner. I'll give those a try, and see what it'll look like.

Thanks for the suggestion!

PR #149, opened 2022-07-08
Multipart Entity Module

Platymemo:
Adds a new module `multipart` that allows easier and more compatible creation of multipart entities.

There are two main interfaces: `MultipartEntity` and `EntityPart`.
`MultipartEntity` represents an `Entity` that has sub-parts, which implement `EntityPart`.

This module also implements these onto the `EnderDragonEntity` and `EnderDragonPart`.

Additionally, it applies two fixes to vanilla bugs:
- Offsets the `id`s for `EnderDragonPart`s by 1 so they don't overlap with the dragon itself.
- `EntityPart`s are tracked separately from their parents so they can be looked up even when the parent is not in the lookup area.


Leo40Git (approved):
LGTM!

williambl:
Going to do code review later, but for now I just want to say I'm really happy to have this. [Writing a lib for this in the past](https://github.com/williambl/multipart-entities) was quite difficult, and this API will make lots of mod ideas more achievable.

library/entity/multipart/src/testmod/resources/quilt.mod.json
```diff
@@ -0,0 +1,22 @@
...
+    "group": "org.quiltmc.qsl.multipart",
+    "id": "qsl_multipart_testmod",
+    "version": "1.0.0",
+    "metadata": {
+      "name": "Quilt Multipart API Test Mod",
+      "license": "Apache-2.0"
+    },
+    "intermediate_mappings": "net.fabricmc:intermediary",
+    "load_type": "always",
+    "entrypoints": {
```
EnnuiL: Nitpick: You could get rid of the `entrypoints` field here

library/entity/multipart/src/testmod/java/org/quiltmc/qsl/multipart/test/mixin/CreeperEntityMixin.java
```diff
@@ -0,0 +1,66 @@
...
+		this.secretHitbox.lastRenderY = this.lastRenderY + 1.1;
+		this.secretHitbox.lastRenderZ = this.lastRenderZ;
+	}
+
+	@Override
+	public void onSpawnPacket(EntitySpawnS2CPacket packet) {
+		super.onSpawnPacket(packet);
+		EntityPart<?>[] parts = this.getEntityParts();
+
+		for(int i = 0; i < parts.length; ++i) {
```
EnnuiL: Nitpick (which might be a personal bias? although I'm not aware of any other places on QSL using `++i`):
```suggestion
		for(int i = 0; i < parts.length; i++) {
```

library/entity/multipart/src/testmod/java/org/quiltmc/qsl/multipart/test/SecretCreeperPart.java
```diff
@@ -0,0 +1,76 @@
...
+	private final EntityDimensions partDimensions;
+
+	public SecretCreeperPart(CreeperEntity creeper, float f, float g) {
+		super(creeper.getType(), creeper.world);
+		this.partDimensions = EntityDimensions.changing(f, g);
+		this.calculateDimensions();
+		this.owner = creeper;
+	}
+
+	protected void initDataTracker() {
```
EnnuiL: Nitpick: Go for `{ }` instead of line breaks

EnnuiL: waaaaait a hecking minute; Are those Entity methods? Yeah, you should add a `@Override` to them

library/entity/multipart/src/main/java/org/quiltmc/qsl/multipart/mixin/client/EntityRenderDispatcherMixin.java
```diff
@@ -0,0 +1,64 @@
...
+					double entityPartX = entityX + MathHelper.lerp(tickDelta, entityPart.lastRenderX, entityPart.getX());
+					double entityPartY = entityY + MathHelper.lerp(tickDelta, entityPart.lastRenderY, entityPart.getY());
+					double entityPartZ = entityZ + MathHelper.lerp(tickDelta, entityPart.lastRenderZ, entityPart.getZ());
+					matrices.translate(entityPartX, entityPartY, entityPartZ);
+					WorldRenderer.drawBox(matrices, vertices, entityPart.getBoundingBox().offset(-entityPart.getX(), -entityPart.getY(), -entityPart.getZ()), 0.25F, 1.0F, 0.0F, 1.0F);
+					matrices.pop();
+				}
+			}
+		}
+
```
EnnuiL: Nitpick:
```suggestion
```

library/entity/multipart/src/main/java/org/quiltmc/qsl/multipart/mixin/WorldMixin.java
```diff
@@ -0,0 +1,83 @@
...
+		return false;
+	}
+
+	/**
+	 * Fixes <a href="https://bugs.mojang.com/browse/MC-158205">MC-158205</a>
+	 * <p>Allows collecting {@link EntityPart}s that are within the targeted {@link Box}
+	 * but are part of {@link Entity entities} in unchecked chunks.</p>
+	 */
+	@Inject(method = "getOtherEntities", at = @At("RETURN"))
+	void getOtherEntityParts(Entity except, Box box, Predicate<? super Entity> predicate, CallbackInfoReturnable<List<Entity>> cir) {
```
EnnuiL: The lack of explicit visibility markers are pretty weird; Add one here

library/entity/multipart/src/main/java/org/quiltmc/qsl/multipart/mixin/WorldMixin.java
```diff
@@ -0,0 +1,83 @@
...
+		}
+	}
+
+	/**
+	 * Fixes <a href="https://bugs.mojang.com/browse/MC-158205">MC-158205</a>
+	 * <p>Allows collecting {@link EntityPart}s that are within the targeted {@link Box}
+	 * but are part of {@link Entity entities} in unchecked chunks.</p>
+	 */
+	@Inject(method = "getEntitiesByType", at = @At("RETURN"))
+	<T extends Entity> void getEntityPartsByType(TypeFilter<Entity, T> filter, Box box, Predicate<? super T> predicate, CallbackInfoReturnable<List<T>> cir) {
```
EnnuiL: Here too

library/entity/multipart/src/main/java/org/quiltmc/qsl/multipart/mixin/PlayerEntityMixin.java
```diff
@@ -0,0 +1,35 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.ModifyVariable;
+
+@Mixin(PlayerEntity.class)
+public class PlayerEntityMixin {
+	@ModifyVariable(method = "attack", at=@At(value = "STORE"), ordinal = 1)
+	private Entity attackEntityPart(Entity entity) {
+		if (entity instanceof EntityPart<?> entityPart) {
+			return entityPart.getOwner();
+		}
```
EnnuiL: Nitpick: there should be a line break here:
```suggestion
		}

```

library/entity/multipart/src/main/java/org/quiltmc/qsl/multipart/api/AbstractEntityPart.java
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	@Override
+	public E getOwner() {
+		return this.owner;
+	}
+
+	@Override
+	protected void initDataTracker() {
+
```
EnnuiL: Make 3 lines into 1, use `{ }` on these methods

library/entity/build.gradle
```diff
@@ -0,0 +1,7 @@
+plugins {
+	id("qsl.library")
+}
+
+qslLibrary {
+	libraryName = "entity"
+}
```
EnnuiL, line 7: I see a lack of final new lines on all the files; EditorConfig should help with adding them

EnnuiL (changes requested):
The API itself looks great! I just have a heckton of nitpicks to be addressed

library/entity/multipart/build.gradle
```diff
@@ -0,0 +1,25 @@
...
+	id = "quilt_multipart"
+	description = "An API to support custom multipart entities."
+	library = "entity"
+	moduleDependencies {
+		core {
+			api("qsl_base")
+		}
+	}
+	injectedInterface("net/minecraft/class_1937") {
+		values = ["org/quiltmc/qsl/multipart/impl/EntityPartTracker"]
```
EnnuiL: A little less nitpicky nitpick: Those interfaces to be injected? Add an `@InjectedInterface` annotation to all of them

EnnuiL (changes requested):
Whoops, I have one more

library/entity/multipart/src/testmod/resources/quilt.mod.json
Platymemo: I had left it in with the thought that maybe I'd also make a whole test entity with parts, but I think the creeper mixin suffices.

library/entity/multipart/src/testmod/java/org/quiltmc/qsl/multipart/test/mixin/CreeperEntityMixin.java
Platymemo: I just copied the for loop from the ender dragon code so it doesn't really matter. I use postfix generally anyways

library/entity/multipart/src/main/java/org/quiltmc/qsl/multipart/mixin/WorldMixin.java
```diff
@@ -0,0 +1,85 @@
...
+	/**
+	 * Fixes <a href="https://bugs.mojang.com/browse/MC-158205">MC-158205</a>
+	 * <p>Allows collecting {@link EntityPart}s that are within the targeted {@link Box}
+	 * but are part of {@link Entity entities} in unchecked chunks.</p>
+	 */
+	@Inject(method = "getOtherEntities", at = @At("RETURN"))
+	private void getOtherEntityParts(Entity except, Box box, Predicate<? super Entity> predicate, CallbackInfoReturnable<List<Entity>> cir) {
+		List<Entity> list = cir.getReturnValue();
+		for (Entity part : this.getEntityParts().values()) {
+			if (part != except && part.getBoundingBox().intersects(box) && predicate.test(part)) {
```
0xJoeMama: Is this supposed to not check if the part is part of the except entity?

library/entity/multipart/src/main/java/org/quiltmc/qsl/multipart/mixin/WorldMixin.java
```diff
@@ -0,0 +1,85 @@
...
+	 * Fixes <a href="https://bugs.mojang.com/browse/MC-158205">MC-158205</a>
+	 * <p>Allows collecting {@link EntityPart}s that are within the targeted {@link Box}
+	 * but are part of {@link Entity entities} in unchecked chunks.</p>
+	 */
+	@Inject(method = "getEntitiesByType", at = @At("RETURN"))
+	private <T extends Entity> void getEntityPartsByType(TypeFilter<Entity, T> filter, Box box, Predicate<? super T> predicate, CallbackInfoReturnable<List<T>> cir) {
+		List<T> list = cir.getReturnValue();
+		for (Entity part : this.getEntityParts().values()) {
+			T entity = filter.downcast(part);
+			if (entity != null && entity.getBoundingBox().intersects(box) && predicate.test(entity)) {
```
0xJoeMama: Same here.

library/entity/multipart/src/main/java/org/quiltmc/qsl/multipart/mixin/EnderDragonEntityMixin.java
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	/**
+	 * <p>The vanilla method sets the {@link EnderDragonEntity#head} to
+	 * the same {@link Entity#getId() id} as the {@link EnderDragonEntity} herself.</p>
+	 */
+	@ModifyArg(method = "onSpawnPacket", at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/boss/dragon/EnderDragonPart;setId(I)V"))
+	private int setProperId(int oldId) {
+		return oldId + 1;
+	}
```
Platymemo: This would be a desync issue when you have a Quilt client (with this library installed of course) connecting to a vanilla-compatible server

Platymemo (commented):
Just realized something painfully obvious and kind of important

EDIT: That something being this review is wrong because I am silly

library/entity/multipart/src/main/java/org/quiltmc/qsl/multipart/mixin/EnderDragonEntityMixin.java
Platymemo: I'm just in a silly goofy mood. The vanilla server expects this behavior, it's the vanilla client that has a bug. See <https://bugs.mojang.com/browse/MC-225055>. That bug is due to entity id desyncs.

Platymemo: Technically, that makes qsl a cheat mod because now you have an unfair advantage over vanilla clients. Hmmm

library/entity/multipart/src/main/java/org/quiltmc/qsl/multipart/mixin/WorldMixin.java
Platymemo: Oops, good catch!

library/entity/multipart/src/main/java/org/quiltmc/qsl/multipart/mixin/WorldMixin.java
Platymemo: I'm conflicted with this one because, for example, maybe you want to filter by entity parts specifically. Vanilla behavior is to check the owner and then the parts if the owner passes the filter, but this limits the ability of this function

library/entity/multipart/src/main/java/org/quiltmc/qsl/multipart/mixin/EnderDragonEntityMixin.java
```diff
@@ -0,0 +1,49 @@
...
+	public abstract EnderDragonPart[] getBodyParts();
+
+	@Override
+	public EntityPart<?>[] getEntityParts() {
+		return this.getBodyParts();
+	}
+
+	/**
+	 * <p>The vanilla method sets the {@link EnderDragonEntity#head} to
+	 * the same {@link Entity#getId() id} as the {@link EnderDragonEntity} herself.</p>
```
williambl: Maybe mention that this fixes https://bugs.mojang.com/browse/MC-225055 in here?

williambl (approved):
A `package-info.java` with a brief overview of the API would be cool. API itself LGTM other than the things pointed out by 0xJoeMama, though - I played around with it a bit, it's awesome :D

library/entity/multipart/src/main/java/org/quiltmc/qsl/multipart/mixin/WorldMixin.java
williambl:
```suggestion
			if (!part.isPartOf(except) && part.getBoundingBox().intersects(box) && predicate.test(part)) {
```

EnnuiL (approved):
And now you have the Ennui's Stamp of Approval! which doesn't count for FCP but it's still good

library/entity/multipart/src/main/java/org/quiltmc/qsl/multipart/api/AbstractEntityPart.java
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	/**
+	 * Rotates this {@link AbstractEntityPart} about the pivot point with the given rotation.
+	 * @param pivot the pivot point to rotate about in relative coordinates
+	 * @param pitch the rotation about z-axis in degrees
+	 * @param yaw the rotation about y-axis in degrees
+	 * @param roll the rotation about x-axis in degrees
+	 */
+	public void rotate(Vec3d pivot, float pitch, float yaw, float roll) {
```
Arathain: Just what I was asking for! Works well, and should cover all usecases.

library/entity/multipart/src/main/java/org/quiltmc/qsl/multipart/api/AbstractEntityPart.java
```diff
@@ -0,0 +1,256 @@
...
+	public boolean collides() {
+		return true;
+	}
+
+	/**
+	 * Scales the {@link EntityPart}'s {@link EntityDimensions dimensions}.
+	 * @param width the ratio to scale the bounding width by
+	 * @param height the ratio to scale the bounding height by
+	 */
+	public void scale(float width, float height) {
```
Arathain:
```suggestion
	public void setScale(float width, float height) {
```
scale is both a noun and verb - it might be more consistent to use setScale so as to clear up the ambiguity of using nouns as method names

library/entity/multipart/src/main/java/org/quiltmc/qsl/multipart/api/AbstractEntityPart.java
```diff
@@ -0,0 +1,256 @@
...
+		this.widthRatio = width;
+		this.heightRatio = height;
+		this.calculateDimensions();
+	}
+
+	/**
+	 * Scales the {@link EntityPart}'s {@link EntityDimensions dimensions}.
+	 * @param ratio the ratio to scale the bounding box by
+	 */
+	public void scale(float ratio) {
```
Arathain:
```suggestion
	public void setScale(float ratio) {
```
same as above.

Arathain (approved):
Great work! just a single nitpick with the scale() method name, otherwise good to go.

library/entity/multipart/src/main/java/org/quiltmc/qsl/multipart/api/AbstractEntityPart.java
williambl: Disagree. We also have `rotate` and `move` in the same class, having `scale` as a verb here is more consistent.

williambl (approved):

library/entity/multipart/src/main/java/org/quiltmc/qsl/multipart/api/AbstractEntityPart.java
Platymemo: maybe author bias, but I prefer `scale` over `setScale` because `setScale` is maybe too close in name and meaning to `setDimensions`. The difference in form also represents the difference in the use case. `setScale` sounds like a one-off, while `scale` is continuous. For one-offs, you have `setDimensions`.

library/entity/multipart/build.gradle
```diff
@@ -0,0 +1,25 @@
+plugins {
+	id("qsl.module")
+}
+
+qslModule {
+	name = "Quilt Multipart Entity API"
+	moduleName = "multipart"
+	id = "quilt_multipart"
```
LambdAurora:
```suggestion
	id = "quilt_entity_multipart"
```

This is entity related, it should state it.
There could be other stuff that could use the multipart wording like blocks (beacon with their base could be called multipart).

This also must be reflected in the package (`org.quiltmc.qsl.entity.multipart`).

library/entity/multipart/src/main/java/org/quiltmc/qsl/multipart/impl/EntityPartTracker.java
```diff
@@ -0,0 +1,31 @@
...
+import it.unimi.dsi.fastutil.ints.Int2ObjectMap;
+import org.jetbrains.annotations.ApiStatus;
+
+import net.minecraft.entity.Entity;
+import net.minecraft.world.World;
+
+import org.quiltmc.qsl.base.api.util.InjectedInterface;
+
+@ApiStatus.Internal
+@InjectedInterface(World.class)
```
LambdAurora: I think this should not use interface injection as it's implementation.

I also think that the method should be `quilt$getEntityParts` if it stays implementation only.

LambdAurora (changes requested):
Tested the API, confirmed working.

I honestly really like it!
I also allowed myself to commit a quick reformat/minor fixing of some code style stuff just so it's not forgotten and having a "suggested change" for each would have been really annoying :p 

Now for the issues:
 - mentioned in one of the attached comments: the mod id cannot be just quilt_multipart as other stuff could use the multipart wording.
 - the package needs to have entity too (due to multipart being too broad)
 - impl interfaces should not get interface injected  
 
Once those issues are fixed I believe we can get this PR into FCP.

library/entity/multipart/src/main/java/org/quiltmc/qsl/multipart/api/AbstractEntityPart.java
```diff
@@ -0,0 +1,276 @@
...
+		this.owner = owner;
+	}
+
+	@Override
+	public E getOwner() {
+		return this.owner;
+	}
+
+	@Override
+	protected void initDataTracker() {
```
EnnuiL: Pulling a sneaky one, I see (although yeah, it's likely to be an auto-formatter's fault, which makes sense); `{ }`

LambdAurora: Auto-formatter doesn't like those at all.
I tried and tried to fix that but it just doesn't want to.

EnnuiL: Hm, in order to compensate for that, perhaps Checkstyle should be configured so when the auto-formatter kicks in, red squiggly lines would appear; I'll investigate that later

LambdAurora: I found the setting, we're saved.

build-logic/src/main/java/qsl/internal/Versions.java
```diff
@@ -21,7 +21,7 @@ public final class Versions {
 	/**
 	 * The QSL version
 	 */
-	public static final String QSL_VERSION = "2.0.0-beta.15";
+	public static final String QSL_VERSION = "2.0.0-beta.16";
```
EnnuiL: I was gonna complain about the changes on Versions *but* they were actually done on upstream instead; oh god, what did you do for the merge

Platymemo: :iea: I just let IntelliJ do the heavy lifting, I have no idea why the merges happened that way

Leo40Git (approved):
what on earth did you do for the merge???

Uh, anyway, apparently my review was requested.
The actual module LGTM!

LambdAurora (approved):
Putting an approve for now.

All my issues were fixed, but I just remembered of something.
Injected interface have one last flaw, in dev environment there's no compile time implementation of the injected interface since it's done via mixin, meaning it's patched at runtime.
This sadly means that javac will explode if someone extends EnderDragonEntity in dev env and doesn't implement themselves the already implemented methods.

One of the possible fix is to put a default body on those injected interfaces.
Or to do nothing.

In this case I'm not sure which one is best.

Platymemo:
> One of the possible fix is to put a default body on those injected interfaces.

My main issue with that is those injected interfaces are also to be used by any modders creating multipart entities, and those methods being `default` doesn't make sense in that case, as they very much so need to be implemented.

Sure it'll probably be a quick fix once they realize it, but why permit that in the first place.

It's a tradeoff, but I think the visibility for other multipart entities is more important.

0xJoeMama:
> It's a tradeoff, but I think the visibility for other multipart entities is more important.

To be honest I don't see a lot of people extending EnderDragonEntity.

PR #152, opened 2022-07-16
Item Rendering Extensions

Leo40Git:
A rewritten version of my old [item overlay customization API](https://github.com/FabricMC/fabric/pull/916).

Also adds the containing `rendering` library.

### TODO
- [x] Customizing parts of the overlay
  - [x] Count label
  - [x] Item bar
    - [x] Provide API for multiple item bars
  - [x] Cooldown overlay
- [x] Callbacks for before/after overlay rendered
- [x] Javadoc
- [x] Testmod
  - [x] Count label
  - [x] Item bars
  - [x] Cooldown overlay
  - [x] Before/after render callbacks


williambl:
oooh, remind me to review this later 👀

oliviathevampire:
This looks pretty good! but waiting for other people to review it, you need to update to the latest changes, and I suggets adding a test mod.

library/item/item_rendering/build.gradle
```diff
@@ -0,0 +1,19 @@
...
+	id = "quilt_item_rendering"
+	description = "An API for modifying and extending the rendering of items in inventories."
+	library = "item"
+	moduleDependencies {
+		core {
+			api("qsl_base")
+		}
+	}
+	injectedInterface("net/minecraft/class_1792") {
+		values = ["org/quiltmc/qsl/item/rendering/client/api/QuiltItemRenderingExtensions"]
```
oliviathevampire: this should be ``org/quiltmc/qsl/item/rendering/api/client/QuiltItemRenderingExtensions`` instead of ``org/quiltmc/qsl/item/rendering/client/api/QuiltItemRenderingExtensions``

0xJoeMama:
Can we get BuiltinItemRendererRegistry as well with this PR? Or do you think it's out of scope?

Leo40Git:
> Can we get BuiltinItemRendererRegistry as well with this PR? Or do you think it's out of scope?

I'll do that in a separate PR after this one gets merged.

library/item/item_rendering/src/main/java/org/quiltmc/qsl/item/rendering/mixin/client/ItemRendererMixin.java
```diff
@@ -0,0 +1,99 @@
...
+public abstract class ItemRendererMixin {
+	@Unique private final ThreadLocal<MatrixStack> quilt$matrices = ThreadLocal.withInitial(MatrixStack::new);
+
+	@Shadow public float zOffset;
+
+	/**
+	 * @author QuiltMC
+	 * @reason Completely refactoring overlay rendering logic to use {@link QuiltItemRenderingExtensions}
+	 */
+	@Overwrite
```
williambl: This overwrite is going to cause a lot of incompatibilities with fabric mods (
image
) - is there any other way we can do it?

williambl (changes requested):
I am super happy with the API itself. Tests are great too. I only see one problem, the overwrite, since some fabric mods will break with it.

library/item/item_rendering/src/main/java/org/quiltmc/qsl/item/rendering/mixin/client/ItemRendererMixin.java
Leo40Git: I made it an `@Overwrite` simply because I couldn't be bothered to redo my old many-parts mixin at the time. I'll try to split it up into `@Inject`s and `@Redirect`s.

Leo40Git: ...also, that third one is 100% copied from my old PR! It has the same comments as well.

OroArmor:
You might hate this, but i think this would fit better under the `rendering` library as an `item` module. Since this entire module is client side and rendering related, i think it fits better there.

0xJoeMama:
Agreed ^^^

I only just realised this was under the item library.

Leo40Git:
IT HATH BEEN MOVED

williambl (approved):
Beautiful. Maybe give it a package-info with a high-level overview javadoc? Otherwise, perfect.

oliviathevampire (approved):

library/rendering/item_rendering/src/main/java/org/quiltmc/qsl/rendering/item/api/client/GuiRendererHelper.java
```diff
@@ -0,0 +1,306 @@
...
+import net.fabricmc.api.EnvType;
+import net.fabricmc.api.Environment;
+
+import net.minecraft.client.render.GameRenderer;
+import net.minecraft.client.util.math.MatrixStack;
+
+/**
+ * Contains various helper methods for rendering GUI elements.
+ */
+public interface GuiRendererHelper {
```
0xJoeMama: To be frank, I think this interface should be split into 2 others. One that holds the static methods, which are general utility methods(you could debate this could be moved to another module as well), and another one with the non-static ones.

library/rendering/item_rendering/src/main/java/org/quiltmc/qsl/rendering/item/api/client/CooldownOverlayRenderer.java
```diff
@@ -0,0 +1,47 @@
...
+import net.fabricmc.api.Environment;
+
+import net.minecraft.client.font.TextRenderer;
+import net.minecraft.client.util.math.MatrixStack;
+import net.minecraft.item.ItemStack;
+
+/**
+ * Renders a cooldown overlay.
+ */
+@FunctionalInterface
```
0xJoeMama, line 29: I suppose you could annotate the whole interface with Environment, instead of doing so on its members.

library/rendering/item_rendering/src/main/java/org/quiltmc/qsl/rendering/item/mixin/client/ItemRendererMixin.java
```diff
@@ -0,0 +1,167 @@
...
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
+
+import org.quiltmc.qsl.rendering.item.api.client.CooldownOverlayRenderer;
+import org.quiltmc.qsl.rendering.item.api.client.CountLabelRenderer;
+import org.quiltmc.qsl.rendering.item.api.client.ItemBarRenderer;
+
+@Mixin(ItemRenderer.class)
+public abstract class ItemRendererMixin {
+	@Unique private final ThreadLocal<MatrixStack> quilt$matrices = ThreadLocal.withInitial(MatrixStack::new);
```
0xJoeMama: Do we really need to use ThreadLocal here? I don't think this is accessed by any other thread, other than the Render Thread.

library/rendering/item_rendering/src/main/java/org/quiltmc/qsl/rendering/item/mixin/client/ItemRendererMixin.java
```diff
@@ -0,0 +1,167 @@
...
+		}
+
+		matrices.push();
+		matrices.translate(x, y, 0);
+		stack.getItem().postRenderOverlay(matrices, renderer, this.zOffset, stack);
+		matrices.pop();
+	}
+
+	@Unique
+	private boolean isItemOverlayRenderingCustomized(Item item) {
```
0xJoeMama: Does adding `@Unique` make sure that the method never collides? 
Does it salt the method or something? If it doesn't I suppose you could prefix it with quilt$ to ensure nothing weird happens.

0xJoeMama: I also think this should be in a separate class. Might be useful for people.

library/rendering/item_rendering/src/testmod/java/org/quiltmc/qsl/rendering/item/test/client/cooldown/FlashingCooldownOverlayRenderer.java
```diff
@@ -0,0 +1,52 @@
...
+import net.minecraft.util.Util;
+import net.minecraft.util.math.MathHelper;
+
+import org.quiltmc.qsl.rendering.item.api.client.CooldownOverlayRenderer;
+import org.quiltmc.qsl.rendering.item.api.client.VanillaCooldownOverlayRenderer;
+
+/**
+ * Shows a full red, flashing overlay if there's more than 80% of the cooldown still remaining.
+ */
+public class FlashingCooldownOverlayRenderer extends VanillaCooldownOverlayRenderer {
```
0xJoeMama, line 33: You might want to add an Environment annotation here.

library/rendering/item_rendering/src/testmod/java/org/quiltmc/qsl/rendering/item/test/client/cooldown/HiddenCooldownOverlayRenderer.java
```diff
@@ -0,0 +1,34 @@
...
+
+import net.minecraft.item.ItemStack;
+
+import org.quiltmc.qsl.rendering.item.api.client.CooldownOverlayRenderer;
+import org.quiltmc.qsl.rendering.item.api.client.VanillaCooldownOverlayRenderer;
+
+/**
+ * Hides the cooldown overlay, even if there is a cooldown, as long as it has more than 20% remaining.
+ */
+public class HiddenCooldownOverlayRenderer extends VanillaCooldownOverlayRenderer {
```
0xJoeMama, line 31: Same here.

library/rendering/item_rendering/src/testmod/java/org/quiltmc/qsl/rendering/item/test/client/ItemDecorations.java
```diff
@@ -0,0 +1,68 @@
...
+import net.minecraft.util.Identifier;
+
+import org.quiltmc.qsl.rendering.item.api.client.GuiRendererHelper;
+import org.quiltmc.qsl.rendering.item.test.ItemRenderingTestmod;
+
+@Environment(EnvType.CLIENT)
+public class ItemDecorations {
+	private ItemDecorations() { }
+
+	public static final Identifier STACK_BORDER_TEXTURE = ItemRenderingTestmod.id("textures/gui/border.png");
```
0xJoeMama, line 34: Hm. Are these textures stitched into the item texture Atlas? If not, this may cause some performance issues if a lot of mods follow this trend.

library/rendering/item_rendering/src/main/java/org/quiltmc/qsl/rendering/item/api/client/GuiRendererHelper.java
```diff
@@ -0,0 +1,306 @@
...
+import com.mojang.blaze3d.vertex.VertexFormats;
+import net.fabricmc.api.EnvType;
+import net.fabricmc.api.Environment;
+
+import net.minecraft.client.render.GameRenderer;
+import net.minecraft.client.util.math.MatrixStack;
+
+/**
+ * Contains various helper methods for rendering GUI elements.
+ */
```
0xJoeMama: `@Environment`

library/rendering/item_rendering/src/main/java/org/quiltmc/qsl/rendering/item/api/client/VanillaCountLabelRenderer.java
```diff
@@ -0,0 +1,83 @@
...
+	@Override
+	public void renderCountLabel(MatrixStack matrices, TextRenderer renderer, float zOffset, ItemStack stack, @Nullable String override) {
+		String label = getCountLabel(stack, override);
+		if (label == null) {
+			return;
+		}
+
+		matrices.push();
+		matrices.translate(0, 0, zOffset + 200);
+		// TODO figure out if we can NOT render immediately here
```
0xJoeMama, line 43: I suppose this TODO means you haven't found a way to avoid Immediate right? That's a bit concerning.

0xJoeMama: I guess vanilla uses Immediate here as well right? I'll have to go look, but I think that's the case.

0xJoeMama (commented):
General Note: Mixin handler methods, are salted at runtime, therefore adding quilt$ in the beginning doesn't really offer a lot in terms of compatibility, while also making the code a bit more annoying to read. Correct me if I am wrong.

I'd also like to note that this review doesn't offer much, in terms of getting the PR approved. It's just some comments I could scratch up.

library/rendering/item_rendering/src/main/java/org/quiltmc/qsl/rendering/item/api/client/CooldownOverlayRenderer.java
Leo40Git: This was purposefully done to prevent any possible "class does not exist" shenanigans on dedicated servers - the interface exists on all sides, but is empty outside of clients.

library/rendering/item_rendering/src/main/java/org/quiltmc/qsl/rendering/item/mixin/client/ItemRendererMixin.java
Leo40Git: `@Unique` with private methods 100% makes the method never collide, AFAIK.

library/rendering/item_rendering/src/testmod/java/org/quiltmc/qsl/rendering/item/test/client/ItemDecorations.java
Leo40Git: They are not, I have no idea how to do that

library/rendering/item_rendering/src/main/java/org/quiltmc/qsl/rendering/item/api/client/VanillaCountLabelRenderer.java
Leo40Git: Yeah, this is copied from vanilla.

library/rendering/item_rendering/src/main/java/org/quiltmc/qsl/rendering/item/api/client/CooldownOverlayRenderer.java
0xJoeMama: ~~Wouldn't this be the perfect use case for `@EnvironmentalInterface`?~~
Forget this, seems I misunderstood was EnvironmentInterface does.

library/rendering/item_rendering/src/main/java/org/quiltmc/qsl/rendering/item/api/client/GuiRendererHelper.java
Leo40Git: Honestly I should just make this a proper utility class

library/rendering/item_rendering/src/main/java/org/quiltmc/qsl/rendering/item/mixin/client/ItemRendererMixin.java
0xJoeMama: image
Yeap javadocs confirms that to be the case.

library/rendering/item_rendering/src/main/java/org/quiltmc/qsl/rendering/item/api/client/GuiRendererHelper.java
```diff
@@ -0,0 +1,183 @@
...
+		int b = color & 0xFF;
+		int a = (color >> 24) & 0xFF;
+
+		RenderSystem.setShader(GameRenderer::getPositionColorShader);
+		buffer.begin(VertexFormat.DrawMode.QUADS, VertexFormats.POSITION_COLOR);
+		buffer.vertex(mat, x, y, 0.0f).color(r, g, b, a).next();
+		buffer.vertex(mat, x, y + height, 0.0f).color(r, g, b, a).next();
+		buffer.vertex(mat, x + width, y + height, 0.0f).color(r, g, b, a).next();
+		buffer.vertex(mat, x + width, y, 0.0f).color(r, g, b, a).next();
+		BufferRenderer.drawWithShader(buffer.end());
```
OroArmor: I can hear jellysquid screaming a thousand screams as someone loops this method and submits 10k draw calls, bringing the FPS down to 0. I would remove the draw call and only set the shader if the current shader is different and draw the buffer as it is. Otherwise this is *extremely* inefficient code and bad for rendering.

library/rendering/item_rendering/src/main/java/org/quiltmc/qsl/rendering/item/api/client/QuiltItemRenderingExtensions.java
```diff
@@ -0,0 +1,85 @@
...
+		return CooldownOverlayRenderer.VANILLA;
+	}
+
+	/**
+	 * Gets the item bar renderers associated with this item.
+	 *
+	 * @return array of item bar renderers, or {@code null} to use the vanilla renderer only
+	 */
+	@Environment(EnvType.CLIENT)
+	default ItemBarRenderer @Nullable [] getItemBarRenderers() {
```
OroArmor, line 73: This `@Nullable` annotation is in a weird place. i would do `@Nullable ItemBarRenderer[]`

library/rendering/item_rendering/src/main/java/org/quiltmc/qsl/rendering/item/mixin/client/ItemRendererMixin.java
OroArmor: Its good practice. Other mods might also render twice like Replay mod (Not sure)

library/rendering/item_rendering/src/main/java/org/quiltmc/qsl/rendering/item/mixin/client/ItemRendererMixin.java
```diff
@@ -0,0 +1,158 @@
...
+
+		matrices.push();
+		matrices.translate(x, y, 0);
+		stack.getItem().postRenderOverlay(matrices, renderer, this.zOffset, stack);
+		matrices.pop();
+	}
+
+	@Unique
+	private void renderCustomGuiItemOverlay(MatrixStack matrices,
+										   TextRenderer renderer, ItemStack stack, @Nullable String countLabel) {
```
OroArmor: The line wrapping and formatting on these are a bit weird. I think making your line length longer would fix it. I can't remember if QSL has a limit

library/rendering/item_rendering/src/testmod/java/org/quiltmc/qsl/rendering/item/test/client/ItemDecorations.java
OroArmor: Texture atlases are extremely complex, and the GUI texture usually used is not atlased. This would be quite an invasive change to Minecraft if this were to be fixed. Not sure the best way to fix it though.

OroArmor (changes requested):
Small nitpicks. I wonder if an accesswidener for the model predicate provider registry could be made in this PR as well.

library/rendering/item_rendering/src/main/java/org/quiltmc/qsl/rendering/item/api/client/GuiRendererHelper.java
Leo40Git: Yeah, I _really_ need to figure out how to make this not horrible (AKA sharing a single `BufferBuilder` and drawing it at the end) while also not making it break should someone set a new shader and not set it back.

library/rendering/item_rendering/src/main/java/org/quiltmc/qsl/rendering/item/api/client/QuiltItemRenderingExtensions.java
Leo40Git: `@Nullable ItemBarRenderer[]` -> array itself has no annotation, elements are `@Nullable`
`ItemBarRenderer @Nullable []` -> array itself is `@Nullable`, elements have no annotations (though they're treated as `@NotNull` in the impl)

OroArmor: Ah ok. Looks weird to me but makes sense.

library/rendering/item_rendering/src/main/java/org/quiltmc/qsl/rendering/item/api/client/GuiRendererHelper.java
OroArmor: I would look at the `DrawableHelper` class. I do think it covers most of these methods.

Leo40Git:
I have committed an initial version of the quad batching API (currently untested since I can't run the testmod for some reason).

Leo40Git:
Closing because _man_ I made this overcomplicated as hell. Will bring back later (possibly after extracting the quad batching stuff into its own PR)?

maximumpower55:
Feels like QSL shouldn't be doing this invasive rendering optimizations, quad batching would fit better in its own rendering optimization mod Imo

Leo40Git:
This optimization only applies to items customized via this API - Vanilla items (and items from mods that do _not_ use this API) are unaffected.

maximumpower55:
Ah👍

PR #153, opened 2022-07-20
Fix ModResourcePackUtil mishandling special characters.

LambdAurora:
This is directly inspired from FabricMC/fabric#2407 and QSL suffers the same issue due to the implementation initially being the same.

Said PR explains really well why this is an issue and how it could happen.

Though for implementation into QSL I have chosen a different path that uses `replace` and escape first all backslashes then escape the quotes. In addition I preferred to directly use the `\u<unicode code>` format rather than creating backslash hell.

I also split the method to have the bit that generates the specific string into a separate method, easing testing which is directly done using method calling of a variety of input strings then parsing the resulting JSON.


EnnuiL (approved):
No objections here; PR approved!

OroArmor (approved):

maximumpower55 (approved):
LGTM (yes i did actually review)

Leo40Git (approved):

oliviathevampire (approved):

PR #154, opened 2022-07-24
Maybe?

0xJoeMama:
People asked me to.

This PR adds the 2 following utility classes:
* The Maybe monad, that is basically a better overall implementation of Java's Optional.
* The Lazy class, that allows for lazy initialization of a value(currently not *that* thread-safe, be warned)

### DIFF
--- a/library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Lazy.java
+++ b/library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Lazy.java
@@ -22,7 +22,6 @@ import java.util.function.Supplier;
 
 /**
  * Represents a value that is only initialized once it's needed.
- *
  * <p>
  * Similar to {@link com.google.common.base.Suppliers#memoize}.
  *
@@ -53,7 +52,6 @@ public abstract sealed class Lazy<T> implements Supplier<T> permits Lazy.Filled,
 
 	/**
 	 * Creates a new {@link Lazy} containing the current value mapped using the provided {@link Function}.
-	 *
 	 * <p>
 	 * <i><b>This does not initialize the contained value!</b></i>
 	 *
@@ -66,7 +64,6 @@ public abstract sealed class Lazy<T> implements Supplier<T> permits Lazy.Filled,
 	/**
 	 * If we contain a value, we instantly return the transformed version, using the {@link Function} provided.
 	 * Otherwise, we return a new {@link Lazy} containing the value of the result of the mapper {@link  Function}.
-	 *
 	 * <p>
 	 * <i><b>This does not initialize the contained value!</b></i>
 	 *
@@ -78,7 +75,6 @@ public abstract sealed class Lazy<T> implements Supplier<T> permits Lazy.Filled,
 
 	/**
 	 * Initializes the current {@link Lazy} and returns the result of applying the provided {@link Function} to the value.
-	 *
 	 * <p>
 	 * <i><b>This initializes the contained value.</b></i>
 	 *
@@ -122,7 +118,6 @@ public abstract sealed class Lazy<T> implements Supplier<T> permits Lazy.Filled,
 
 	/**
 	 * Initializes the current value and applies it to the provided {@link Consumer}.
-	 *
 	 * <p>
 	 * <i><b>This initializes the contained value.</b></i>
 	 *
@@ -143,7 +138,6 @@ public abstract sealed class Lazy<T> implements Supplier<T> permits Lazy.Filled,
 
 	/**
 	 * Returns the contained value if it is initialized, otherwise the provided {@link T} instance.
-	 *
 	 * <p>
 	 * <i><b>This does not initialize the contained value.</b></i>
 	 *
@@ -162,7 +156,6 @@ public abstract sealed class Lazy<T> implements Supplier<T> permits Lazy.Filled,
 
 	/**
 	 * A {@link Lazy} with an already initialized value.
-	 *
 	 * <p>
 	 * This class can either be created using the {@link Lazy#filled} method or by mapping a {@link Lazy} with an initialized value.
 	 * A {@link Lazy} cannot be converted by itself into a {@link Filled}.
--- a/library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Lazy.java
+++ b/library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Lazy.java
@@ -21,62 +21,70 @@ import java.util.function.Function;
 import java.util.function.Supplier;
 
 /**
- * Represents a value that is only initialized once it's needed.<br/>
+ * Represents a value that is only initialized once it's needed.
+ *
+ * <p>
  * Similar to {@link com.google.common.base.Suppliers#memoize}.
  *
- * @param <T> The type of the value this {@link Lazy} represents.
+ * @param <T> the type of the value this {@link Lazy} represents
  */
 public abstract sealed class Lazy<T> implements Supplier<T> permits Lazy.Filled, Lazy.OfSupplier {
 	/**
 	 * Wraps the provided {@link Supplier} into a {@link OfSupplier} instance.
 	 *
-	 * @param supplier The {@link Supplier} to a value.
-	 * @param <T>      The type of the wrapped value.
-	 * @return A new {@link OfSupplier} instance.
+	 * @param supplier the {@link Supplier} to a value
+	 * @param <T>      The type of the wrapped value
+	 * @return a new {@link OfSupplier} instance
 	 */
-	public static <T> Lazy<T> of(Supplier<? extends T> supplier) {
+	public static <T> Lazy.OfSupplier<T> of(Supplier<? extends T> supplier) {
 		return new OfSupplier<>(supplier);
 	}
 
 	/**
 	 * Returns an {@link Filled} instance wrapping the provided {@link T} value.
 	 *
-	 * @param value The value to wrap.
-	 * @param <T>   The type of the wrapped value.
-	 * @return A new {@link Filled} instance.
+	 * @param value the value to wrap
+	 * @param <T>   The type of the wrapped value
+	 * @return a new {@link Filled} instance
 	 */
-	public static <T> Lazy<T> filled(T value) {
+	public static <T> Lazy.Filled<T> filled(T value) {
 		return new Filled<>(value);
 	}
 
 	/**
-	 * Creates a new {@link Lazy} containing the current value mapped using the provided {@link Function}.<br/>
+	 * Creates a new {@link Lazy} containing the current value mapped using the provided {@link Function}.
+	 *
+	 * <p>
 	 * <i><b>This does not initialize the contained value!</b></i>
 	 *
-	 * @param mapper The function to apply to the contained value.
-	 * @param <U>    The new contained type.
-	 * @return A new {@link Lazy} instance, that may or may not be initialized that contains the mapped value.
+	 * @param mapper the function to apply to the contained value
+	 * @param <U>    The new contained type
+	 * @return a new {@link Lazy} instance, that may or may not be initialized that contains the mapped value
 	 */
 	public abstract <U> Lazy<U> map(Function<T, ? extends U> mapper);
 
 	/**
-	 * If we contain a value, we instantly return the transformed version, using the {@link Function} provided.<br/>
-	 * Otherwise, we return a new {@link Lazy} containing the value of the result of the mapper {@link  Function}.<br/>
+	 * If we contain a value, we instantly return the transformed version, using the {@link Function} provided.
+	 * Otherwise, we return a new {@link Lazy} containing the value of the result of the mapper {@link  Function}.
+	 *
+	 * <p>
 	 * <i><b>This does not initialize the contained value!</b></i>
 	 *
-	 * @param mapper The {@link Function} to use.
-	 * @param <U>    The new contained type.
-	 * @return The mapped {@link Lazy} value.
+	 * @param mapper the {@link Function} to use
+	 * @param <U>    The new contained type
+	 * @return the mapped {@link Lazy} value
 	 */
 	public abstract <U> Lazy<U> flatMap(Function<T, ? extends Lazy<U>> mapper);
 
 	/**
-	 * Initializes the current {@link Lazy} and returns the result of applying the provided {@link Function} to the value.<br/>
+	 * Initializes the current {@link Lazy} and returns the result of applying the provided {@link Function} to the value.
+	 *
+	 * <p>
 	 * <i><b>This initializes the contained value.</b></i>
 	 *
-	 * @param function The {@link Function} to use.
-	 * @param <M>      The return type.
-	 * @return The {@link M} value produced from the application of the provided {@link Function}}.
+	 * @param function the {@link Function} to use
+	 * @param <M>      The return type
+	 * @return the {@link M} value produced from the application of the provided {@link Function}}
 	 */
 	public <M> M mapUnwrap(Function<T, ? extends M> function) {
 		return function.apply(this.unwrap());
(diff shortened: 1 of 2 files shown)


library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Lazy.java
```diff
@@ -21,62 +21,70 @@
...
+	 * @param mapper the {@link Function} to use
+	 * @param <U>    The new contained type
+	 * @return the mapped {@link Lazy} value
 	 */
 	public abstract <U> Lazy<U> flatMap(Function<T, ? extends Lazy<U>> mapper);
 
 	/**
-	 * Initializes the current {@link Lazy} and returns the result of applying the provided {@link Function} to the value.<br/>
+	 * Initializes the current {@link Lazy} and returns the result of applying the provided {@link Function} to the value.
+	 *
```
LambdAurora:
```suggestion
```

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Lazy.java
```diff
@@ -21,62 +21,70 @@
...
+	 * @return a new {@link Lazy} instance, that may or may not be initialized that contains the mapped value
 	 */
 	public abstract <U> Lazy<U> map(Function<T, ? extends U> mapper);
 
 	/**
-	 * If we contain a value, we instantly return the transformed version, using the {@link Function} provided.<br/>
-	 * Otherwise, we return a new {@link Lazy} containing the value of the result of the mapper {@link  Function}.<br/>
+	 * If we contain a value, we instantly return the transformed version, using the {@link Function} provided.
+	 * Otherwise, we return a new {@link Lazy} containing the value of the result of the mapper {@link  Function}.
+	 *
```
LambdAurora:
```suggestion
```

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Lazy.java
```diff
@@ -21,62 +21,70 @@
...
 	 */
-	public static <T> Lazy<T> filled(T value) {
+	public static <T> Lazy.Filled<T> filled(T value) {
 		return new Filled<>(value);
 	}
 
 	/**
-	 * Creates a new {@link Lazy} containing the current value mapped using the provided {@link Function}.<br/>
+	 * Creates a new {@link Lazy} containing the current value mapped using the provided {@link Function}.
+	 *
```
LambdAurora:
```suggestion
```

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Lazy.java
```diff
@@ -21,62 +21,70 @@
 import java.util.function.Supplier;
 
 /**
- * Represents a value that is only initialized once it's needed.<br/>
+ * Represents a value that is only initialized once it's needed.
+ *
```
LambdAurora:
```suggestion
```

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Lazy.java
```diff
@@ -85,39 +93,41 @@ public <M> M mapUnwrap(Function<T, ? extends M> function) {
...
 	 *
-	 * @return Whether there is a value in this {@link Lazy} that is initialized.
+	 * @return {@code true} if the current value isn't initialized, or {@code false} otherwise
 	 */
 	public abstract boolean isEmpty();
 
 	/**
-	 * Initializes the current value and applies it to the provided {@link Consumer}.<br/>
+	 * Initializes the current value and applies it to the provided {@link Consumer}.
+	 *
```
LambdAurora:
```suggestion
```

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Lazy.java
```diff
@@ -133,35 +143,37 @@ public T unwrap() {
 
 	/**
 	 * Returns the contained value if it is initialized, otherwise the provided {@link T} instance.
+	 *
```
LambdAurora:
```suggestion
```

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Lazy.java
```diff
@@ -133,35 +143,37 @@ public T unwrap() {
...
-	 * @return Either the contained value or the result of {@link Supplier#get} when called on the provided {@link Supplier}.
+	 * @param defaultSupplier a provider of a {@link T} instance
+	 * @return either the contained value or the result of {@link Supplier#get} when called on the provided {@link Supplier}
 	 */
 	public abstract T unwrapOrGet(Supplier<? extends T> defaultSupplier);
 
 	/**
-	 * A {@link Lazy} with an already initialized value.<br/>
+	 * A {@link Lazy} with an already initialized value.
 	 *
```
LambdAurora:
```suggestion
```

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Maybe.java
```diff
@@ -16,97 +16,104 @@
...
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 
+import org.jetbrains.annotations.Nullable;
+
 /**
- * Represents a monad that can either contain a value or no value.<br/>
+ * Represents a monad that can either contain a value or no value.
+ *
```
LambdAurora:
```suggestion
```

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Maybe.java
```diff
@@ -16,97 +16,104 @@
...
+ * <p>
  * Similar to the {@link Optional} class but with some performance improvements
- * and quality of life changes.<br/>
+ * and quality of life changes.
  *
```
LambdAurora:
```suggestion
```

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Maybe.java
```diff
@@ -16,97 +16,104 @@
...
  * <p>
- * Can either be {@link Just} or {@link Nothing}.<br/>
- * {@link Just} instances contain a value of type {@link T}.<br/>
- * {@link Nothing} instances all contain no data.<br/>
+ * Can either be {@link Just} or {@link Nothing}.
  *
```
LambdAurora:
```suggestion
```

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Maybe.java
```diff
@@ -16,97 +16,104 @@
...
- * @param <T> The type of the wrapped object
+ * <p>
+ * {@link Just} instances contain a value of type {@link T}.
+ *
```
LambdAurora:
```suggestion
```

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Maybe.java
```diff
@@ -16,97 +16,104 @@
...
 	 * Wraps a {@link T} value into an {@link Maybe}.
 	 *
-	 * @param value A {@link Nullable} value that will be wrapped into a {@link Maybe}.
-	 * @param <T>   The type of object the {@link Maybe} instance will hold.
-	 * @return An instance of {@link Just} if the provided value is not <code>null</code>,<br/>
-	 * otherwise a {@link Nothing} instance.
+	 * @param value a {@link Nullable} value that will be wrapped into a {@link Maybe}
+	 * @param <T>   the type of object the {@link Maybe} instance will hold
+	 * @return an instance of {@link Just} if the provided value is not <code>null</code>,
+	 * otherwise a {@link Nothing} instance
```
LambdAurora:
```suggestion
	 * or a {@link Nothing} instance otherwise
```

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Maybe.java
```diff
@@ -255,9 +262,9 @@ public static <T> Maybe<T> fromOptional(Optional<T> opt) {
 	 * Works like {@link Maybe#unwrap()} except that it doesn't throw if the current instance if {@link Nothing}.
 	 * In that case it returns the provided {@link T} value.
 	 *
-	 * @param defaultValue The value to get if the current instance if {@link Nothing}.
-	 * @return The current {@link T} value if the current instance is {@link Just}, otherwise the provided {@link T}
-	 * value.
+	 * @param defaultValue the value to get if the current instance if {@link Nothing}
+	 * @return the current {@link T} value if the current instance is {@link Just}, otherwise the provided {@link T}
+	 * value
```
LambdAurora:
```suggestion
	 * @return the current {@link T} value if the current instance is {@link Just}, or the provided {@link T}
	 * value otherwise
```

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/util/Maybe.java
```diff
@@ -266,46 +273,46 @@ public static <T> Maybe<T> fromOptional(Optional<T> opt) {
...
 	 * {@link Supplier}.
 	 *
-	 * @param supplier A {@link Supplier} that will create the {@link T} value to be returns if the current instance
-	 *                    is {@link Nothing}.
-	 * @return The current {@link T} value if the current instance if {@link Just},
-	 * otherwise the value produces by calling {@link Supplier#get()} on the provided {@link Supplier}.
+	 * @param supplier a {@link Supplier} that will create the {@link T} value to be returns if the current instance
+	 *                 is {@link Nothing}
+	 * @return the current {@link T} value if the current instance if {@link Just},
+	 * otherwise the value produces by calling {@link Supplier#get()} on the provided {@link Supplier}
```
LambdAurora:
```suggestion
	 * or the value produces by calling {@link Supplier#get()} on the provided {@link Supplier} otherwise
```

LambdAurora (commented):
We don't have empty lines before a `<p>` line usually in QSL.

TheGlitch76:
This PR is fine but I don't see the purpose behind it. Why does QSL need these utils?

oliviathevampire:
> This PR is fine but I don't see the purpose behind it. Why does QSL need these utils?

If I remember correctly these are used in the component api

0xJoeMama:
> If I remember correctly these are used in the component api

Also, there are quite a lot of performance and overall improvements both functionality-wise and performance-wise.

lukebemish:
It might be worth documenting on `Maybe` what the differences from `Optional` are; currently, it's a bit hard to tell why one might want to use this in place of `Optional`

kb-1000:
As far as I understand with my little JVM knowledge, this cannot be any "better than Optional" performance-wise.
Optional is a final class with a final, nullable field, and single implementations of all its methods.
Having an abstract class with abstract methods and 2 implementations can only be slower... do you have any proof for it being better?
The Optional -> Maybe helper also looks very suspicious.

That aside... avoiding to use Optional (or Maybe, or whatever you want to replace it with) may look a bit worse for APIs... but does that overweigh the performance benefits of using simple null/non null values instead?

On Lazy: multiple libraries Minecraft depends on provide facilities that lazily compute a value when required, why would one use this over those?

TelepathicGrunt:
> > If I remember correctly these are used in the component api
> 
> Also, there are quite a lot of performance and overall improvements both functionality-wise and performance-wise.

I would like to see actual numbers and tests ran to prove that as that doesn’t seem right. There are other options besides reimplementing optional. This is definitely a case of take a step back and look over the whole goal again as some people have stated this PR is basically one big code smell which may be indication of a bigger underlying issue that this pr is masking.

0xJoeMama:
> one big code smell which may be indication of a bigger underlying issue that this pr is masking.

So I take it this is targeting the component API.

Firstly, I can assure you that the component API doesn't *require* the Maybe class per se. I had a discussion in Discord with Lambda and Platymemo about this. I just made the Maybe class for another project I was working on, and after using it I thought that this would be cool to work with on other projects as well. I asked and people seemed to not have a big problem with it.

Component API was using Optional in the beginning as well. It doesn't require Maybe nor is Maybe hiding an *underlying* component API issue(whatever that might entail or refer to).

I feel a bit attacked by this, but I guess its not that big a deal. It's only natural for people to be suspicious of things like this.

> I would like to see actual numbers and tests ran to prove that as that doesn’t seem right. 

I am in the middle of running some tests right now. I will post the results once I am done. 

## Finally
I would like to say that whether Maybe and Lazy make it into the qsl or not, isn't that big a deal for me.
 If  people don't approve of them, I am just fine using Optional and Suppliers.memoize for everything.

alcatrazEscapee:
Since nobody seems to have yet benchmarked this properly before making performance claims, I did. The benchmark code is [here](https://github.com/alcatrazEscapee/performance-tests/blob/main/src/jmh/java/maybe/MaybeVsOptionalBenchmark.java), uses JMH on Java 17, feel free to try and replicate my results.

Bottom line: `Maybe` is just as fast as `Optional`, within the margin of error and in a few cases very slightly slower. There is **zero evidence** that `Maybe` is faster in any tangible way.

Results:

```
Benchmark                                                 Mode  Cnt   Score   Error  Units
MaybeVsOptionalBenchmark.benchmarkOptionalNew             avgt   15  25.028 ± 1.066  ns/op
MaybeVsOptionalBenchmark.benchmarkMaybeNew                avgt   15  29.014 ± 3.828  ns/op

MaybeVsOptionalBenchmark.benchmarkOptionalIsPresent       avgt   15  24.688 ± 0.767  ns/op
MaybeVsOptionalBenchmark.benchmarkMaybeIsPresent          avgt   15  27.994 ± 4.216  ns/op

MaybeVsOptionalBenchmark.benchmarkOptionalOrElse          avgt   15  24.883 ± 0.797  ns/op
MaybeVsOptionalBenchmark.benchmarkMaybeOrElse             avgt   15  24.838 ± 0.935  ns/op

MaybeVsOptionalBenchmark.benchmarkOptionalOr              avgt   15  24.803 ± 0.647  ns/op
MaybeVsOptionalBenchmark.benchmarkMaybeOr                 avgt   15  24.569 ± 0.126  ns/op

MaybeVsOptionalBenchmark.benchmarkOptionalMapToString     avgt   15  67.577 ± 1.989  ns/op
MaybeVsOptionalBenchmark.benchmarkMaybeMapToString        avgt   15  67.084 ± 0.307  ns/op

MaybeVsOptionalBenchmark.benchmarkOptionalFilterHashCode  avgt   15  40.836 ± 0.501  ns/op
MaybeVsOptionalBenchmark.benchmarkMaybeFilterHashCode     avgt   15  40.831 ± 0.194  ns/op
```

lukebemish:
My take on this is that, with that benchmark in mind, `Maybe` isn't particularly valuable; whether the components system should use optionals or just pass null values is another question, that I won't comment on without looking at those more closely. In terms of `Lazy`, I can see the value it has as opposed to `Suppliers.memoize`; after all, if you get a `Lazy`, you can be sure that it is lazy and not have to manually memoize it as you would a supplier. However, calling `memoize` on an already memoized supplier shouldn't have an appreciable performance impact in most cases (I haven't actually benchmarked this, so I won't claim more than that).

I believe, though am not certain, that the potential "underlying" issue in the components API being referred to was the use of optionals at all, as opposed to passing null values directly. (i.e, the difference between having a `NotNull` optional that is either empty or not, or just passing a `Nullable` value straight away), though I'm not certain myself on the costs/benefits of either approach, or whether this is the right place to discuss that.

0xJoeMama:
Well, it seems like a decision has been reached. I will therefore close this PR and move components over to Optional or a Nullable method return value. Thanks everyone for your contributions. I am sorry this didn't happen earlier but I was on holidays this past week.

PR #155, opened 2022-07-25
Better ordering of resource reloaders

LambdAurora:
Ok, big PR.

## Better ordering API

The goal of this PR was to allow re-ordering of the resource reloaders in a more flexible manner, especially by supporting to make another reloader depend on yours.
This is useful as some people want to fire their reloaders before the Vanilla reloaders.

So to achieve this a new method has been added to `ResourceLoader` called `addPhaseOrdering(first, second)`. We also deprecate the dependecies declarations as those are unneeded and unnecessary complexity.

## Event Phase Ordering more like Public API

To achieve the new sorter a short idea came, what about just recycling the existing sorter used for event phases?
So I made that API, but in a way it can be extended for other uses (as you can see in this PR).

## Uh... care to explain what in hell happened in the resource reloaders mixins?

Uh.... during my testing I might have realized that Fabric and Quilt *might* have omitted **a lot** of injections, causing a lot of resource reloaders to be unnamed. Oops.

## Questions needing answer.

As you can see in `ResourceLoaderImpl` there's something odd, we name non `IdentifiableResourceReloaders`, I'm not a big fan of it and the question is should we log a warning for those that don't implement it?

## Conclusion

I hope y'all like it.


library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceLoaderImpl.java
```diff
@@ -108,6 +119,21 @@ public void registerReloader(IdentifiableResourceReloader resourceReloader) {
 							+ " already in resource reloader set!"
 			);
 		}
+
+		// Keep this for compatibility.
+		//noinspection removal
```
Leo40Git: > IDEA-only warning suppression comment

hiss

Leo40Git (approved):
LGTM, except for this one tiny nitpick

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceLoaderImpl.java
LambdAurora: Replaced :p

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceReloaderPhaseData.java
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	static class AfterVanilla extends ResourceReloaderPhaseData {
+		AfterVanilla(Identifier id) {
+			super(id, null);
+			this.setVanillaStatus(VanillaStatus.VANILLA);
+		}
+
+		@Override
+		public void markBefore() {}
```
EnnuiL, line 102: I was gonna say "`{ }`"!!!" But actually, I prefer `{}`, it's just that one annoying Checkstyle rule

EnnuiL (approved):

oliviathevampire (approved):

Leo40Git (approved):
These annotations activate my neurons

LGTM!

PepperCode1 (approved):

Ecorous (approved):
LGTM!

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/reloader/ResourceReloaderKeys.java
```diff
@@ -25,15 +25,44 @@
  * @see IdentifiableResourceReloader
  */
 public final class ResourceReloaderKeys {
+	/**
+	 * Represents the before Vanilla resource reloaders application phase.
```
Leo40Git:
```suggestion
	 * Represents the application phase before Vanilla resource reloaders are invoked.
```

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/reloader/ResourceReloaderKeys.java
```diff
@@ -25,15 +25,44 @@
...
+	 * <p>
+	 * No resource reloaders are assigned to this identifier.
+	 *
+	 * @see org.quiltmc.qsl.resource.loader.api.ResourceLoader#addReloaderOrdering(Identifier, Identifier)
+	 */
+	public static final Identifier BEFORE_VANILLA = new Identifier("quilt", "before_vanilla");
+	/**
+	 * Represents the after Vanilla resource reloaders application phase.
```
Leo40Git:
```suggestion
	 * Represents the application phase after Vanilla resource reloaders are invoked.
```

Leo40Git (changes requested):
OOPS WAIT almost missed these javadocs!

Leo40Git (approved):
Okay, _now_ it LGTM!

PR #164, opened 2022-08-09
Expose Rendering for EntityParts

Platymemo:
As per:
image

Changes the responsibility of rendering of entity parts to themselves, so they can modify it as necessary.


Arathain (approved):
:+1:

OroArmor (approved):

maximumpower55 (approved):
LGTM lets get this merged.

library/entity/multipart/src/main/java/org/quiltmc/qsl/entity/multipart/api/EntityPart.java
```diff
@@ -30,4 +38,29 @@
...
+	 * @param matrices the {@link MatrixStack matrix stack} used for rendering
+	 * @param vertices the {@link VertexConsumer vertex consumer} used for rendering
+	 * @param ownerX the {@link #getOwner() owner's} rendered X coordinate
+	 * @param ownerY the {@link #getOwner() owner's} rendered Y coordinate
+	 * @param ownerZ the {@link #getOwner() owner's} rendered Z coordinate
+	 * @param owner the {@link #getOwner() owner}
+	 * @param tickDelta progress for linearly interpolating between the previous and current game state
+	 */
+	@Environment(EnvType.CLIENT)
+	default void renderHitbox(MatrixStack matrices, VertexConsumer vertices, double ownerX, double ownerY, double ownerZ, Entity owner, float tickDelta) {
```
LambdAurora, line 55: Oh damn, a client-only method in a normal class?

We would scare some people with that :P
I'm not sure what other option we have though. Definitely means we won't be able to split QSL into common/client source sets, not that we were planning to.
Will take a bit to think about it, but don't take my comment as blocking.

EnnuiL (approved):

williambl (approved):
Seems like the best solution available.

library/entity/multipart/src/main/java/org/quiltmc/qsl/entity/multipart/api/EntityPart.java
Platymemo: I think this is the best solution. We could create some kind of optional EntityPartRenderingRegistry, but I think this method is best for discoverability and ease of use.

Arathain:
any progress on this / anything blocking this entering FCP?

LambdAurora (approved):

PR #167, opened 2022-08-16
Fix DataBreaker compatibility

Leo40Git:
...by slightly over-complicating the DataFixerUpper module's internals.

There are now 2 implementations of `DataFixerUpperInternals`:
1. `DataFixerUpperInternalsImpl`, which contains the code the module used until now (including the code that causes a crash with DataBreaker),
2. `NopDataFixerUpperInternals`, which as the name implies is a no-op implementation (that still complies with the contracts) that is used if DataBreaker is installed.

### DIFF
--- a/library/misc/datafixerupper/src/main/java/org/quiltmc/qsl/datafixerupper/impl/QuiltDataFixesInternals.java
+++ b/library/misc/datafixerupper/src/main/java/org/quiltmc/qsl/datafixerupper/impl/QuiltDataFixesInternals.java
@@ -16,14 +16,15 @@
 
 package org.quiltmc.qsl.datafixerupper.impl;
 
+import com.mojang.datafixers.DataFixUtils;
 import com.mojang.datafixers.DataFixer;
 import com.mojang.datafixers.schemas.Schema;
 import org.jetbrains.annotations.*;
 
+import net.minecraft.SharedConstants;
 import net.minecraft.datafixer.DataFixTypes;
+import net.minecraft.datafixer.Schemas;
 import net.minecraft.nbt.NbtCompound;
-
-import org.quiltmc.loader.api.QuiltLoader;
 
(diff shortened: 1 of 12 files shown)


library/misc/datafixerupper/src/main/java/org/quiltmc/qsl/datafixerupper/impl/QuiltDataFixesInternals.java
```diff
@@ -16,100 +16,56 @@
...
+	private static QuiltDataFixesInternals instance;
 
-	public static void registerFixer(@NotNull String modId,
-									 @Range(from = 0, to = Integer.MAX_VALUE) int currentVersion,
-									 @NotNull DataFixer dataFixer) {
-		if (modDataFixers.containsKey(modId)) {
-			throw new IllegalArgumentException("Mod '" + modId + "' already has a registered data fixer");
+	public static @NotNull QuiltDataFixesInternals get() {
+		if (instance == null) {
+			if (QuiltLoader.isModLoaded("databreaker")) {
```
LambdAurora: This doesn't feel like a good fix.
I'd rather avoid any modid hardcoding.

You said on Discord that with databreaker the Vanilla schemas would be null, I'd rather check that instead.

LambdAurora (changes requested):

library/misc/datafixerupper/src/main/java/org/quiltmc/qsl/datafixerupper/impl/QuiltDataFixesInternals.java
Leo40Git: Hm, that'd probably be better, yeah.

LambdAurora (approved):

EnnuiL (approved):

PR #172, opened 2022-08-20
Enchanting Boosters Content Registry

Kroppeb:
Adds an REA to register blocks that provide a certain amount of enchantment power increase in "Bookshelf Equivalents"


OroArmor (changes requested):
This could probably go under the block content registry, but the current api looks great

SilverAndro:
Im not a fan of `enchantment_stuff`/`enchantmentstuff` as a package name, there must be a better name than that, no?
(Also is there a reason it isnt consistent?)

OroArmor:
I believe that is temporary

Kroppeb:
> Im not a fan of `enchantment_stuff`/`enchantmentstuff` as a package name, there must be a better name than that, no? (Also is there a reason it isnt consistent?)

I was not gonna use it as an actual package name, I just wanted to actually implement it and not spend an hour bikeshedding the name and whether it's supposed to go in the existing block content registry module

library/block/enchantment_stuff/src/main/java/org/quiltmc/qsl/block/enchantmentstuff/api/EnchantmentPowerAttachment.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class EnchantmentPowerAttachment {
+	/**
+	 * A {@link RegistryEntryAttachment} for enchantment power levels in bookshelf equivalents.
+	 * <p>
+	 * Values can be set via code and through a data-pack with the file {@code data/quilt_block_enchantment_stuff/attachments/minecraft/block/power.json}
+	 */
+	public static final RegistryEntryAttachment<Block, Float> POWER_LEVEL = RegistryEntryAttachment
+			.builder(Registry.BLOCK,
+					new Identifier("quilt_block_enchantment_stuff", "power"),
```
Leo40Git:
```suggestion
					new Identifier("quilt_block_enchantment_stuff", "enchanting_power"),
```

library/block/enchantment_stuff/src/main/java/org/quiltmc/qsl/block/enchantmentstuff/impl/BlockEnchantmentStuffInitializer.java
```diff
@@ -0,0 +1,29 @@
...
+
+import net.minecraft.block.Blocks;
+import org.quiltmc.loader.api.ModContainer;
+import org.quiltmc.qsl.base.api.entrypoint.ModInitializer;
+import org.quiltmc.qsl.block.enchantmentstuff.api.EnchantmentPowerAttachment;
+
+public class BlockEnchantmentStuffInitializer implements ModInitializer {
+	@Override
+	public void onInitialize(ModContainer mod) {
+		EnchantmentPowerAttachment.POWER_LEVEL.put(Blocks.BOOKSHELF, 1f);
```
Leo40Git: It might be better to put this in a static initializer in `EnchantmentPowerAttachment`, so mods can properly override the built-in value without having to worry about load order shenanigans.

Maybe I should add a `with(K, V)` method to the attachment builder...

Kroppeb: I thought you weren't supposed to access the `Blocks` class before the initializer?

Leo40Git: Hm, seems like `block_content_registry` also waits for the initializer to copy the vanilla entries into the attachment. This is probably fine for now then, I'll look into engineering a solution later.

OroArmor (approved):

library/block/block_content_registry/src/main/java/org/quiltmc/qsl/block/content/registry/mixin/EnchantingTableBlockMixin.java
```diff
@@ -0,0 +1,60 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+import java.util.Optional;
+
+@Mixin(EnchantingTableBlock.class)
+public class EnchantingTableBlockMixin {
+	@Redirect(method = "isValidForBookshelf", at = @At(value = "INVOKE", target = "Lnet/minecraft/block/BlockState;isOf(Lnet/minecraft/block/Block;)Z"))
+	private static boolean hasEnchantmentPower(BlockState blockState, Block ignored) {
+		return BlockContentRegistries.ENCHANTMENT_POWER_LEVEL.get(blockState.getBlock()).orElse(0f) != 0f;
```
Leo40Git:
```suggestion
		return BlockContentRegistries.ENCHANTMENT_POWER_LEVEL.get(blockState.getBlock()).isPresent();
```

Leo40Git (approved):
LGTM!

library/block/block_content_registry/src/main/java/org/quiltmc/qsl/block/content/registry/mixin/EnchantingTableBlockMixin.java
Leo40Git: Wait nevermind, this lets you disable bookshelf equivalents.

LambdAurora (changes requested):
I love this PR, it expands way beyond any other implementation I know that only use a tag without any wiggle room!

I'm just a bit unsure about the name of the REA.
It doesn't strike as me as an enchantment table/bookshelf related thing at first when reading the name.

Kroppeb:
Forge uses a method called `getEnchantPowerBonus`, so I initially used `enchantment_power` but ended up adding "level".
I wouldn't mind a different name, but I can't come up with anything good.

williambl:
Maybe `enchanting_power`? `enchanting` rather than `enchantment` since it affects the enchanting process, rather than any enchantment directly. The `_level` bit feels unnecessary.

LemmaEOF:
I called them `enchantment_boosters` back in VMulti on 1.16, which is what I created [MCDict](https://github.com/cottonmc/mcdict), the precursor to REAs, for.

EnnuiL:
Ohh, boosters are a good name; After all, this is a block REA and so, having the registry attachment refer to the blocks themselves would be better;
Maybe `enchanting_boosters` would be perfect, combining that with the feedback about using "enchanting" instead

williambl:
I like `enchanting_boosters`.

LemmaEOF:
`enchanting_boosters` sounds great!

Leo40Git:
Another vote towards `enchanting_boosters` from me.

Kroppeb:
I renamed it to `enchanting_booster` given that all other block attachments seems to be singular.

EnnuiL:
Actually, what we really should have done is adopt something similar to tag conventions, making this `quilt:enchanting_boosters`; Unfortunately, we didn't, so oh well, we'll stick with that for 1.19

EnnuiL (changes requested):
In a discussion at the Discord server, we decided that moving towards the new naming scheme (`quilt:enchantment_boosters`) should already begin by having new REAs use it; So, yeah, move to it, and we should finally be good

EnnuiL (approved):

library/block/block_content_registry/src/main/java/org/quiltmc/qsl/block/content/registry/mixin/EnchantmentScreenHandlerMixin.java
```diff
@@ -0,0 +1,63 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.ModifyVariable;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+import java.util.Iterator;
+import java.util.List;
+
+@Mixin(EnchantmentScreenHandler.class)
+public class EnchantmentScreenHandlerMixin {
+	@Redirect(method = "m_mpsetdhw", at = @At(value = "INVOKE", target = "Ljava/util/List;iterator()Ljava/util/Iterator;"))
```
EnnuiL: Due to the migration to Intermediary QM, you'll want to change these two in order to not crash in-prod:
```suggestion
	@Redirect(method = "m_mpsetdhw(Lnet/minecraft/item/ItemStack;Lnet/minecraft/world/World;Lnet/minecraft/util/math/BlockPos;)V", at = @At(value = "INVOKE", target = "Ljava/util/List;iterator()Ljava/util/Iterator;"))
```

library/block/block_content_registry/src/main/java/org/quiltmc/qsl/block/content/registry/mixin/EnchantmentScreenHandlerMixin.java
```diff
@@ -0,0 +1,63 @@
...
+	private Iterator<?> iterator(List<?> instance) {
+		// Cancel old loop
+		return ObjectIterators.emptyIterator();
+	}
+
+	@ModifyVariable(method = "m_mpsetdhw", at = @At(value = "INVOKE", target = "Lnet/minecraft/util/random/RandomGenerator;setSeed(J)V"))
```
EnnuiL:
```suggestion
	@ModifyVariable(method = "m_mpsetdhw(Lnet/minecraft/item/ItemStack;Lnet/minecraft/world/World;Lnet/minecraft/util/math/BlockPos;)V", at = @At(value = "INVOKE", target = "Lnet/minecraft/util/random/RandomGenerator;setSeed(J)V"))
```

EnnuiL (changes requested):
It looks like the Intemediary QM migration broke stuff on this PR; But don't worry! These should fix stuff:

EnnuiL (approved):
Excellent!

EnnuiL:
This has been waiting for way too long; I'm starting the FCP; I still have nitpicks that are important to be addressed, but I decided to do that after merge; after way too much poking, now i'm afraid of annoying more...

library/block/block_content_registry/src/main/java/org/quiltmc/qsl/block/content/registry/api/BlockContentRegistries.java
```diff
@@ -129,5 +131,17 @@ public class BlockContentRegistries {
...
+	/**
+	 * A {@link RegistryEntryAttachment} for enchantment power levels in bookshelf equivalents.
+	 * <p>
+	 * Values can be set via code and through a data-pack with the file {@code data/quilt/attachments/minecraft/block/enchantment_boosters.json}
+	 */
+	public static final RegistryEntryAttachment<Block, Float> ENCHANTMENT_BOOSTERS = RegistryEntryAttachment
+			.builder(Registry.BLOCK,
+					new Identifier(QUILT_NAMESPACE, "enchantment_boosters"),
+					Float.class,
+					Codec.floatRange(0f, Float.MAX_VALUE))
```
xanderstuff, line 144: I'm not entirely familiar with how this works, but the last parameter to ENCHANTMENT_BOOSTERS.put(...) would be the amount of equivalent bookshelves to add, correct? And does this line restrict that amount to positive values only?

If so, it might be nice to allow negative values as well. As for why someone might want to ***reduce*** the power level of the enchantment table, that would require some creativity from the modder, but I think it could have valid uses. 

For instance, maybe someone might add a special enchanting block that has both benefits and drawbacks, like something that enables mending as a possible enchantment but reduces the enchanting level by 8 (encouraging you to compensate by adding even more and/or higher tiered "bookshelves"). Or perhaps another mod might overhaul enchanting such that having low enchanting power is useful as well. Or... maybe even something that gives different enchants for positive vs negative enchanting power (like a white arts/dark arts dichotomy). Although I suppose for the last 2 examples it might be more reasonable to override this API, but the first example still stands.

PR #176, opened 2022-08-21
Add the ability to set boolean block settings values directly

Kroppeb:
Allows for stuff like `QuiltBlockSettings.copyOf(Blocks.VINE).ticksRandomly(false)`

### DIFF
--- a/library/block/block_extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/QuiltBlockSettings.java
+++ b/library/block/block_extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/QuiltBlockSettings.java
@@ -56,21 +56,21 @@ public class QuiltBlockSettings extends AbstractBlock.Settings {
 		this.hardness(otherAccessor.getHardness());
 		this.resistance(otherAccessor.getResistance());
 		this.collidable(otherAccessor.getCollidable());
-		thisAccessor.setRandomTicks(otherAccessor.getRandomTicks());
+		this.ticksRandomly(otherAccessor.getRandomTicks());
 		this.luminance(otherAccessor.getLuminance());
-		thisAccessor.setMapColorProvider(otherAccessor.getMapColorProvider());
+		this.mapColorProvider(otherAccessor.getMapColorProvider());
 		this.sounds(otherAccessor.getSoundGroup());
 		this.slipperiness(otherAccessor.getSlipperiness());
 		this.velocityMultiplier(otherAccessor.getVelocityMultiplier());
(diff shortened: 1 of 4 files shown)


library/block/block_extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/QuiltBlockSettings.java
```diff
@@ -265,6 +265,31 @@ public QuiltBlockSettings collidable(boolean collidable) {
...
+		((AbstractBlockSettingsAccessor) this).setDynamicBounds(dynamicBounds);
+		return this;
+	}
+
+	public QuiltBlockSettings requiresTool(boolean requiresTool) {
+		((AbstractBlockSettingsAccessor) this).setToolRequired(requiresTool);
+		return this;
+	}
+
+	public QuiltBlockSettings isAir(boolean isAir) {
```
LambdAurora: I really don't like using a getter-like name for this.
Maybe `airLike(boolean isAir)`?

Kroppeb: Ah, I thought the normal setter was also `settings.isAir()`, but apparently it's `settings.air()`

Leo40Git (approved):
LGTM!

OroArmor (approved):

library/block/block_extensions/src/main/java/org/quiltmc/qsl/block/extensions/api/QuiltBlockSettings.java
LambdAurora:
```suggestion
	public QuiltBlockSettings air(boolean isAir) {
```

Just change that and I'll be good with FCP-ing this PR.

LambdAurora (approved):

EnnuiL (changes requested):
A comment on line 51 at QuiltBlockSettings: We could get rid of thisAccessor, right? Also, renaming `otherAccessor` would be interesting

0xJoeMama (approved):

Kroppeb:
@EnnuiL 
I got rid of the variable, I don't see why the other one needs to be renamed, but if you want a different name, feel free to suggest one.

EnnuiL:
> @EnnuiL I got rid of the variable, I don't see why the other one needs to be renamed, but if you want a different name, feel free to suggest one.

Yeah, renaming it was only something to consider, however, yeah, maybe the current one is good enough

EnnuiL (approved):
Let's go!

PR #175, opened 2022-08-21
Make TrackedDataHandler registration use registry

Patbox:
This change allows for better mod compatibility, as now TrackedDataHandler's ids can be synced between server and client

PR #178, opened 2022-08-22
Custom Entity Spawn Data API

TropheusJ:
A small API for sending extra data with entity spawn packets.

Handles making a custom packet for you. Just implement `QuiltCustomSpawnDataEntity` and you can read/write directly to/from a buffer to modify spawn data. Useful when a data tracker is insufficient.

Merging this and #175 into 1 module should be considered.


library/entity/custom_spawn_data/src/main/java/org/quiltmc/qsl/entity/custom_spawn_data/impl/CustomEntitySpawnDataClientInitializer.java
```diff
@@ -0,0 +1,37 @@
...
+		ClientPlayNetworking.registerGlobalReceiver(QuiltCustomSpawnDataEntity.EXTENDED_SPAWN_PACKET, (client, handler, buf, sender) -> {
+			EntitySpawnS2CPacket spawnPacket = new EntitySpawnS2CPacket(buf);
+			buf.retain(); // save to let the entity read it
+			client.execute(() -> {
+				spawnPacket.apply(handler);
+				Entity spawnedEntity = client.world.getEntityById(spawnPacket.getId());
+				if (!(spawnedEntity instanceof QuiltCustomSpawnDataEntity customDataEntity)) {
+					logger.error("[Quilt] invalid entity received for extended spawn packet; expected a QuiltCustomSpawnDataEntity, got: " +
+							(spawnedEntity == null ? "null" : Registry.ENTITY_TYPE.getId(spawnedEntity.getType()).toString()));
+					return;
```
Leo40Git: `buf` is leaked here!
If you move this `if` into the `try` statement, this will resolve the leak (AFAICR, I _think_ Java runs the `finally` block before returning?)

Leo40Git (changes requested):
oh dear

LambdAurora (commented):
I would combine this module and #175's module into a entity_networking module.

I'll take a closer look to this later.

library/entity/custom_spawn_data/src/main/java/org/quiltmc/qsl/entity/custom_spawn_data/impl/CustomEntitySpawnDataClientInitializer.java
SilverAndro: Yes, finally will always run after a try..catch block, essentially "storing" the value to return/throw until after

TropheusJ: whoops, fixed. Just put the whole thing in the try.

TheGlitch76:
Marking as draft because this PR is marked as `s: not working`; feel free to mark it as ready for review once everything is working again and this is OK to go into QSL :pineapple:

TheGlitch76:
actually--i see that was done by Aurora, not you. I'll leave it to you to determine if this is OK or not.

TropheusJ:
on hold until #175 is merged (to make merging not terrible) and I have time

TropheusJ:
going to resurrect this as a new PR, sorry for letting this sit for so long for no real reason

PR #179, opened 2022-08-22
NetworkChannels.

0xJoeMama:
This PR adds:
* NetworkChannels to the networking API. They started as a test for NetworkCodecs but ended up making me like them enough to make them public API. They are a wrapper around Server/ClientPlayerNetworking that makes it so the user never has to interact with a PacketByteBuf. They contain a NetworkCodec and they use that as well as a supplied handler in order to handle messages. The user just needs to create the channel and specify a way to handle it.

More tests coming soon(TM).
This depends on #180 since it uses NetworkCodecs.

### DIFF
--- a/library/core/networking/build.gradle
+++ b/library/core/networking/build.gradle
@@ -12,6 +12,8 @@ qslModule {
 		core {
 			api("qsl_base")
 			api("lifecycle_events")
+			testmodOnly("registry")
+			testmodOnly("resource_loader")
 		}
 		management {
 			testmodOnly("command")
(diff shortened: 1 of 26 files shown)


LambdAurora:
With Patbox we were thinking of making so you could extend Packet in a way so you could use Vanilla's methods but internally they would turn back into CustomPayload packets. 
So I'm not sure I am a fan of the network channel stuff.
Also it kinda looks complicated? I'm not sure what to think of it.

0xJoeMama:
> Also it kinda looks complicated? I'm not sure what to think of it.

What is the complicated part? NetworkCodecs or the channels?

LambdAurora:
The channels.

The codec stuff looks straightforward, I mean, it's codecs.

0xJoeMama:
> The channels.

Do you mean in terms of implementation or in terms of user experience?

LambdAurora:
Both.

PR #181, opened 2022-08-22
Ignite block event

### LINKED ISSUE #177: Event triggered on item usage with QoL features.
Let's say I have a flint and steel, and want to burn vines so they don't grow more.
I then need some kind of injection into the flint and steel item to add this behavior.

Same with a custom campfire-like block, like a brazier, if it's extinguished I would need an injection into the flint and steel item to add the turning it back on behavior!

This is where an event could come into place, allowing to easily add those behaviors!

[Here's an existing mixin that I personally use to create the behaviors described earlier.](https://github.com/LambdAurora/AurorasDecorations/blob/1.19/src/main/java/dev/lambdaurora/aurorasdeco/mixin/item/FlintAndSteelItemMixin.java)

While I could've used FAPI's ItemUseCallback I didn't, I don't entirely remember the reason but it's something along the lines of it being a bit awkward.

But now we got some considerations for implementation:

First of all, let's look at the mixin and let's think about the stuff that could have been easily forgotten by modders in order of less likely to most likely:
 - forgetting to damage the tool
 - forgetting to emit the game event
 - forgetting to check for adventure mode
 
This could be solved by extending ItemUsageContext with injected interfaces, like `context.canModifyWorld()`, etc.

Second, look into make this event as less confusing as possible!

Third, make sure it's easy to use!

Leo40Git:
~~Closes~~ #177.

Adds the new `item_extensions` module.


library/item/item_extensions/src/main/java/org/quiltmc/qsl/item/extensions/api/event/ItemInteractionEvents.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.quiltmc.qsl.base.api.event.Event;
+import org.quiltmc.qsl.base.api.event.EventAwareListener;
+
+public final class ItemInteractionEvents {
+	private ItemInteractionEvents() {
+		throw new UnsupportedOperationException("BlockInteractionEvents only contains static declarations.");
+	}
+
+	public static final Event<IgniteBlock> IGNITE_BLOCK = Event.create(IgniteBlock.class, callbacks -> context -> {
```
LambdAurora, line 32: I think I'd rather delay this until we have an actual general event before looking into more specialized events!

Leo40Git: Fair enough, I'll transform this into a generic Item Interaction Events PR later.

Leo40Git:
(closing this so I can rename the branch, don't mind me)

PR #182, opened 2022-08-22
Item Interaction Events

Leo40Git:
As the name implies, this adds events that fire when using items.

This PR adds the `item_events` module.

### TODO
- [x] "Used" event
- [x] "Used on block" event
- [x] "Used on entity" event
- [x] "Finished using" event
  - Not actually sure what uses this. Bows and crossbows, maybe?


OroArmor (changes requested):
Looks good, but i think it would fit better under an `item_event` module

Leo40Git:
Moved to `item_events`!

0xJoeMama:
Hm it all looks good, but what is your opinion on a adding generic UseItem event as well? One that fires when the 'use' method is called?
Maybe an 'inventoryTick' event as well, although that's not really an interaction and is therefore not in line with this PR's description.

My point is, this can be expanded. We should look into that.

library/item/item_events/src/main/java/org/quiltmc/qsl/item/events/api/ItemInteractionEvents.java
```diff
@@ -0,0 +1,103 @@
...
+import net.minecraft.util.Hand;
+import net.minecraft.util.TypedActionResult;
+import net.minecraft.world.World;
+
+import org.quiltmc.qsl.base.api.event.Event;
+import org.quiltmc.qsl.base.api.event.EventAwareListener;
+
+public final class ItemInteractionEvents {
+	private ItemInteractionEvents() {
+		throw new UnsupportedOperationException("BlockInteractionEvents only contains static declarations.");
```
Leo40Git:
```suggestion
		throw new UnsupportedOperationException("ItemInteractionEvents only contains static declarations.");
```

PR #183, opened 2022-08-22
Initial implementation of Entity Interaction Events

Quplet:
Adds an `interaction` module containing (as of now) the basic Player interaction events similar to Fabric's implementation, and a LivingEntityAttack event.

TODO:

- [ ] Separate cancellable callbacks into `BEFORE` and `AFTER` events
  - [x] LivingEntityAttackEvents
  - [ ] AttackBlockEvents (Player)
  - [x] AttackEntityEvents (Player)
  - [x] UseBlockEvents (Player)
  - [x] UseEntityEvents (Player)
  - [x] UseItemEvents (Player)
- [ ] Add more LivingEntity events (like EnderManPickUpBlock event)


library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/mixin/LivingEntityMixin.java
```diff
@@ -0,0 +1,46 @@
...
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+@Mixin(LivingEntity.class)
+public abstract class LivingEntityMixin extends Entity {
+
+	@Inject(method = "damage", at = @At("HEAD"), cancellable = true)
+	private void onTakeDamage(DamageSource source, float amount, CallbackInfoReturnable<Boolean> cir) {
+		if (source.getAttacker() instanceof LivingEntity attacker) {
+			boolean result = LivingEntityAttackCallback.EVENT.invoker().onAttack(attacker, this, source, amount);
```
0xJoeMama: What if someone wanted to modify the amount of damage dealt to the entity?

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/mixin/ServerPlayerInteractionManagerMixin.java
```diff
@@ -0,0 +1,111 @@
...
+	protected ServerPlayerEntity player;
+
+	@Shadow
+	protected ServerWorld world;
+
+	@Inject(method = "interactItem", at = @At("HEAD"), cancellable = true)
+	private void onPlayerInteractItem(ServerPlayerEntity player, World world, ItemStack stack, Hand hand, CallbackInfoReturnable<ActionResult> cir) {
+		if (player.isSpectator()) return;
+
+		TypedActionResult<ItemStack> result = UseItemCallback.EVENT.invoker().onUseItem(player, world, hand);
```
0xJoeMama: Since you are just using result.getResult() it might be better to make the event return an ActionResult

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/player/UseItemCallback.java
```diff
@@ -0,0 +1,62 @@
...
+	 * Invoked when a player uses (right-clicks) with an item.
+	 *
+	 * @param player the interacting player
+	 * @param world the world the event occurs in
+	 * @param hand the hand used
+	 * @return {@link TypedActionResult#success(Object)} to cancel processing and send a packet to the server,
+	 * {@link TypedActionResult#pass(Object)} to fall back to further processing,
+	 * {@link TypedActionResult#fail(Object)} to cancel further processing.
+	 */
+	TypedActionResult<ItemStack> onUseItem(PlayerEntity player, World world, Hand hand);
```
0xJoeMama: I think providing the ItemStack in the player's hand in this method is gonna be helpful. I can't count the amount of times I had to write player.getStackInHand(hand) when listening to this event.

0xJoeMama (commented):
Some generic nitpicks and ideas.

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/mixin/LivingEntityMixin.java
Quplet: I've been trying to think of a way to best do this while still being able to cancel the method and possibly pass the new damage value from one listener to the next.

0xJoeMama: A shared context could work. Like a mutable object that you can pass in as a callback parameter that's just gonna contain 1. damage amount and 2. whether the callback was cancelled or not.
This way all listeners would get the modified damage from *all* the previous ones and anyone can cancel.

0xJoeMama: I can make a draft if you want.

0xJoeMama: Something like this
image

Quplet: If you would I'd appreciate it. I'm curious how you would set the damage in the method while still still remaining cancellable. I've considered `@ModifyVariable` or `@ModifyArg` but I don't believe you can cancel a method with them. I was going to look into possibly multiple injects, but I'm not sure.

Quplet: > Something like this image

Thx, the context thing itself isn't what I'm struggling with. I'm struggling with actually setting the new damage value in the mixin while also remaining cancellable. I think I have a working system that is kinda scuffed imo. I'm going to get some feedback from the toolchain discord before I push it to my repo.

Thanks for the suggestions!

LambdAurora: I am worried about the way this would be handled.
The issue with the cancellability is how do you listen for something and act upon it without it beign canceled? since if it cancels your action could still trigger and that's an issue.

Cypher121: Two-phase event, one for possible cancellations, second to act in case event has not been cancelled?

Cypher121: Alternatively, one pass, but collect a list of deferred actions and final cancellation results they should be executed on from every listener

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/player/AttackBlockCallback.java
```diff
@@ -0,0 +1,62 @@
...
+ * A callback that is invoked if a player attacks (left-clicks) a block.
+ * <p>
+ * Is hooked before the spectator check, so make sure to check the player's game mode!
+ * <ul>
+ *     <li>{@link ActionResult#SUCCESS} cancels further processing and, on the client, sends a packet to the server.</li>
+ * 	   <li>{@link ActionResult#PASS} falls back to further processing.</li>
+ * 	   <li>{@link ActionResult#FAIL} cancels further processing and does not send a packet to the server.</li>
+ * </ul>
+ */
+public interface AttackBlockCallback {
```
TheGlitch76, line 38: I know this is based on the QM name, but I don't like calling it "attacking" a block, because it's really the beginning of a player breaking (destroying) a block.

Would this maybe make more sense if it was part of `BreakBlockEvents`?

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/player/BreakBlockEvents.java
```diff
@@ -0,0 +1,124 @@
...
+		/**
+		 * Invoked if the block breaking event has been canceled.
+		 *
+		 * @param player the player breaking the block
+		 * @param world the world the block in broken in
+		 * @param pos the position of the block
+		 * @param state the block state <strong>before</strong> the block is broken
+		 * @param blockEntity the block entity <strong>before</strong> the block is broken, can be {@code null}
+		 */
+		void cancelPlayerBreakBlock(PlayerEntity player, World world, ItemStack stack, BlockPos pos, BlockState state, @Nullable BlockEntity blockEntity);
```
TheGlitch76: I'd prefer `onCancelPlayerBreakBlock` to signify that this is a listener; the current name implies the method somehow cancels the event.

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/player/BreakBlockEvents.java
```diff
@@ -0,0 +1,124 @@
...
+		 *
+		 * @param player the player breaking the block
+		 * @param world the world the block in broken in
+		 * @param pos the position of the block
+		 * @param state the block state <strong>before</strong> the block is broken
+		 * @param blockEntity the block entity <strong>before</strong> the block is broken, can be {@code null}
+		 * @return {@code false} to cancel the event and the block breaking action,
+		 * otherwise {@code true} to pass to the next listener
+		 */
+		boolean beforePlayerBreakBlock(PlayerEntity player, World world, ItemStack stack, BlockPos pos, BlockState state, @Nullable BlockEntity blockEntity);
```
TheGlitch76: All of these should probably conjugate the verb like `beforePlayerBreaksBlock`

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/player/BreakBlockEvents.java
```diff
@@ -0,0 +1,124 @@
...
+		 * @param world the world the block in broken in
+		 * @param pos the position of the block
+		 * @param state the block state <strong>before</strong> the block is broken
+		 * @param blockEntity the block entity <strong>before</strong> the block is broken, can be {@code null}
+		 */
+		void afterPlayerBreakBlock(PlayerEntity player, World world, ItemStack stack, BlockPos pos, BlockState state, @Nullable BlockEntity blockEntity);
+	}
+
+	@FunctionalInterface
+	public interface Cancel {
```
TheGlitch76: `Canceled` to match the event name?

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/player/AttackBlockCallback.java
Quplet: I think they have different purposes... its kinda hard to describe. The *action* of left clicking vs the *event* of breaking.
Break block events to me are designed to be acted upon the block's breaking more than the player's action while this is more the player's action than the block breaking, if that makes any sense.

I'm fine with moving it over tho if people agree it fits better with the block breaking events.

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/DamageContext.java
```diff
@@ -0,0 +1,107 @@
...
+import net.minecraft.entity.Entity;
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.entity.damage.DamageSource;
+import net.minecraft.item.ItemStack;
+
+/**
+ * An encapsulation of several arguments relevant to a damage context with a
+ *     mutable damage value and cancellation that is shared across all listeners.
+ */
+public class DamageContext {
```
Platymemo, line 29: Annotating members with `@Nullable` and `@NotNull` would be really helpful!

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/LivingEntityAttackEvents.java
```diff
@@ -0,0 +1,83 @@
...
+		/**
+		 * Invoked <strong>after</strong> a {@link LivingEntity} damages another.
+		 *
+		 * @param attacker the attacking entity
+		 * @param stack the {@link ItemStack} in the attacker's main-hand
+		 * @param target the target {@link LivingEntity}
+		 * @param source the {@link DamageSource} of the attack
+		 * @param damage the damage that was dealt
+		 */
+		void afterDamage(LivingEntity attacker, ItemStack stack, LivingEntity target, DamageSource source, float damage);
```
Platymemo, line 81: I wonder if there should be two `DamageContext`s. A `MutableDamageContext` for the before event, and an `ImmutableDamageContext` (names very much up for debate) for the after event. Then these interfaces can even be merged into one.

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/player/AttackBlockCallback.java
```diff
@@ -0,0 +1,62 @@
...
+import net.minecraft.util.ActionResult;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.Direction;
+import net.minecraft.world.World;
+import org.quiltmc.qsl.base.api.event.Event;
+
+/**
+ * A callback that is invoked if a player attacks (left-clicks) a block.
+ * <p>
+ * Is hooked before the spectator check, so make sure to check the player's game mode!
```
Platymemo, line 31:
```suggestion
 * This event fires before the spectator check, so make sure to check the player's game mode!
```

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/player/AttackBlockCallback.java
```diff
@@ -0,0 +1,62 @@
...
+ * <ul>
+ *     <li>{@link ActionResult#SUCCESS} cancels further processing and, on the client, sends a packet to the server.</li>
```
Platymemo, line 33: linking the class of the packet would be useful! (also in all the other places that have similar docs)

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/player/UseBlockEvents.java
```diff
@@ -0,0 +1,96 @@
...
+public class UseBlockEvents {
+
+	/**
+	 * A callback that is invoked <strong>before</strong> a {@link PlayerEntity} right-clicks a block.
+	 * <p>
+	 * Implementations should not assume the action has been completed.
+	 * <p>
+	 * Upon return:
+	 * <ul>
+	 *     <li>{@link ActionResult#SUCCESS}/{@link ActionResult#CONSUME}/{@link ActionResult#CONSUME_PARTIAL}
```
Platymemo:
```suggestion
	 *     <li>{@link ActionResult#SUCCESS}, {@link ActionResult#CONSUME}, or {@link ActionResult#CONSUME_PARTIAL}
```
Alternatively, iirc there is a method in `ActionResult` that checks this for you that you could link instead.

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/player/UseEntityEvents.java
```diff
@@ -0,0 +1,97 @@
...
+public class UseEntityEvents {
+
+	/**
+	 * A callback that is invoked <strong>before</strong> a {@link PlayerEntity} right-clicks an {@link Entity}.
+	 * <p>
+	 * Implementations should not assume the action was completed.
+	 * <p>
+	 * Upon return:
+	 * <ul>
+	 *     <li>{@link ActionResult#SUCCESS}/{@link ActionResult#CONSUME}/{@link ActionResult#CONSUME_PARTIAL}
```
Platymemo:
```suggestion
	 *     <li>{@link ActionResult#SUCCESS}, {@link ActionResult#CONSUME}, or {@link ActionResult#CONSUME_PARTIAL}
```

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/player/UseBlockEvents.java
```diff
@@ -0,0 +1,96 @@
...
+	@FunctionalInterface
+	public interface Before {
+		/**
+		 * Invoked <strong>before</strong> a {@link PlayerEntity} right-clicks a block.
+		 *
+		 * @param player the interacting {@link PlayerEntity}
+		 * @param world the {@link World} the event occurs in
+		 * @param hand the {@link Hand} used
+		 * @param hitResult the {@link BlockHitResult} of the interaction
+		 * @return {@link ActionResult#SUCCESS}/{@link ActionResult#CONSUME}/{@link ActionResult#CONSUME_PARTIAL}
```
Platymemo, line 76:
```suggestion
		 * @return {@link ActionResult#SUCCESS}, {@link ActionResult#CONSUME}, or {@link ActionResult#CONSUME_PARTIAL}
```

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/player/UseEntityEvents.java
```diff
@@ -0,0 +1,97 @@
...
+		/**
+		 * Invoked <strong>before</strong> a player uses (right-clicks) an entity.
+		 *
+		 * @param player the interacting {@link PlayerEntity}
+		 * @param world the {@link World} the event occurs in
+		 * @param hand the {@link Hand} used
+		 * @param stack the {@link ItemStack} used by the player
+		 * @param entity the right-clicked {@link Entity}
+		 * @param hitResult the {@link EntityHitResult} of the interaction
+		 * @return {@link ActionResult#SUCCESS}/{@link ActionResult#CONSUME}/{@link ActionResult#CONSUME_PARTIAL}
```
Platymemo, line 75:
```suggestion
		 * @return {@link ActionResult#SUCCESS}, {@link ActionResult#CONSUME}, or {@link ActionResult#CONSUME_PARTIAL}
```

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/player/UseItemEvents.java
```diff
@@ -0,0 +1,89 @@
...
+public class UseItemEvents {
+
+	/**
+	 * A callback that is invoked <strong>before</strong> a {@link PlayerEntity} right-clicks with an item.
+	 * <p>
+	 * Implementations should not assume the action has been completed.
+	 * <p>
+	 * Upon return:
+	 * <ul>
+	 *     <li>{@link ActionResult#SUCCESS}/{@link ActionResult#CONSUME}/{@link ActionResult#CONSUME_PARTIAL}
```
Platymemo:
```suggestion
	 *     <li>{@link ActionResult#SUCCESS}, {@link ActionResult#CONSUME}, or {@link ActionResult#CONSUME_PARTIAL}
```

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/player/UseItemEvents.java
```diff
@@ -0,0 +1,89 @@
...
+
+	@FunctionalInterface
+	public interface Before {
+		/**
+		 * Invoked <strong>before</strong> a {@link PlayerEntity} right-clicks with an item.
+		 *
+		 * @param player the interacting {@link PlayerEntity}
+		 * @param world  the {@link World} the event occurs in
+		 * @param hand   the {@link Hand} used
+		 * @return {@link ActionResult#SUCCESS}/{@link ActionResult#CONSUME}/{@link ActionResult#CONSUME_PARTIAL}
```
Platymemo, line 70:
```suggestion
		 * @return {@link ActionResult#SUCCESS}, {@link ActionResult#CONSUME}, or {@link ActionResult#CONSUME_PARTIAL}
```

Platymemo (approved):
I like this! Just some nitpicks.

Also, an alternative for the `DamageContext` is to have a separate `MODIFY_DAMAGE` event that fires if the before event is not canceled. That way all the events are cohesive.

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/DamageContext.java
Quplet: Imo, nothing in it should ever be null. Should I just annotate everything with `@notNull`?

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/LivingEntityAttackEvents.java
Quplet: So two more classes extending a abstract `DamageContext`? I think that can work.

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/player/AttackBlockCallback.java
Quplet: Easy enough to do, but I'm not so sure why that would be helpful. All the packet really does is tell the logical server to process the event on their side.

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/LivingEntityAttackEvents.java
Quplet: The separation of the `Before` and `After` functional interfaces I believe are more for clarification purposes however. The documentation could become a bit confusing if they are merged.

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/mixin/LivingEntityMixin.java
```diff
@@ -0,0 +1,61 @@
...
+		if (source.getAttacker() instanceof LivingEntity attacker) {
+			DamageContext context = new DamageContext(attacker, attacker.getMainHandStack(), (LivingEntity)(Object)this, source, damage);
+			LivingEntityAttackEvents.BEFORE.invoker().beforeDamage(context);
+			quilt$damageCanceled = context.isCanceled();
+			return context.getDamage();
+		}
+		return damage;
+	}
+
+	@Inject(method = "damage", at = @At(value = "HEAD", shift = At.Shift.AFTER), cancellable = true)
```
Platymemo, line 47: I'd say give this method priority so other `@Inject`s at `HEAD` get run after this and it can cancel before other mods do funky things.

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/LivingEntityAttackEvents.java
Platymemo: Coming back to this, I think a separate `MODIFY_DAMAGE` event would be better. That way you have the ability to cancel it, modify it, or react to it in clear, separate events.

Platymemo (changes requested):

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/player/AttackBlockCallback.java
Platymemo: Just lets devs jump around the codebase a little easier, in case they want to look at how the interaction is processed

library/entity/interaction/src/main/java/org/quiltmc/qsl/entity/interaction/api/LivingEntityAttackEvents.java
Quplet: I believe this would create a lot of duplicate code on the user's part. I think logic that determines cancellation or modifying damage would align a lot of the time and I'd rather not force users to duplicate code to perform these tasks.

PR #188, opened 2022-08-31
Fix REA syncing too early

Leo40Git:
This PR moves re-synchronization on `/reload` to the `ResourceLoaderEvents.END_DATA_PACK_RELOAD` event.

Most importantly, this means that REAs will be applied _after_ tags.


EnnuiL:
On my own test with Crooked Crooks and EMI, I can confirm that the issue persists

Leo40Git:
this PR a) is a complete mess thanks to the licenser acting weird b) doesn't fix the problem
closing this for now, will create a new one once we actually figure out why REA is completely breaking for Ennui

PR #190, opened 2022-09-04
Rename registry_entry_attachments to registry_entry_attachment

Leo40Git:
The old ID will still work, as a "provides" entry has been added.

This PR also touches the `block_content_registry` and `item_content_registry` modules (hence Core reviews being requested). And by that, I mean that there is a one character change in their `build.gradle` file.


LambdAurora (approved):

OroArmor (approved):

EnnuiL (approved):

EnnuiL (changes requested):
Wait! The testmod has to be renamed too

EnnuiL (approved):

PR #191, opened 2022-09-05
Virtual Resource Pack API

LambdAurora:
This pull request so far introduces:

- `MutableResourcePack`, an extension of `ResourcePack` which can add resources.
- `InMemoryResourcePack`, a resource pack implementation that keeps in RAM the resources
- Events to register default resource packs (may be virtual)
- Events to register resource packs that are on top of the pack hierarchy (may be virtual).
- Fix up Tag API data pack discovery so it now includes virtual resource packs.

This will hopefully cover most use cases (and hopefully won't introduce a big oopsie).


LambdAurora:
Not entirely complete as I'm missing 1 or 2 testmods (and some documentation), but the global structure should be done!

Marking as ready for review as a lot of the API can be already reviewed.

LambdAurora:
Almost forgot, there is one thing to ask:

Should InMemoryResourcePack be API or only the Nameable subclass would be instantiable through a static method?
The second option allow to easily change details of the implementation without breaking API.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourceLoader.java
```diff
@@ -78,6 +80,24 @@ public interface ResourceLoader {
 	 */
 	void registerResourcePackProfileProvider(@NotNull ResourcePackProvider provider);
 
+	/**
+	 * {@return the registration of default resource packs event}
+	 * <p>
+	 * This event is triggered when the default resources are created and allow to register additional resource packs
+	 * that are displayed to the user as the {@code "Default"} resource pack.
```
Leo40Git, line 87: Does this mean that the resource packs are merged into the "Default" resource pack?

Leo40Git (approved):
API LGTM! Just one thing needs some clearing up...

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourceLoader.java
LambdAurora: Yes, they are merged into that resource pack as it simplifies a ton their injection, they are merged in a way they have more priority than Vanilla but less priority than modded resource packs from disk.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/InMemoryResourcePack.java
```diff
@@ -0,0 +1,297 @@
...
+				.collect(Collectors.toUnmodifiableSet());
+	}
+
+	@Override
+	public <T> @Nullable T parseMetadata(ResourceMetadataReader<T> metaReader) throws IOException {
+		if (!this.root.containsKey(ResourcePack.PACK_METADATA_NAME)) {
+			var json = new JsonObject();
+			var packJson = new JsonObject();
+			packJson.addProperty("description", "A virtual resource pack.");
+			packJson.addProperty("pack_format", 5); // This is like, not read by any significant system when invisible to users.
```
EnnuiL, line 136: i know that this isn't being read by anyone here, but
```suggestion
			packJson.addProperty("pack_format", 10); // This is like, not read by any significant system when invisible to users.
```
pretty please? :P

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/InMemoryResourcePack.java
```diff
@@ -0,0 +1,297 @@
...
+
+/**
+ * Represents an in-memory resource pack.
+ * <p>
+ * The resources of this pack are stored in memory instead of it being on-disk.
+ */
+public abstract class InMemoryResourcePack implements MutableResourcePack {
+	private static final Logger LOGGER = LogUtils.getLogger();
+	private static final ExecutorService EXECUTOR_SERVICE;
+	private static final boolean DUMP = TriState.fromProperty("quilt.resource_loader.debug.pack.dump_from_in_memory")
```
EnnuiL, line 67: Add this to the module's README, 'cause something tells me this will be important later on

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/InMemoryResourcePack.java
```diff
@@ -0,0 +1,297 @@
...
+	private Map<Identifier, Supplier<byte[]>> getResourceMap(ResourceType type) {
+		return switch (type) {
+			case CLIENT_RESOURCES -> this.assets;
+			case SERVER_DATA -> this.data;
+		};
+	}
+
+	static {
+		int threads = Math.max(Runtime.getRuntime().availableProcessors() / 2 - 1, 1);
+		final String property = "quilt.resource_loader.pack.virtual_async_threads";
```
EnnuiL: This should perhaps be a static field on the top of the file, for better discoverability

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/QuiltMultiResourcePackManagerHooks.java
```diff
@@ -1,5 +1,5 @@
 /*
```
EnnuiL: Commentary about the entire diff itself: gitttttttttttttttt

EnnuiL (changes requested):
So far? So good! I only have few nitpicks, mostly about the properties

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/InMemoryResourcePack.java
LambdAurora: I chose `5` because it's the smallest number that doesn't have special treatment by MC.
This feels very arbitrary.

EnnuiL: oh, I see; I mentioned 10 'cause I imagined that the current number was going to be the one not subject to things such as DFU

EnnuiL (approved):

maximumpower55 (approved):
LGTM, ~~only thing I would like is a way to make the in-memory resource pack invisible in the selection screen~~ (a way for this already exists, answered on discord)

oliviathevampire (approved):

OroArmor (approved):

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/MutableResourcePack.java
```diff
@@ -0,0 +1,307 @@
...
+	/**
+	 * Puts an image resource into the resource pack's root.
+	 * <p>
+	 * <b>Note:</b> this method is only available on the client.
+	 *
+	 * @param fileName the name of the file
+	 * @param image    the resource content
+	 * @see #putResource(String, byte[])
+	 */
+	@Environment(EnvType.CLIENT)
```
TheEpicBlock, line 196: It seems weird to me that all the methods here are made to be side-agnostic except these ones. I'm not really sure why these types were included in particular. `NativeImage` doesn't really seem like a useful format to be working with for dynamic images either. `ImageIO` includes more tools to edit images at least.

LambdAurora: I personally am using NativeImage for anything image-related in MC.
And it's a client-only class so those methods must be client-only.

I usually avoid touching anything in the java.awt package due to the famous macOs issues.
It shouldn't happen with ImageIO though, so idk.
I've always found the BufferedImages more annoying to work with compared to NativeImage, at least one the format is easy.

TheEpicBlock: Fair enough, I guess that would be personal preference, and most of the actually useful image editing is actually in `java.awt` so it might not be the best of bets. I really like how most of the stuff in here is side-agnostic (for PolyMc hackery reasons) so I was just curious about this one case. I just hope people won't start registering their resource packs inside the client mod initializer but I know they'll probably do that anyway.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/InMemoryResourcePack.java
Juuxel: Isn't there a constant for the latest resource pack version? I'd use that to always match the latest MC.

LambdAurora: The issue is the pack version is side dependent, data packs and resource packs don't have the same versions, making it impossible to pair them up as one value.

PR #192, opened 2022-09-07
Entity tick events

williambl:
# Entity Tick Events

**This PR requires #42 to be merged first.** 

## Motivation

Mods may often want to run some code every tick for an entity. This requires two separate injections, as different tick methods are run depending on whether the entity is a passenger or not. A unified event makes this simpler.

## Content

This PR adds two events, `ClientEntityTickCallback` and `ServerEntityTickCallback`. They are invoked on the logical client and server respectively every time an entity is ticked. They have an Entity parameter for the entity being ticked, and a boolean parameter which specifies whether the entity is a passenger of another.

It also adds tests for these events.


williambl:
Marking as draft until #42 is merged.

williambl:
diff is horribly broken i'm just going to cherry pick the actual changes onto a new branch

PR #194, opened 2022-09-17
Registry Entry Attachment: verify entry is registered

### LINKED ISSUE #193: Registry Entry Attributes: Verify registry entry is registered
(this is just so I don't inevitably forget about this _again_ - it's important, and also QFAPI already has mechanisms in place to prevent it from being a Fabric compat problem)

Leo40Git:
Closes #193.
Also does some light Checkstyle fixes.

_This PR was created live on stage!_


EnnuiL (approved):

OroArmor (approved):

EnnuiL:
Testing with a broken version of BetterEnd that broke because of the block of charcoal being registered in the fuel map REA earlier than its own registration, I managed to heap multiples of 9 charcoal; Marking this as `s: not working` for now

library/data/registry_entry_attachment/src/main/java/org/quiltmc/qsl/registry/attachment/impl/RegistryEntryAttachmentImpl.java
```diff
@@ -176,6 +176,10 @@ public Iterator<TagEntry<R, V>> tagEntryIterator() {
 
 	@Override
 	public void put(R entry, V value) {
+		if (this.registry.getId(entry) == null) {
```
Leo40Git, line 179: I just realized something: this check does not work with defaulted registries, such as entity types.
Is there a `Registry.contains` method?

EnnuiL:
...turns out the test wasn't correct for this PR due to it actually being part of a QFAPI issue; A proper test case has been especially built for this PR (and the QFAPI issue) which should return a more concrete result

EnnuiL (approved):
The test went well! Let's do this

PR #199, opened 2022-10-08
More flexible ItemGroup addition API

lukebemish:
This relatively small API adds a single method to register ItemGroups not created through the QuiltItemGroup builder.

PR #200, opened 2022-10-09
Custom ItemGroup Icon Renderers

maximumpower55:
Title pretty much said it all.
Javadocs for the new classes and methods aren't really there so would appreciate for code suggestions that add them


library/item/item_group/src/main/java/org/quiltmc/qsl/item/group/mixin/client/CreativeInventoryScreenMixin.java
```diff
@@ -165,6 +173,18 @@ private void renderTabIcon(MatrixStack matrixStack, ItemGroup itemGroup, Callbac
 		if (this.quilt$isGroupNotVisible(itemGroup)) {
 			info.cancel();
 		}
+
+		this.quilt$renderTabIcon$matrices = matrixStack;
+		this.quilt$renderTabIcon$itemGroup = itemGroup;
+	}
+
+	@ModifyArgs(method = "renderTabIcon", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/render/item/ItemRenderer;renderInGuiWithOverrides(Lnet/minecraft/item/ItemStack;II)V"))
+	private void renderCustomTabIcon(Args args) {
```
lukebemish: You can get the method args in a ModifyArgs; there's no reason for the fields you're using

library/item/item_group/src/main/java/org/quiltmc/qsl/item/group/api/QuiltItemGroup.java
```diff
@@ -217,6 +236,18 @@ public Builder icon(@NotNull Supplier<ItemStack> iconSupplier) {
 			return this;
 		}
 
+		/**
+		 * This is used to add an icon to the {@link ItemGroup}.
+		 *
+		 * @param icon
+		 * @return {@code this}
+		 */
+		public Builder icon(@NotNull ItemGroupIcon icon) {
```
lukebemish: Does this work on a dedicated server? I'm a bit wary of it causing loading of MatrixStack (whether by nature or when someone tries to use it), which as far as I'm aware doesn't exist on the server.

library/item/item_group/src/main/java/org/quiltmc/qsl/item/group/api/QuiltItemGroup.java
```diff
@@ -45,13 +46,15 @@
  */
 public final class QuiltItemGroup extends ItemGroup {
 	private final Supplier<ItemStack> iconSupplier;
+	private Optional<ItemGroupIcon> customIcon;
 	private @Nullable ItemStack icon;
 	private final @Nullable Consumer<List<ItemStack>> stacksForDisplay;
 	private final @Nullable Text displayText;
 
-	private QuiltItemGroup(int index, String id, Supplier<ItemStack> iconSupplier, @Nullable Consumer<List<ItemStack>> stacksForDisplay, @Nullable Text displayText) {
+	private QuiltItemGroup(int index, String id, Supplier<ItemStack> iconSupplier, Optional<ItemGroupIcon> customIcon, @Nullable Consumer<List<ItemStack>> stacksForDisplay, @Nullable Text displayText) {
```
lukebemish: I very much don't like that this is limited to just QuiltItemGroups, for two reasons: first of all, it's not very future proof. If we ever want to extend this system, we've got to add stuff to the constructor, so we can't make the constructor public and so people can't extend the functionality of the class. Second of all, it limits anyone wanting to do custom rendering to having to use QuiltItemGroup; especially with #199, I see no reason for this; it also causes pain for anyone doing multiloader stuff.

library/item/item_group/src/testmod/java/org/quiltmc/qsl/item/group/test/ItemGroupTest.java
```diff
@@ -48,6 +49,9 @@ public class ItemGroupTest implements ModInitializer {
 							.forEach(stacks::add)
 			).build();
 
+	private static final QuiltItemGroup ITEM_GROUP_WITH_TEXTURE_ICON = QuiltItemGroup.createWithIcon(new Identifier(NAMESPACE, "test_texture_icon_group"),
```
lukebemish: Yeah, this definitely cannot work on the server.

library/item/item_group/src/main/java/org/quiltmc/qsl/item/group/impl/ItemGroupTextureIcon.java
```diff
@@ -0,0 +1,68 @@
...
+import net.minecraft.client.texture.MissingSprite;
+import net.minecraft.client.texture.TextureManager;
+import net.minecraft.client.util.math.MatrixStack;
+import net.minecraft.resource.ResourceManager;
+import net.minecraft.util.Identifier;
+
+@ApiStatus.Internal
+public final class ItemGroupTextureIcon implements ItemGroupIcon, ClientResourceLoaderEvents.EndResourcePackReload {
+	// NB: These have to be suppliers since these objects are created before minecraft has a instance?
+	private final Supplier<TextureManager> textureManagerSupplier = () -> MinecraftClient.getInstance().getTextureManager();
```
lukebemish: Is there any reason to have a field for these at all if you're just getting them every time you use them anyways?

lukebemish:
Yeah, that looks like a dedicated server failure. I suspect that this is going to require a fundamentally different approach to work with dedicated servers - there's a reason stuff like block render layers is set up in a map the way it is. Attaching this stuff to the object directly could work, if you have a client-only mixin that just shoves an interface onto ItemGroups, and some client side API for adding renderers to ItemGroups using that interface. Otherwise, I think you'd have to go with some sort of map

library/item/item_group/src/main/java/org/quiltmc/qsl/item/group/api/QuiltItemGroup.java
Leo40Git: ...also, that `Optional<ItemGroupIcon>` should just be a `@Nullable ItemGroupIcon`.

LambdAurora (changes requested):
Considering @lukebemish's review, please change the implementation to using a `Map<ItemGroup, ItemGroupIcon>`.

`ItemGroupIcon` must be placed into a `client` subpackage and be annotated with `@Environment(EnvType.CLIENT)`.

The ItemGroupIcon probably should take as a generic ItemGroup parameter.

Which gives us `interface ItemGroupIcon<IG extends ItemGroup> { void render(IG itemGroup, MatrixStack matrices, int x, int y, float tickDelta); /* ... */}`.

Also please consider including the tickDelta to properly create animations.

maximumpower55:
Aurora's and lukebemish's comments have been resolved.

library/item/item_group/src/main/java/org/quiltmc/qsl/item/group/api/client/ItemGroupIconRenderer.java
```diff
@@ -0,0 +1,47 @@
...
+import net.fabricmc.api.Environment;
+import net.minecraft.client.util.math.MatrixStack;
+import net.minecraft.item.ItemGroup;
+import net.minecraft.util.Identifier;
+
+@Environment(EnvType.CLIENT)
+public interface ItemGroupIconRenderer<IG extends ItemGroup> {
+	static <IG extends ItemGroup> ItemGroupIconRenderer<IG> register(@NotNull IG itemGroup, @NotNull Function<IG, ItemGroupIconRenderer<IG>> renderer) {
+		Objects.requireNonNull(itemGroup, "itemGroup may not be null");
+		Objects.requireNonNull(renderer, "renderer may not be null");
```
lukebemish, line 37: Can we make this nullable as a way to *un*-register renderers? Maybe not necessary - just an idea.

library/item/item_group/src/main/java/org/quiltmc/qsl/item/group/mixin/client/CreativeInventoryScreenMixin.java
```diff
@@ -165,6 +174,19 @@ private void renderTabIcon(MatrixStack matrixStack, ItemGroup itemGroup, Callbac
 		if (this.quilt$isGroupNotVisible(itemGroup)) {
 			info.cancel();
 		}
+
+		this.quilt$renderTabIcon$matrices = matrixStack;
+		this.quilt$renderTabIcon$itemGroup = itemGroup;
+	}
+
+	@Redirect(method = "renderTabIcon", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/render/item/ItemRenderer;renderInGuiWithOverrides(Lnet/minecraft/item/ItemStack;II)V"))
```
lukebemish: Can we use a ModifyArgs here instead? (And just change the item arg to render an empty itemstack). This just seems like something that others might be liable to target (if they were already doing custom rendering, or on fabric mods)

library/item/item_group/src/main/java/org/quiltmc/qsl/item/group/mixin/client/CreativeInventoryScreenMixin.java
```diff
@@ -165,6 +174,19 @@ private void renderTabIcon(MatrixStack matrixStack, ItemGroup itemGroup, Callbac
 		if (this.quilt$isGroupNotVisible(itemGroup)) {
 			info.cancel();
 		}
+
+		this.quilt$renderTabIcon$matrices = matrixStack;
```
lukebemish: Once again - no reason to capture these to a field. Just capture them in the `Redirect` below (or preferably a `ModifyArgs`)

lukebemish (changes requested):
Looks mostly good to me - however, I'd clean up that mixin a bit. There's no reason to use the temporary fields there, and I'd say prefer ModifyArgs over Redirect because I could see existing mods (or fabric mods) targeting that same spot.

library/item/item_group/src/main/java/org/quiltmc/qsl/item/group/api/client/ItemGroupIconRenderer.java
maximumpower55: eh I will do it if more people want it.

library/item/item_group/src/main/java/org/quiltmc/qsl/item/group/mixin/client/CreativeInventoryScreenMixin.java
maximumpower55: Can't use `ModifyArgs` here because there is no guarantee, that the item group that has a custom icon renderer doesn't have a regular icon supplier and I don't want the custom icon renderer to render over or under the regular item icon, so I have to use `Redirect` to cancel that vanilla behaviour if the item group has a custom icon renderer, please correct me If I am wrong.

library/item/item_group/src/main/java/org/quiltmc/qsl/item/group/mixin/client/CreativeInventoryScreenMixin.java
maximumpower55: Eh so I could use a local capture but first off that can't capture the matrices to my knowledge, second off `renderTabIcon` has about 8 locals that I don't want to filter through.

library/item/item_group/src/main/java/org/quiltmc/qsl/item/group/mixin/client/CreativeInventoryScreenMixin.java
maximumpower55: I could use a `Inject` with a `CallbackInfo.cancel` if you are really worried about mod compatibility.

maximumpower55: Oh crap lmao I didn't read the `And just change the item arg to render an empty itemstack`, oopsie doopsie will change it now!

library/item/item_group/src/main/java/org/quiltmc/qsl/item/group/mixin/client/CreativeInventoryScreenMixin.java
lukebemish: No, not a local capture. Just capture the method args, right?

maximumpower55: ah

library/item/item_group/src/main/java/org/quiltmc/qsl/item/group/api/client/ItemGroupIconRenderer.java
maximumpower55: Resolved.

library/item/item_group/src/main/java/org/quiltmc/qsl/item/group/mixin/client/CreativeInventoryScreenMixin.java
maximumpower55: Resolved.

library/item/item_group/src/main/java/org/quiltmc/qsl/item/group/mixin/client/CreativeInventoryScreenMixin.java
maximumpower55: Resolved.

EnnuiL (changes requested):
A quick glance points to a need for a Checkstyle pass; Do that, and I'll continue the rest of my review

library/item/item_group/src/main/java/org/quiltmc/qsl/item/group/impl/ItemGroupIconRendererRegistry.java
```diff
@@ -0,0 +1,45 @@
...
+import net.minecraft.item.ItemGroup;
+
+@ApiStatus.Internal
+@Environment(EnvType.CLIENT)
+public final class ItemGroupIconRendererRegistry {
+	private static final Map<ItemGroup, ItemGroupIconRenderer<? extends ItemGroup>> RENDERERS = new Object2ObjectOpenHashMap<>();
+
+	@SuppressWarnings("unchecked")
+	public static <IG extends ItemGroup> ItemGroupIconRenderer<IG> register(IG itemGroup, ItemGroupIconRenderer<IG> renderer) {
+		return (ItemGroupIconRenderer<IG>) RENDERERS.put(itemGroup, (ItemGroupIconRenderer<? extends ItemGroup>) renderer);
```
lukebemish:
```java
(ItemGroupIconRenderer<? extends ItemGroup>) renderer
```
This cast isn't necessary

maximumpower55: Actually it is, it won't compile without it on my IDE

lukebemish: If that's the case, that seems like a bug in your IDE. Since `IG` extends `ItemGroup`, a parameter of `IG` is already a `? extends ItemGroup`. However, no harm in leaving it in, I guess, though it's a bit confusing to read.

maximumpower55: wait hm, it does compile now for some reason, my IDE is haunted I guess perfect for spooky month!

maximumpower55:
> A quick glance points to a need for a Checkstyle pass; Do that, and I'll continue the rest of my review

Forgot to say this yesterday but the hopefully only needed once checkstyle pass has been pushed.

EnnuiL (approved):
It seems good! However, yeah, I'll want more pairs of eyes on this, or at least one that's more knowledgeable on this than me :P

OroArmor (approved):

Mysterious-Dev (approved):

PR #203, opened 2022-10-16
Update GitHub Actions + two extra libraries.

Kichura:
I don't think i need to explain this since the title says everything, Although edits are welcomed just in case something goes wrong.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,7 +2,7 @@ name: Release
 on: [workflow_dispatch] # Manual trigger
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
         with:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,7 +5,7 @@ jobs:
   build:
     strategy:
       matrix:
-        java: [17, 19]
+        java: [17, 18]
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,9 +5,8 @@ jobs:
   build:
     strategy:
       matrix:
-        # Keep the matrix for when jdk 17 rc come out
-        java: [17, 18]
-    runs-on: ubuntu-20.04
+        java: [17, 19]
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
         with:
--- a/build-logic/build.gradle
+++ b/build-logic/build.gradle
@@ -20,7 +20,7 @@ repositories {
 
 dependencies {
 	implementation "org.quiltmc:loom:1.0.+"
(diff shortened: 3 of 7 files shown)


.github/workflows/publish.yml
```diff
@@ -2,7 +2,7 @@ name: Release
 on: [workflow_dispatch] # Manual trigger
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
EnnuiL, line 5: Hm, despite this being 100% a change that I'd do, I'm conflicted between this and using `ubuntu-latest` for future-proofing, even though iirc, it's still at 20.04

.github/workflows/test.yml
```diff
@@ -5,9 +5,8 @@ jobs:
   build:
     strategy:
       matrix:
-        # Keep the matrix for when jdk 17 rc come out
-        java: [17, 18]
-    runs-on: ubuntu-20.04
+        java: [17, 19]
```
EnnuiL: I'd keep it at 18 until a new Gradle release supporting Java 19 appears, since yeah, gradle's kinda the dealbreaker of the newer Java versions and ughhhhh i hate that so much, but unfortunately, that's life

build-logic/build.gradle
```diff
@@ -20,7 +20,7 @@ repositories {
 
 dependencies {
 	implementation "org.quiltmc:loom:1.0.+"
-	implementation "net.kyori:indra-git:3.0.0"
+	implementation "net.kyori:indra-git:3.0.1"
```
EnnuiL, line 23: whaaaaaaaaa??? wdym they updated already?
(dismiss this, it's just an observation :P)

EnnuiL (changes requested):
Seems good! Just a few changes that I ask

.github/workflows/publish.yml
Kichura: GitHub's team still relays on 20.04 LTS when using the "latest" term - so i just thought of upstreaming this if this is even okay for this including rest of libraries that i planned to PR.

.github/workflows/test.yml
LambdAurora: Even then, I would just add 19 as a separate item.

.github/workflows/publish.yml
EnnuiL: ...actually, let's keep it at 22.04 and then move to latest once the rollout is done

.github/workflows/test.yml
Kichura: Undoing that then.

build-logic/build.gradle
Kichura: Everything gets updated faster than people can blink lol.

.github/workflows/test.yml
EnnuiL: meh, I would drop Java 18 as soon as possible

.github/workflows/publish.yml
Kichura: Alright i gotcha.

EnnuiL (approved):
We'll only need testing now, but yeah, this seems good!

PR #206, opened 2022-10-24
Signal Channels

Cypher121:
Provides an API for publishing and subscribing to events. Supports positional (origin+range, box), world, and global events, and positional, (block)entity-bound, and worldwide/global listeners.

Current implementation is a Proof-of-Concept and has no global listener support yet, as the listener subscriptions need to be heavily re-organized (either per-chunk storage or a [data structure with quick access for neighboring coordinates](https://cs.stackexchange.com/questions/148036/what-are-good-data-structure-algorithm-for-fast-3d-coordinates-search)).

The test mod has a regeneration block that emits events within 4 blocks of itself. Events are caught by player listeners and apply Regen I.


library/misc/signal_channel/src/main/java/org/quiltmc/qsl/signal_channel/impl/WorldSubscriptionManager.java
```diff
@@ -0,0 +1,116 @@
...
+		return subscriptions.stream()
+				.filter(s ->
+						s.getTracker() == null ||
+								positionFilter.test(s.getTracker().getPosition())
+				).toList();
+	}
+
+	public List<SubscriptionImpl.Named<?>> getNamedSubscriptions(Identifier id) {
+		cleanSubscriptionMaps();
+		return namedSubscriptions.computeIfAbsent(id, k -> new ArrayList<>()); //TODO change raw gets to computes!
```
SilverAndro, line 76: > //TODO change raw gets to computes!

Is this outdated?

SilverAndro:
Im a bit confused/concerned about the passing of `Codec<T>` everywhere, what problem is this trying to solve?

Named channels seems to already have an ID, so there must be some AoT communication about what data to expect, right?

SilverAndro (commented):
Also this minor thing

library/misc/signal_channel/src/main/java/org/quiltmc/qsl/signal_channel/impl/WorldSubscriptionManager.java
Cypher121: it is, yeah

Cypher121:
Every instance of the named channel given by `SignalChannel.getOrCreateNamed` gets its own codec but can share the ID with others. ID determines who gets it, and it's then re-encoded from the source codec to the target.

The intent of named channels is to be a way of communication between mods without a dependency. Some lightweight compat may not call for an optional dependency with all its pitfalls, so instead a named channel with the same ID as the other mod is created and subscribed to. The original type is not available, so a new one with a compatible Codec is created and used as the message type itself.

The conversion (kinda scuffed, it definitely should cache both codec steps) is in SubscriptionImpl.Named I believe.

lukebemish:
How often is that codec based conversion likely to fire, in practice?

Cypher121:
For each event fired, once for every unique subscribing codec, plus possibly an LRU cache on top for repeated identical events. So realistically once or twice per event at most? It's also not likely to be a big object to serialize to begin with.

Outside of better caching, the part I'm most unhappy about with this one is converting it through JSON, which is a lossy format.

library/misc/signal_channel/src/main/resources/quilt_signal_channel.accesswidener
```diff
@@ -0,0 +1,3 @@
+accessWidener v2 named
+
+transitive-accessible   class   net/minecraft/block/entity/BlockEntityType$BlockEntityFactory
```
SilverAndro, line 3: What is this used for? I dont see this class referenced anywhere in your code

Cypher121: Implicitly converted to from a constructor reference in the testmod when creating a BlockEntityType with the builder. Had to be copied from the block_entity module due to transitive AWs not working

SilverAndro: Ah I see, alright

Is it possible to keep it in the testmod resources or does loom throw a fit with that?

SilverAndro:
Taking a bit of a deeper look, it doesnt seem like this currently support switching the bind at all? I can think of a few usages for that (i.e something that switches between block and entity form?) but you can cancel and and create a new subscription so I dont think its too major

As for the storage system, iirc minecraft itself already has something similar that could be reused? theres an `EntityLookup<T extends EntityLike>` that might be possible to build around?

PR #211, opened 2022-10-31
Better dynamic registries setup event

LambdAurora:
DynamicRegistryManager setup event got implement very hastily to get surface rules API back up and running on 22w43a, but after looking at it and having to use it a second time but for the Biome API testmod, I'm definitely not happy with its current design.

So time to take our "snapshot is for experimentation" philosophy and apply it :P
Here's a second iteration of the event, this time with a context object which will make the event body more resistant to breakage.
It also introduces loads of helpers to deal with the DynamicRegistryManager more easily. You can find examples of it being used in the surfaces rules API and biome API testmods.


OroArmor (approved):
Looks great!

EnnuiL (approved):
oh wow, comparing it to the old API? This one is definitely the way to go!

PR #219, opened 2022-11-20
Pass decrement `amount` value through to recipe remainder handler

### LINKED ISSUE #218: Recipe output slot mixins ignore `amount` parameter
If another mod changes the `amount` parameter to take more than 1 item from the input slots, the QSL Recipe Remainder API mixins ignore that and only take 1.

magneticflux-:
Closes #218

This correctly handles the `amount` parameter from the output slot mixins instead of ignoring it. This fixes mods that alter the amount of items consumed by a recipe through mixins (or some potential future API).

### DIFF
--- a/library/item/item_setting/src/main/java/org/quiltmc/qsl/item/setting/api/RecipeRemainderLogicHandler.java
+++ b/library/item/item_setting/src/main/java/org/quiltmc/qsl/item/setting/api/RecipeRemainderLogicHandler.java
@@ -58,15 +58,24 @@ public interface RecipeRemainderLogicHandler {
 	 * Excess items that cannot be returned to a slot are dropped in the world.
 	 *
 	 * @param input the original item stack
+	 * @param amount the amount by which to decrease the stack
 	 * @param recipe the recipe being used
 	 * @param inventory the inventory
 	 * @param index the index of the original stack in the inventory
 	 * @param world the world
 	 * @param location the location to drop excess remainders
 	 */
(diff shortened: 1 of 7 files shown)


OroArmor (approved):

magneticflux-:
Confirmed working with my changes to Stonecutter Recipe Tags in https://github.com/TropheusJ/stonecutter-recipe-tags/pull/6

Before crafting:
image
After crafting:
image

magneticflux-:
I found another, broader, issue with the remainder API mixins: `handleRemainderForScreenHandler` doesn't "split" the stack and return the split-off part, it just decrements a stack. The mixins return the current stack, _not_ the split-off stack with 0-to-amount items.

What do you think we should do about this? As far as I can tell, only `StonecutterOutputSlotMixin` is problematic because it uses the size of the split stack to determine if something was crafted.

Should we:
1. refactor `handleRemainderForScreenHandler` _even more_
    1. to either return the split stack OR
    2. to return the size that the stack would be
2. write some hack that just makes an `amount`-size stack to return, regardless of how many were actually consumed

I'm going to implement option 2 for now since that's the easiest thing that works, but it might break in the future.

EnnuiL (approved):
I would be interested on Approach 1, we could break things for 1.19.3 after all; But yeah, for now? This looks good!

TheGlitch76 (approved):

LambdAurora (approved):

OroArmor (approved):

PR #222, opened 2022-11-26
Change Multipart Entity API's "rotate" methods to work as intended

Arathain:
This PR changes the rotate() methods of AbstractMultipartEntity to rotate the vectors via pitch and yaw only, which also fixes the bug where pitch and yaw rotations would not stack (as result of using quaternions, presumably).
It addionally adds the option to switch between degrees and radians without having to use *(float)Math.PI/180f every time.

### DIFF
--- a/library/entity/multipart/src/main/java/org/quiltmc/qsl/entity/multipart/api/AbstractEntityPart.java
+++ b/library/entity/multipart/src/main/java/org/quiltmc/qsl/entity/multipart/api/AbstractEntityPart.java
@@ -184,9 +184,9 @@ public abstract class AbstractEntityPart<E extends Entity> extends Entity implem
 	 * @param pitch the rotation about x-axis in degrees
 	 * @param yaw   the rotation about y-axis in degrees
 	 */
-	public void rotate(Vec3d pivot, float pitch, float yaw) {
+	public void rotate(Vec3d pivot, float pitch, float yaw, boolean degrees) {
 		this.setPivot(pivot);
-		this.rotate(pitch, yaw, true);
+		this.rotate(pitch, yaw, degrees);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


EnnuiL (changes requested):
I'll make an exception to my break to point something: Could you readd the removing methods for 1.19.2? Breaking compatibility might not be a great idea despite it being the kind of fundamental change that is needed *but* 1.19.3 is coming soon; I'll want to fast-track this PR so I can already remove the deprecated functions before 1.19.3 releases

Arathain:
Like so?

library/entity/multipart/src/main/java/org/quiltmc/qsl/entity/multipart/api/AbstractEntityPart.java
```diff
@@ -181,30 +181,42 @@ public void move(double dx, double dy, double dz) {
...
 		this.move(transformedPos);
 	}
+	/**
+	 * @deprecated  Use {@link #rotate(float, float, boolean)} instead.
+	 */
+	@Deprecated
+	public void rotate(float pitch, float yaw, float roll) {
+		rotate(pitch, yaw, false);
+	}
+	/**
```
EnnuiL:
```suggestion
	}

	/**
```

library/entity/multipart/src/main/java/org/quiltmc/qsl/entity/multipart/api/AbstractEntityPart.java
```diff
@@ -181,30 +181,42 @@ public void move(double dx, double dy, double dz) {
...
-		var transformedPos = this.getAbsolutePivot().subtract(this.getAbsolutePosition())
-				.add(relativePos.getX(), relativePos.getY(), relativePos.getZ());
-
+	public void rotate(float pitch, float yaw, boolean degrees) {
+		Vec3d rel = this.getAbsolutePosition().subtract(this.getAbsolutePivot());
+		rel = rel.rotateX(-pitch * (degrees ? (float)Math.PI/180f : 1)).rotateY(-yaw * (degrees ? (float)Math.PI/180f : 1));
+		Vec3d transformedPos = this.getAbsolutePivot().subtract(this.getAbsolutePosition()).add(rel);
 		this.move(transformedPos);
 	}
+	/**
```
EnnuiL:
```suggestion
	}

	/**
```

EnnuiL (changes requested):
Minor nitpicks, but yep! Like that! But don't worry, we'll quickly get rid of those on 1.19.3

Arathain:
alright, done

EnnuiL (approved):

LambdAurora (approved):

library/entity/multipart/src/main/java/org/quiltmc/qsl/entity/multipart/api/AbstractEntityPart.java
```diff
@@ -181,29 +181,43 @@ public void move(double dx, double dy, double dz) {
 	 * Rotates this {@link AbstractEntityPart} about the pivot point with the given rotation.
 	 *
 	 * @param pivot the pivot point to rotate about in relative coordinates
-	 * @param pitch the rotation about z-axis in degrees
+	 * @param pitch the rotation about x-axis in degrees
 	 * @param yaw   the rotation about y-axis in degrees
```
williambl: Javadoc is inaccurate.

```suggestion
	 * @param pitch   the rotation about x-axis
	 * @param yaw     the rotation about y-axis
	 * @param degrees whether the rotation should be done in degrees or radians
```

williambl (changes requested):
Just one nitpick.

PR #227, opened 2022-12-04
Fix typo in RegistrySync that sends and receives RESTORE for APPLY.

Ampflower:
Note: This requires a version bump and backward compatibility code to avoid breaking the client by the server not sending expected data. This won't break the server as the server is an authority, and won't require either the server or client to update as is with the compatibility code in place.

Tested with an older QSL server and client without this patch, and registry syncing still functions as intended.


LambdAurora (approved):
Note to the merger: support for version 1 needs to be removed for 1.19.3.

EnnuiL (approved):

PR #228, opened 2022-12-04
Fix ItemColors not being rebuilt on registry sync.

### LINKED ISSUE #225: Inconsistent behaviour of registry sync on client from Fabric API
*Carry-over from VazkiiMods/Botania#4234, as Botania exposed symptoms of a bad registry sync.*

While diagnosing a bug that causes Botania to apply the wrong renderer to various items, it seems that the registry sync module in Quilt is inconsistent with the registry sync module of Fabric in terms of how it deals with items.

## Steps to Reproduce
1. Install Botania + QSL + dependencies on a server.
2. Install Botania + QSL + Endermantic Overhaul + dependencies on a client.
3. Try to join the server with purposefully desynced registries.

## Expected Results
 - The Wand of the Forest to have the petal colours you gave it instead of white.
 - Mana mirrors when linked to filled pools to have a light blue center, same colour as mana, instead of white.
   - This may also apply to mana tablets and mana bands.
 - Black Hole Talismans to be pink, white and black, not be tinted dark blue.
 - Lenses, mana guns, petals, manasteel, runes to not crash the client by bad renderers.

All of the above are symptoms of the underlying item object being separated from the item ID.

## Actual Results
 - The backing item object seems to be severed from the item ID and subsequently shuffled to where it shouldn't be.
 - The opposite of expected results, instead having crashes with select items, incorrect application of renderers.

## Notes
 - When removing Endermantic Overhaul but have any other mod combination on the client, Botania behaves as expected.
 - Replacing QSL with FAPI on the client while keeping Botania, Endermantic Overhaul and Quilt Loader also allows Botania to behave correctly on the client despite only Botania and QSL on the server.
 - ~~This may be caused by Endermantic Overhaul messing with Fabric API internals as well and causing this issue to propagate on the client.~~ Update: This doesn't seem to be the case.
 - So far, no other mod has experienced symptoms or caused unexpected internal shuffling of the registry.
 - ~~It has yet been determined if Endermantic Overhaul has anything in regards to modifying the API when it shouldn't. Will look through the code to see what's going on.~~ Update: It doesn't appear that Endermantic Overhaul does anything that it shouldn't with regards to the Fabric API, it just initialises blocks at bootstrap and mixins solely into Minecraft at safe spots.

## Minimal mod list:
 - Minecraft 1.19.2
 - Botania 1c28212604ff5cca3e2bf25abfc7e6c682f889d4
 - [Endermantic Overhaul](https://modrinth.com/mod/endermantic-overhaul) 3.0.0 or 2.0.0, only on client.
 - Quilt Standard Libraries 3.0.0-beta.22 + Quilted Fabric API 4.0.0-beta.23 (Based on 0.68.0+1.19.2) on both the client and server.
 - Quilt Loader 0.18.1-beta.21
 - Fabric API 0.68.0+1.19.2, replacing QSL on the client as a reference.

Ampflower:
Fixes QuiltMC/quilt-standard-libraries#225

### DIFF
--- a/library/core/registry/src/main/java/org/quiltmc/qsl/registry/mixin/client/ItemRendererAccessor.java
+++ b/library/core/registry/src/main/java/org/quiltmc/qsl/registry/mixin/client/ItemRendererAccessor.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2022 QuiltMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.quiltmc.qsl.registry.mixin.client;
+
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Accessor;
+
+import net.minecraft.client.color.item.ItemColors;
+import net.minecraft.client.render.item.ItemRenderer;
+
+/**
+ * Allows for {@link ItemColors} to be resynchronised with the registry, fixing the bug
+ * of the incorrect color renderer being applied to items due to the backing raw identifiers
+ * being shifted.
+ *
+ * @author KJP12
+ * @since 3.0.0
+ */
+@Mixin(ItemRenderer.class)
+public interface ItemRendererAccessor {
+	@Accessor
+	ItemColors getColors();
+}
--- a/library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/client/ClientRegistrySync.java
+++ b/library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/sync/client/ClientRegistrySync.java
@@ -200,7 +200,7 @@ public final class ClientRegistrySync {
 		((RebuildableIdModelHolder) models).quilt$rebuildIds();
 		models.reloadModels();
 
-		var itemColors = ((ItemRendererAccessor) client.getItemRenderer()).getItemColors();
+		var itemColors = ((ItemRendererAccessor) client.getItemRenderer()).getColors();
 		((RebuildableIdModelHolder) itemColors).quilt$rebuildIds();
 	}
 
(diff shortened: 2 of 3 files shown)


library/core/registry/src/main/java/org/quiltmc/qsl/registry/mixin/client/ItemRendererAccessor.java
```diff
@@ -0,0 +1,21 @@
...
+import net.minecraft.client.render.item.ItemRenderer;
+
+/**
+ * Allows for ItemColors to be resynchronised with the registry, fixing the bug
+ * of the incorrect color renderer being applied to items due to the backing IDs
+ * being shifted.
+ *
+ * @author KJP12
+ * @since 3.0.0
+ **/
```
LambdAurora:
```suggestion
/**
 * Allows for {@link ItemColors} to be resynchronised with the registry, fixing the bug
 * of the incorrect color renderer being applied to items due to the backing raw identifiers
 * being shifted.
 *
 * @author KJP12
 * @since 3.0.0
 */
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/mixin/client/ItemRendererAccessor.java
```diff
@@ -0,0 +1,21 @@
...
+@Mixin(ItemRenderer.class)
+public interface ItemRendererAccessor {
+	@Accessor("colors")
+	ItemColors getItemColors();
```
LambdAurora: Why isn't this called `getColors` to respect the accessor conventions?

LambdAurora (changes requested):

library/core/registry/src/main/java/org/quiltmc/qsl/registry/mixin/client/ItemRendererAccessor.java
Ampflower: Fair, I'll go ahead and change it to be just `getColors`.

Ampflower:
*Forgot about the extra asterisk, removed it now.*

library/core/registry/src/main/java/org/quiltmc/qsl/registry/mixin/client/ItemRendererAccessor.java
```diff
@@ -0,0 +1,21 @@
+package org.quiltmc.qsl.registry.mixin.client;// Created 2022-04-12T11:27:59
```
LambdAurora:
```suggestion
package org.quiltmc.qsl.registry.mixin.client;
```

And please run `gradle :core:registry:applyLicenses`

Ampflower: Oh, forgot about the timestamp. Seems I need to remove that for QSL in case I ever need to PR new classes lol

LambdAurora (approved):

EnnuiL (approved):

PR #230, opened 2022-12-08
Debug Renderer API

williambl:
This is an API for Debug Renderers (`DebugRenderer`)

# Motivation

Vanilla provides a few Debug Renderers, ~all of which are disabled:
 - pathfinding viewer
 - water debug
 - chunk borders
 - heightmap viewer
 - brain debug
 - etc.

The `DebugRenderer` class provides various utility methods, and custom Debug Renderers can be created by implementing `DebugRenderer.Renderer`.

I've been using this occasionally in fabric mods, just by bridging the `SimpleDebugRenderer` to a `WorldRenderEvents.Last` and then conditionally registering that depending on a public static final boolean. This isn't *ideal*. In addition, turning on the vanilla Debug Renderers has to be done either with mixins or manually running them in a render event.

Many vanilla Debug Renderers require extra data to be sent by the server: these are sent through custom payload packets (see the IDs in `CustomPayloadS2CPacket` which start with `minecraft:debug/`). These are sent in various static methods in `DebugInfoSender`. However, most of these methods have their contents stripped out.

# Content

Therefore I am making a Debug Renderer API, which adds the following:
 - [x] `DebugFeature` class, with methods:
```java
Identifier id();
// whether it needs to be enabled on the server in order to render on client
boolean needsServer();
// is enabled locally
boolean isEnabled();
// used on server
boolean isEnabledOnServerAndClient(ServerPlayerEntity player);
@ClientOnly
boolean isEnabledOnServerAndClient();
@ClientOnly
boolean shouldRender();
```

- [x] Debug Feature registry/factory (common-side), to which IDs can be added (+ a boolean which determines if it needs to be enabled on both client and server to render), and it creates a `DebugFeature`. (`DebugFeature myDebugFeature = DebugFeatureRegistry.register(new Identifier("my_mod:test"), true)`)
- [x] Debug Renderer registry (`DebugRendererRegistry.register(myDebugFeature, MyDebugRenderer::new)`)
- [x] Debug Features are enabled/disabled individually with a command `/debug enable x:x`, and on client with a client command `/debug_client enable x:x`
- [x] Debug Renderers only run if the corresponding Feature is enabled (and, if the feature requires that it is enabled on the server too, it checks that)
- [x] We register Debug Features for each of the vanilla Debug Renderers
- [ ] We re-implement the missing methods in `DebugInfoSender`, checking that the corresponding Debug Feature is active on server and only sending the packets to clients with it also active


library/misc/debug_renderers/src/main/java/org/quiltmc/qsl/debug_renderers/api/DebugFeature.java
```diff
@@ -0,0 +1,76 @@
...
+	public boolean isEnabled() {
+		return DebugFeaturesImpl.isEnabled(this);
+	}
+
+	public boolean isEnabledOnServerAndClient(ServerPlayerEntity player) {
+		return DebugFeaturesImpl.isEnabled(this) && DebugFeaturesImpl.isEnabledForPlayer(player, this);
+	}
+
+	@ClientOnly
+	boolean isEnabledOnServerAndClient() {
```
TheGlitch76: I don't like having an overload that is client-only. This would be really easy to use on accident

williambl: True - come to think of it, the client overload isn't needed at all, and can just be folded into `shouldRender`.

williambl:
Some vanilla debug renderers enabled with this API:

williambl:
Other than re-implementing some missing MC methods, this is now fully functional.

Redfan2:
What is the state of this?

OroArmor:
Definitely stalled, but I was looking into it recently and it seems fairly interesting and not too bad to update

PR #231, opened 2022-12-09
Switch to using loader.api.CachedFileSystem rather than loader internals.

AlexIIL:
Since `JoinedFileSystem` might be used for read-only mods in the future, and mutable mods might not use `JoinedFileSystem` - however calling [`CachedFileSystem`](https://github.com/QuiltMC/quilt-loader/blob/develop/src/main/java/org/quiltmc/loader/api/CachedFileSystem.java#L21-L26).[`isPermanentlyReadOnly()`](https://github.com/QuiltMC/quilt-loader/blob/develop/src/main/java/org/quiltmc/loader/api/CachedFileSystem.java#L28-L33) will ensure mutable mods never get cached, and immutable mods will always get cached.

### DIFF
--- a/build-logic/src/main/java/qsl/internal/Versions.java
+++ b/build-logic/src/main/java/qsl/internal/Versions.java
@@ -43,7 +43,7 @@ public final class Versions {
 	/**
 	 * The version of Quilt Loader to use.
 	 */
-	public static final String LOADER_VERSION = "0.17.6";
+	public static final String LOADER_VERSION = "0.17.7";
 
 	/**
 	 * The target Java version.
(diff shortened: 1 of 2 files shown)


AlexIIL:
Oh, and I should probably mention:
 1. this bumps the minimum loader version to `0.17.7` *or* `0.18.1-beta.20` - beta 19 and before don't have this API.
 2. I haven't tested this yet :/ (Hopefully this doesn't break anything :p)

EnnuiL (approved):
Hmm, I wonder if we should merge it soon or wait a little more; Either way, we do need to eventually merge this

TheGlitch76:
This looks ok to merge soon, as it's expected that betas can break in mysterious ways.

TheGlitch76 (approved):

LambdAurora (approved):

EnnuiL:
Yeah, this seems FCPable; I believe that QLoader 0.17.8 should be more common these days anyway, and yeah, it should be safe to bump to it

Leo40Git (approved):

PR #259, opened 2023-01-29
Add TAWs for non-public vanilla item constructors

EnnuiL:
This addresses the lack of TAWs for item constructors that I noticed with [QFAPI's issue #62](https://github.com/QuiltMC/quilted-fabric-api/issues/62); I genuinely thought that we already had those ;-;

So, this PR grabs the Quilt Block Extensions module's AW generator, adapts it to items, and puts it on the Item Extension module, although I do wonder if this should be on the Quilt Item Setting API instead...?

This adds a testmod for testing the AW; i wish i could use the credits music ;-;


EnnuiL:
Closed in favor of PR #307

PR #260, opened 2023-01-29
Add REA filters

### LINKED ISSUE #255: Make an entry filter for REAs
Right now REAs blindly accept any entry given to them, a way to filter out these values would be useful.

OroArmor:
Fixes #255 

Very simple API


library/data/registry_entry_attachment/src/main/java/org/quiltmc/qsl/registry/attachment/api/RegistryEntryAttachment.java
```diff
@@ -501,6 +513,17 @@ public Builder<R, V> defaultValueProvider(@Nullable DefaultValueProvider<R, V> d
 			return this;
 		}
 
+		/**
+		 * Sets the filter for the attachment.
```
Leo40Git:
```suggestion
		 * Sets the registry entry filter for the attachment.
```

library/data/registry_entry_attachment/src/main/java/org/quiltmc/qsl/registry/attachment/api/RegistryEntryAttachment.java
```diff
@@ -287,6 +288,15 @@ default Optional<V> get(R entry) {
 	 */
 	boolean remove(TagKey<R> tag);
 
+	/**
+	 * The entry filter can prevent entries from being shown in the attachment.
```
Leo40Git:
```suggestion
	 * Gets the entry filter that determines if certain registry entries can be associated with values in this attachment.
```

Leo40Git (approved):
LGTM, apart from minor Javadoc nitpicks

LambdAurora (commented):
Absolutely not the API I expected tbh.

I think I'd rather call this "condition" or something.

I'd rather see the filter concept applied to the entry logic, as-in being able to define:
```json5
{
    id: "#minecraft:wood",
    filter: [
        "!#minecraft:non_burnable_wood"
    ]
}
```

Or something in the line, this could also be the basis of the future "copy" system or something.

Anyway, this system is also fine despite not serving the same purpose as my idea.

library/block/block_content_registry/src/main/java/org/quiltmc/qsl/block/content/registry/api/BlockContentRegistries.java
```diff
@@ -105,6 +105,7 @@ public class BlockContentRegistries {
 
 						return DataResult.success(block);
 					}))
+			.filter(block -> block.getDefaultState().contains(Properties.AXIS))
```
LambdAurora: Just noticed this.

While I like the idea, I also find it a bit bad, thought that comes from strippable being kind of awful from the start.
Allowing other kind of blocks to be strippable would be interesting but that's a debate for another PR. But it shows that the conditions system has potential.

TheGlitch76:
This PR needs to be updated to 1.19.4

OroArmor:
Updated to 1.20

library/data/registry_entry_attachment/src/main/java/org/quiltmc/qsl/registry/attachment/api/RegistryEntryAttachment.java
```diff
@@ -287,6 +288,26 @@ default Optional<V> get(R entry) {
 	 */
 	boolean remove(TagKey<R> tag);
 
+	/**
+	 * The entry validation can prevent entries from being shown in the attachment.
```
TheGlitch76, line 292: "Being shown" is misleading; it implies that this validator will silently drop entries, but you seem to throw a hard exception when an entry that fails the predicate is added:

https://github.com/QuiltMC/quilt-standard-libraries/blob/c8e05fbe16d52e2daf7df8d80743726060f57167/library/data/registry_entry_attachment/src/main/java/org/quiltmc/qsl/registry/attachment/impl/RegistryEntryAttachmentImpl.java#L197-L199

TheGlitch76 (changes requested):

OroArmor:
We will need this for 1.21.2 and the new fuel system. I finally though of a good way of doing it, and ill update the pr soon.

PR #261, opened 2023-01-29
Remove Unneccessary Brackets

halotroop2288:
This bug may have been repeated countless times, but here's a fix for just the first occurrence I noticed.

Fixes this JavaDoc rendering bug:
JavaDoc bug


LambdAurora:
Those brackets are actually intended.

[It renders absolutely fine with the official JavaDoc tool](https://maven.quiltmc.org/repository/release/org/quiltmc/qsl/4.0.0-beta.10+1.19.3/qsl-4.0.0-beta.10+1.19.3-fat-javadoc.jar/org/quiltmc/qsl/screen/api/client/QuiltScreen.html).

This is a new feature of JavaDoc 16 which was added [here](https://github.com/openjdk/jdk/pull/1355).

This is, actually, an Intellij bug, which means such bug report should be reported to JetBrains instead.

PR #264, opened 2023-02-04
API for datapackable event callbacks (and datapackable surface rule modifications as an example usage)

lukebemish:
This PR provides a general API for making arbitrary events datapack-controllable. The API is designed as follows:

- Event callbacks that should be datapackable extend `CodecAware`; specific callback subclasses that can be used in a datapack should override `getCodecIdentifier` and return a non-null value
- Identifiers with corresponding codecs should be registered to a `CodecMap` accompanying the event; the identifiers provided here are the ones that should be provided in `CodecAwareCallback#getCodecIdentifier`
- For each datapackable event, a `DynamicEventCallbackSource` should be provided alongside the `Event`. This object allows for two things:
  - First, it collects and applies all the callbacks loaded from data
  - Second, it allows mod authors to register their own callbacks alongside an identifier, which allows datapacks to override those provided callbacks
- This `DynamicEventCallbackSource` should be used to load data; it has a method which can accept a ResourceManager and updates the dynamic entries, which coould be called either during registry load (for stuff like surface rules) or with a reload listener.

An example of the use of this to make an event datapack-able has been provided for the surface rule events; I think there are likely some other candidate events in QSL at present as well. I would love some feedback on the general structure of this API, as well as names; I'll try to go through and add documentation as I have free time.

This API can also be used for systems not based on events; an example of this can be found in my changes to biome modifiers. This I'm a bit dicier on my implementation of, and could definitely use some feedback; it's mostly just difficult to design this sort of API around biome modification the way quilt implements it.


lukebemish:
I'm thinking, for this PR, I'll leave it at implementations for surface rules and biome modification; they provide a good example of both how to attach this API to something event based and how it can be used without events, respectively. I'll try to go through and document everything when I have some spare time; however, I am wondering about a couple of things:
- What module/package should stuff (`CodecMap`, `CodecAwareCallback`, `DynamicEventCallbackSource`) go in?
- Does anyone have a better name for `CodecAwareCallback` (or, heck, anything else), especially since it can be used outside the context of events?
- What "default" biome modifiers or surface rule callbacks should QSL provide? Currently, I've implemented rules to add a material rule at the beginning or end of a surface rule (these could probably be merged into a single callback type), as well as rules to add and remove features from biomes. I'm sure there's a number of other common ones that would be worth implementing - this could be saved for a future PR as well.

lukebemish:
I renamed `CodecAwareCallback` to `CodecAware` (which then makes something like `CodecAwarePredicate` make more sense). Documentation... *exists*, though at this point in time I don't know if it's any good. I'm trying to get some default biome modifier and surface rule callback types put together; I'll probably want to write some more extensive test mods while I'm at it, including test mods of the whole "replace code-registered stuff" side of things

lukebemish:
It still definitely needs a ton of work but I think I'm as far along as I'd like to go without some more in-depth review. I'm still not sure about two major things:
- which module the API should go into
- how the JSON formats of default callbacks/etc. should be documented

lukebemish:
...I'm not going to worry about that test failure after I updated the PR too much for now

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/event/data/CodecAware.java
```diff
@@ -0,0 +1,35 @@
...
+ * {@link #getCodecIdentifier()} to return the identifier of the codec if they are encodable, or leave the default
+ * implementation if they are not. Identifiers provided should correspond to the identifiers of codecs registered in
+ * some {@link CodecMap}.
+ */
+public interface CodecAware {
+
+	/**
+	 * {@return the identifier of the codec that can be used to encode this object, or {@code null} if this object is not encodable}
+	 */
+	default Identifier getCodecIdentifier() {
```
OroArmor:
```suggestion
	default @Nullable Identifier getCodecIdentifier() {
```

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/event/data/CodecMap.java
```diff
@@ -0,0 +1,127 @@
...
+		return Codec.pair(Codec.either(Identifier.CODEC, Identifier.CODEC.listOf())
+				.xmap(either -> either.map(List::of, list -> list), list -> list.size() == 1
+						? Either.left(list.get(0))
+						: Either.right(list))
+				.optionalFieldOf("phase", List.of(Event.DEFAULT_PHASE)).codec(), this.createDelegatingCodec(descriptor));
+	}
+
+	/**
+	 * Creates a delegating codec based off this map that decodes or encodes objects, storing the codec identifier in
+	 * the "type" field
```
OroArmor:
```suggestion
	 * Creates a delegating codec based off this map that decodes or encodes objects, storing the codec identifier in
	 * the "type" field.
```

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/event/data/predicate/NonePredicate.java
```diff
@@ -0,0 +1,46 @@
...
+package org.quiltmc.qsl.base.api.event.data.predicate;
+
+import com.mojang.serialization.Codec;
+
+import net.minecraft.util.Identifier;
+
+/**
+ * A predicate that is never true.
+ */
+public final class NonePredicate<T> implements CodecAwarePredicate<T> {
```
OroArmor: Maybe `NeverPredicate`?

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/event/data/predicate/AllPredicate.java
```diff
@@ -0,0 +1,46 @@
...
+package org.quiltmc.qsl.base.api.event.data.predicate;
+
+import com.mojang.serialization.Codec;
+
+import net.minecraft.util.Identifier;
+
+/**
+ * A predicate that is always true.
+ */
+public final class AllPredicate<T> implements CodecAwarePredicate<T> {
```
OroArmor: Maybe `AlwaysPredicate`?

library/core/qsl_base/src/main/java/org/quiltmc/qsl/base/api/event/data/predicate/PredicateCodecMap.java
```diff
@@ -0,0 +1,115 @@
...
+import com.google.common.collect.HashBiMap;
+import com.mojang.serialization.Codec;
+
+import net.minecraft.util.Identifier;
+import net.minecraft.util.dynamic.Codecs;
+
+import org.quiltmc.qsl.base.api.event.data.CodecMap;
+
+/**
+ * A CodecMap for {@link CodecAwarePredicate}s that bundles a set of codices created by shared providers. This class
```
OroArmor: codices?

library/worldgen/biome/src/main/java/org/quiltmc/qsl/worldgen/biome/api/codec/AddSpawnersModifier.java
```diff
@@ -0,0 +1,63 @@
...
+								  List<SpawnSettings.SpawnEntry> spawners,
+								  Optional<SpawnGroup> group) implements BiomeModifier {
+	public static final Codec<AddSpawnersModifier> CODEC = RecordCodecBuilder.create(instance -> instance.group(
+			BiomeModifier.BIOME_SELECTOR_CODEC.fieldOf("selector").forGetter(AddSpawnersModifier::selector),
+			Codec.either(SpawnSettings.SpawnEntry.CODEC, SpawnSettings.SpawnEntry.CODEC.listOf())
+					.xmap(either -> either.map(List::of, list -> list), list -> list.size() == 1 ? Either.left(list.get(0)) : Either.right(list))
+					.fieldOf("spawners").forGetter(AddSpawnersModifier::spawners),
+			SpawnGroup.CODEC.optionalFieldOf("group").forGetter(AddSpawnersModifier::group)
+	).apply(instance, AddSpawnersModifier::new));
+	public static final Identifier IDENTIFIER = new Identifier("quilt", "add_spawners");
```
OroArmor: Have a consistent order of the identifier and codec, i would go with identifier first

library/worldgen/biome/src/main/java/org/quiltmc/qsl/worldgen/biome/api/codec/AddSpawnersModifier.java
```diff
@@ -0,0 +1,63 @@
...
+import org.quiltmc.qsl.worldgen.biome.api.BiomeModifier;
+import org.quiltmc.qsl.worldgen.biome.api.BiomeSelectionContext;
+
+public record AddSpawnersModifier(CodecAwarePredicate<BiomeSelectionContext> selector,
+								  List<SpawnSettings.SpawnEntry> spawners,
+								  Optional<SpawnGroup> group) implements BiomeModifier {
+	public static final Codec<AddSpawnersModifier> CODEC = RecordCodecBuilder.create(instance -> instance.group(
+			BiomeModifier.BIOME_SELECTOR_CODEC.fieldOf("selector").forGetter(AddSpawnersModifier::selector),
+			Codec.either(SpawnSettings.SpawnEntry.CODEC, SpawnSettings.SpawnEntry.CODEC.listOf())
+					.xmap(either -> either.map(List::of, list -> list), list -> list.size() == 1 ? Either.left(list.get(0)) : Either.right(list))
```
OroArmor: Would a helper method to generate a codec that is either a list or a single value be useful? ive seen this code in a couple places.

library/worldgen/surface_rule/src/main/java/org/quiltmc/qsl/worldgen/surface_rule/api/SurfaceRuleEvents.java
```diff
@@ -46,6 +49,25 @@ public final class SurfaceRuleEvents {
...
 	});
+	/**
+	 * A {@link CodecMap} for the {@link OverworldModifierCallback} event. Can be used to register codecs for loading event callbacks from datapacks.
+	 */
+	public static final CodecMap<OverworldModifierCallback> MODIFY_OVERWORLD_CODECS = new CodecMap<>(context -> {});
+	/**
+	 * A {@link DynamicEventCallbackSource} for the {@link OverworldModifierCallback} event. Can be used to register
+	 * event callbacks alongside an identifier, allowing them to be overridden with datapacks.
+	 */
+	public static final DynamicEventCallbackSource<OverworldModifierCallback> MODIFY_OVERWORLD_IDENTIFIER = new DynamicEventCallbackSource<>(
```
OroArmor: This field name doesnt make that much sense.

OroArmor (changes requested):
Looks really impressive! Just a couple small changes.

OroArmor:
> * which module the API should go into

I think a new module in the `data` library, maybe something like `resource_modification`?

> * how the JSON formats of default callbacks/etc. should be documented

This is a hard one. The wiki would help if it wasn't dead. I do think a `README.md` for the module would be fine since it can explain things in a more textual manner, and also reference other modules using the api.

library/worldgen/biome/src/main/java/org/quiltmc/qsl/worldgen/biome/api/codec/AddSpawnersModifier.java
lukebemish: That's probably not the only helper method I could use; I'll try to pull out any common constructions I've been using.

library/worldgen/surface_rule/src/main/java/org/quiltmc/qsl/worldgen/surface_rule/api/SurfaceRuleEvents.java
lukebemish: Yeah, I was debating the name here. The idea is that it's sorta like the normal event, except callbacks have identifiers; I'm really not sure how best to name it, with that in mind.

lukebemish:
I'm not too sure about `resource_modification` since that's not what this does, really; it just allows for arbitrary data-driven event callbacks through registration of serializers, at its heart. Maybe `codec_callbacks` or something like that?

Should documentation for, say, the built in biome modifier serializers or surface rule callback serializers go in a README file for those modules? I can put some general stuff in a README for the new module but yeah, this is all a bit annoying without the wiki.

lukebemish:
Also, yeah, we need to establish a naming convention for the "identifier-ed event" fields - like, if we have an event `SOME_EVENT`, what should the corresponding identifier-supporting-special-event-callback be called? I'd been doing `SOME_EVENT_IDENTIFIER` or something, but that's terrible. Maybe `SOME_EVENT_IDENTIFIED`? I'd love some ideas

lukebemish:
Alright, I made a few changes. Notably:
- the API itself has been moved to a new module, `data_callbacks`.
- the identifier-aware equivalents of events which this API is enabled for are named `EVENT_NAME_DATA` instead of `EVENT_NAME_IDENTIFIER`; so, for `SurfaceRuleEvents.MODIFY_OVERWORLD`, we now also have `MODIFY_OVERWORLD_CODEC` (for registering serializer codecs) and `MODIFY_OVERWORLD_DATA` (for registering callbacks alongside an identifier, allowing datapacks to replace the callback).

The big thing to do now is to make more test mods for this, and probably some new ones for the API itself not just its uses. I'll try to work on that later this week.

library/worldgen/surface_rule/src/main/java/org/quiltmc/qsl/worldgen/surface_rule/api/SurfaceRuleEvents.java
lukebemish: I think I've got a better alternative now

lukebemish:
Still needs testmods for the biome modification part of it but otherwise it should be good on my end

lukebemish:
Alright, I've got test mods for the biome modification stuff too! This PR is probably ready for some more in depth review at this point.

lukebemish:
Hmm, I just realized that I handle phases two different ways between the biome modification API and the event callback implementation. Just fixed that.

library/core/data_callbacks/src/testmod/java/org/quiltmc/qsl/lifecycle/test/event/DataCallbackTests.java
```diff
@@ -0,0 +1,119 @@
...
+
+		ServerPlayConnectionEvents.JOIN.register((handler, sender, server) ->
+				// Yeah, it's a bit boring, but so it goes...
+				SERVER_JOIN.invoker().onPlayReady(handler, sender, server));
+
+		// This callback is overridden by data and should not fire.
+		JOIN_SERVER_DATA.register(new Identifier(mod.metadata().id(), "after"), new ServerJoinChat("Registered in the after phase from code!", Style.EMPTY), AFTER_PHASE);
+		JOIN_SERVER_DATA.register(new Identifier(mod.metadata().id(), "overridden"), (handler, sender, server) -> {
+			throw new RuntimeException("This callback should have been overridden by data!");
+		});
```
OroArmor: Comment in the wrong place.

library/core/data_callbacks/src/testmod/java/org/quiltmc/qsl/lifecycle/test/event/DataCallbackTests.java
```diff
@@ -0,0 +1,119 @@
...
+				}
+			});
+
+	@Override
+	public void onInitialize(ModContainer mod) {
+		JOIN_SERVER_CODECS.register(ServerJoinChat.IDENTIFIER, ServerJoinChat.CODEC);
+
+		ServerPlayConnectionEvents.JOIN.register((handler, sender, server) ->
+				// Yeah, it's a bit boring, but so it goes...
+				SERVER_JOIN.invoker().onPlayReady(handler, sender, server));
```
OroArmor:
```suggestion
		ServerPlayConnectionEvents.JOIN.register(SERVER_JOIN.invoker()::onPlayReady);
```

OroArmor (changes requested):
Small nitpicks. Really enjoying this API

TheGlitch76 (dismissed):
You need to split your code into `qsl.data.callbacks.api` and `qsl.data.callbacks.impl` like other QSL modules (if the `impl` package is empty thats ok)

lukebemish:
I somehow forgot that; it should be fixed now, assuming nothing else broke

lukebemish:
Erm, whoops, looks like I clicked that button more than once. Sorry for any repeat notifications there

OroArmor (approved):

OroArmor:
Let's merge this! (soon)

library/core/data_callbacks/src/main/java/org/quiltmc/qsl/data/callbacks/api/predicate/CodecAwarePredicate.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.quiltmc.qsl.data.callbacks.api.CodecAware;
+
+/**
+ * A predicate which may be aware of a codec that can be used to encode it.
+ * @param <T> the type of the input to the test
+ */
+public interface CodecAwarePredicate<T> extends CodecAware, Predicate<T> {
+
+}
```
EnnuiL:
```suggestion
public interface CodecAwarePredicate<T> extends CodecAware, Predicate<T> {}
```

library/worldgen/surface_rule/src/main/java/org/quiltmc/qsl/worldgen/surface_rule/api/SurfaceRuleEvents.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 QuiltMC
+ * Copyright 2022, 2023 QuiltMC
```
EnnuiL:
```suggestion
 * Copyright 2022-2023 QuiltMC
```

library/worldgen/surface_rule/src/main/java/org/quiltmc/qsl/worldgen/surface_rule/impl/QuiltSurfaceRuleInitializer.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 QuiltMC
+ * Copyright 2022, 2023 QuiltMC
```
EnnuiL:
```suggestion
 * Copyright 2022-2023 QuiltMC
```

EnnuiL (approved):
Looks pretty good! Just a few nitpicks:

library/core/data_callbacks/src/main/java/org/quiltmc/qsl/data/callbacks/api/CodecAware.java
```diff
@@ -0,0 +1,37 @@
...
+ * {@link #getCodecIdentifier()} to return the identifier of the codec if they are encodable, or leave the default
+ * implementation if they are not. Identifiers provided should correspond to the identifiers of codecs registered in
+ * some {@link CodecMap}.
+ */
+public interface CodecAware {
+
+	/**
+	 * {@return the identifier of the codec that can be used to encode this object, or {@code null} if this object is not encodable}
+	 */
+	default @Nullable Identifier getCodecIdentifier() {
```
EnnuiL: Hm, these many instances of `identifier` and `IDENTIFIER` needs to be shortened to `id`/`ID`, with `ID` variables perhaps needing a prefix

library/core/data_callbacks/src/main/java/org/quiltmc/qsl/data/callbacks/api/predicate/AndPredicate.java
```diff
@@ -0,0 +1,60 @@
...
+
+	private AndPredicate(List<CodecAwarePredicate<T>> values) {
+		this.values = values;
+	}
+
+	private static <T> Codec<AndPredicate<T>> makeCodec(Codec<CodecAwarePredicate<T>> predicateCodec) {
+		return RecordCodecBuilder.create(instance -> instance.group(
+				predicateCodec.listOf().fieldOf("values").forGetter(predicate -> predicate.values)
+		).apply(instance, AndPredicate::new));
+	}
```
EnnuiL: Hm, I feel like returning the `PROVIDER` and having this code be on the `PROVIDER` var creation would be better

library/core/data_callbacks/src/testmod/resources/quilt.mod.json
```diff
@@ -0,0 +1,25 @@
...
+    "metadata": {
+      "name": "Quilt Data Callbacks Test Mod",
+      "license": "Apache-2.0"
+    },
+    "intermediate_mappings": "net.fabricmc:intermediary",
+    "load_type": "always",
+    "entrypoints": {
+      "init": [
+        "org.quiltmc.qsl.lifecycle.test.event.DataCallbackTests"
+      ]
```
EnnuiL:
```suggestion
      "init": "org.quiltmc.qsl.lifecycle.test.event.DataCallbackTests"
```

EnnuiL (changes requested):
oh dear i do have objections

library/core/data_callbacks/src/main/java/org/quiltmc/qsl/data/callbacks/api/predicate/AndPredicate.java
lukebemish: I'm not quite sure what you mean; the provider is just a functional interface referencing things that make codecs. You can't really do this on provider var creation because it can run any number of times, depending on what's being compared with the predicate

library/core/data_callbacks/src/main/java/org/quiltmc/qsl/data/callbacks/api/CodecAware.java
lukebemish: So, just change every `IDENTIFIER` or `identifier` to `ID` or `id`, or maybe `CODEC_ID` or something?

library/core/data_callbacks/src/main/java/org/quiltmc/qsl/data/callbacks/api/predicate/AndPredicate.java
lukebemish: Resolved; see discussion on discord. Basically, this would be a lambda except java doesn't like generics in lambda signatures

library/core/data_callbacks/src/main/java/org/quiltmc/qsl/data/callbacks/api/CodecMap.java
```diff
@@ -0,0 +1,121 @@
...
+ * Relates codecs and identifiers, to be used for a dispatching codec without a full registry. {@link CodecAware} objects
+ * parameterized by the type of this map can be encoded and decoded using the map's codec, and those that are encodable
+ * should provide codec identifiers for a corresponding codec in this map.
+ * @param <T> the codec-aware type that this map dispatches to
+ */
+public class CodecMap<T extends CodecAware> {
+	private final BiMap<Identifier, Codec<? extends T>> codecs = HashBiMap.create();
+
+	public CodecMap() {
+	}
```
EnnuiL:
```suggestion
	public CodecMap() {}
```

library/core/data_callbacks/src/main/java/org/quiltmc/qsl/data/callbacks/api/CodecMap.java
```diff
@@ -0,0 +1,121 @@
...
+
+	/**
+	 * Creates a codec map pre-filled with an operation representing a value that has no effect. For instance, this can
+	 * be used to initialize with an event callback that does nothing, or a biome modification that does not modify the
+	 * biome. This provided value is useful as a value that datapacks can use to disable something.
+	 * @param nullOperation a value that has no effect
+	 */
+	public CodecMap(T nullOperation) {
+		codecs.put(new Identifier("quilt","nothing"), Codec.unit(nullOperation));
```
EnnuiL:
```suggestion
		codecs.put(new Identifier("quilt", "nothing"), Codec.unit(nullOperation));
```

library/core/data_callbacks/build.gradle
```diff
@@ -0,0 +1,19 @@
...
+	moduleName = "data_callbacks"
+	hasMixins = false
+	id = "quilt_data_callbacks"
+	description = "APIs for making events or APIs data-driven"
+	library = "core"
+	moduleDependencies {
+		core {
+			api("qsl_base")
+			testmodOnly("networking")
+			testmodOnly("resource_loader")
```
LambdAurora:
```suggestion
			api("resource_loader")
```

library/worldgen/surface_rule/src/main/java/org/quiltmc/qsl/worldgen/surface_rule/impl/QuiltSurfaceRuleInitializer.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 QuiltMC
+ * Copyright 2022-2023 QuiltMC
```
LambdAurora:
```suggestion
 * Copyright 2022 QuiltMC
```

After some talks, headers should not be updated anymore for existing files.

library/worldgen/surface_rule/src/main/java/org/quiltmc/qsl/worldgen/surface_rule/api/SurfaceRuleEvents.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 QuiltMC
+ * Copyright 2022-2023 QuiltMC
```
LambdAurora:
```suggestion
 * Copyright 2022 QuiltMC
```

library/worldgen/biome/src/main/java/org/quiltmc/qsl/worldgen/biome/api/codec/ModifyEffectsModifier.java
```diff
@@ -0,0 +1,99 @@
...
+import org.quiltmc.qsl.worldgen.biome.api.BiomeModificationContext;
+import org.quiltmc.qsl.worldgen.biome.api.BiomeModifier;
+import org.quiltmc.qsl.worldgen.biome.api.BiomeSelectionContext;
+
+/**
+ * A biome modifier that modifies the effects of a biome.
+ * @see BiomeModificationContext.EffectsContext
+ */
+public record ModifyEffectsModifier(CodecAwarePredicate<BiomeSelectionContext> selector,
+									Optional<Integer> fogColor,
```
LambdAurora: Can't we use OptionalInt for those?... and I swear if that's a DFU restriction.

LambdAurora (changes requested):
I'm wondering if the API shouldn't be called `data_callback` instead of `data_callbacks`.

Also some code style issues:
 - please use the checkstyle Intellij plugin and import in the Code Style stuff the checkstyle.xml present in `codeformat` as the project code style, then reformat the code of this PR.
 - please let empty lines between code and the start/end of a block of code to improve readability.
 - please use `this` when referring to members of the same class
 - please leave an empty line in javadoc between the description and the `@` statements.
 
 also be warned that `{@code}` doesn't prevent HTML tag parsing, so you might have to escape every `<` characters.
 
 Aside from all of this, for my first look at the API I don't see any obvious red flags. Once those changes are made I'll have a second look at the PR and it probably should be good for me.

library/core/data_callbacks/build.gradle
lukebemish: Given that resource_loader is actually entirely optional to use this, I'm not sure why this change needs to be made.

library/worldgen/biome/src/main/java/org/quiltmc/qsl/worldgen/biome/api/codec/ModifyEffectsModifier.java
lukebemish: We probably could with an xmap, though I'm not sure of the advantage, honestly - you'd still be serializing or deserializing it as an `Optional<Integer>` because, well, DFU is all generic stuff; but yeah, that's easy enough to switch around

LambdAurora: Mmh, with DFU forcing an Optional<Integer> anyway there's not a big insensitive to switch then.

library/core/data_callbacks/build.gradle
LambdAurora: Uh. Missed this when reading. Though, to realistically use the API resource loader would be needed right?

lukebemish: Not necessarily. I think I use it for either surface events or biome modifications without touching the resource loader API, using registry loading events instead

LambdAurora: Which would still be needed for the modded resources to even be visible to the registry loading events.

library/worldgen/surface_rule/src/main/java/org/quiltmc/qsl/worldgen/surface_rule/impl/QuiltSurfaceRuleInitializer.java
lukebemish: Does running applyLicenses now correctly apply licenses? If so, I'll just run that and see the results

LambdAurora: I don't think it will do it accurately yet.

lukebemish:
I'm not sure, personally, whether this should be `data_callback` or `data_callbacks` - let me know which and I'll make sure it's that

lukebemish:
Also,
> also be warned that `{@code}` doesn't prevent HTML tag parsing, so you might have to escape every `<` characters.

This isn't correct, `@code` *does* prevent HTML tag parsing and it is safe to use `<` and `>` within. (see https://docs.oracle.com/javase/7/docs/technotes/guides/javadoc/whatsnew-1.5.0.html)

library/worldgen/surface_rule/src/main/java/org/quiltmc/qsl/worldgen/surface_rule/impl/QuiltSurfaceRuleInitializer.java
lukebemish: Should be fixed by latest commit

lukebemish:
>please let empty lines between code and the start/end of a block of code to improve readability.

I'm a bit confused what this would look like. Can you give me an example?

lukebemish:
I went through everything with checkstyle and reformated it - I *believe* I did that right though I always question checkstyle in IntelliJ

LambdAurora:
> Also,
> 
> > also be warned that `{@code}` doesn't prevent HTML tag parsing, so you might have to escape every `<` characters.
> 
> This isn't correct, `@code` _does_ prevent HTML tag parsing and it is safe to use `<` and `>` within. (see https://docs.oracle.com/javase/7/docs/technotes/guides/javadoc/whatsnew-1.5.0.html)

Oops, my bad, I got confused with some of the `<pre><code>` that's around in the code base!

LambdAurora:
> > please let empty lines between code and the start/end of a block of code to improve readability.
> 
> I'm a bit confused what this would look like. Can you give me an example?

For example if you have

```java
this.callCode();
if (i == 2) {
    /* more code */
}
this.callMoreCode();
for (int j = 0; j < 5; j++) {
    if (j % 2 == 0) {
      /* code */
    }
}
if (a == b) {
    /* Even more code */
}
```

It should change to:
```java
this.callCode();

if (i == 2) {
    /* more code */
}

this.callMoreCode();

for (int j = 0; j < 5; j++) {
    if (j % 2 == 0) {
      /* code */
    }
}

if (a == b) {
    /* Even more code */
}
```

lukebemish:
Alright, makes sense! I'll try to find places where I can do that

library/core/data_callbacks/src/main/java/org/quiltmc/qsl/data/callbacks/api/CodecMap.java
```diff
@@ -0,0 +1,133 @@
...
+
+	/**
+	 * Creates a codec map pre-filled with an operation representing a value that has no effect. For instance, this can
+	 * be used to initialize with an event callback that does nothing, or a biome modification that does not modify the
+	 * biome. This provided value is useful as a value that datapacks can use to disable something.
+	 *
+	 * @param nullOperation a value that has no effect
+	 */
+	public CodecMap(T nullOperation) {
+		codecs.put(new Identifier("quilt", "nothing"), Codec.unit(nullOperation));
```
LambdAurora:
```suggestion
		this.codecs.put(new Identifier("quilt", "nothing"), Codec.unit(nullOperation));
```

library/core/data_callbacks/src/main/java/org/quiltmc/qsl/data/callbacks/api/CodecMap.java
```diff
@@ -0,0 +1,133 @@
...
+	 * @param codec the codec to add
+	 * @throws IllegalArgumentException if the identifier is already registered, or if the codec is already registered
+	 */
+	public void register(Identifier id, Codec<? extends T> codec) {
+		if (codecs.containsKey(id)) {
+			throw new IllegalArgumentException("Duplicate codec id: " + id);
+		} else if (codecs.containsValue(codec)) {
+			throw new IllegalArgumentException("Duplicate codec: " + codec);
+		}
+		codecs.put(id, codec);
```
LambdAurora:
```suggestion
		if (this.codecs.containsKey(id)) {
			throw new IllegalArgumentException("Duplicate codec id: " + id);
		} else if (codecs.containsValue(codec)) {
			throw new IllegalArgumentException("Duplicate codec: " + codec);
		}
		this.codecs.put(id, codec);
```

library/core/data_callbacks/src/main/java/org/quiltmc/qsl/data/callbacks/api/CodecMap.java
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	/**
+	 * {@return the codec associated with the provided identifier, or {@code null} if no codec is associated with the identifier}
+	 *
+	 * @param id the identifier to look up
+	 */
+	public Codec<? extends T> lookup(Identifier id) {
+		return codecs.get(id);
```
LambdAurora:
```suggestion
		return this.codecs.get(id);
```

library/core/data_callbacks/src/main/java/org/quiltmc/qsl/data/callbacks/api/CodecMap.java
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	/**
+	 * {@return the identifier associated with the provided codec, or {@code null} if no identifier is associated with the codec}
+	 *
+	 * @param codec the codec to look up
+	 */
+	public Identifier lookup(Codec<? extends T> codec) {
+		return codecs.inverse().get(codec);
```
LambdAurora:
```suggestion
		return this.codecs.inverse().get(codec);
```

library/core/data_callbacks/src/main/java/org/quiltmc/qsl/data/callbacks/api/DynamicEventCallbackSource.java
```diff
@@ -0,0 +1,216 @@
...
+	 *
+	 * @param id       the identifier of the callback, to be used when replacing it in data
+	 * @param listener the callback to listen with
+	 */
+	public void register(Identifier id, T listener) {
+		this.register(id, listener, Event.DEFAULT_PHASE);
+	}
+
+	private void updateListeners(Identifier phase) {
+
```
LambdAurora:
```suggestion
```

library/core/data_callbacks/src/main/java/org/quiltmc/qsl/data/callbacks/api/predicate/NeverPredicate.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * A predicate that is never true.
+ */
+public final class NeverPredicate<T> implements CodecAwarePredicate<T> {
+
+	public static final Identifier ID = new Identifier("quilt", "never");
+	public static final PredicateCodecProvider PROVIDER = NeverPredicate::makeCodec;
+
+	private NeverPredicate() {
+	}
```
LambdAurora:
```suggestion
	private NeverPredicate() {}
```

library/core/data_callbacks/src/main/java/org/quiltmc/qsl/data/callbacks/api/DynamicEventCallbackSource.java
```diff
@@ -0,0 +1,216 @@
...
+		if (!this.registeredEvents.contains(phase)) {
+			this.registeredEvents.add(phase);
+			this.event.register(phase, this.combiner.apply(() -> this.getListeners(phase)));
+		}
+	}
+
+	private void updateDynamicListeners(Map<Identifier, Pair<Identifier, T>> dynamicListeners) {
+		this.dynamicListeners.clear();
+		this.dynamicListeners.putAll(dynamicListeners);
+		dynamicListeners.values().stream().map(Pair::getFirst).distinct().forEach(this::updateListeners);
```
LambdAurora:
```suggestion
		dynamicListeners.values().stream()
				.map(Pair::getFirst)
				.distinct()
				.forEach(this::updateListeners);
```

library/core/data_callbacks/src/main/java/org/quiltmc/qsl/data/callbacks/api/DynamicEventCallbackSource.java
```diff
@@ -0,0 +1,216 @@
...
+
+	/**
+	 * Updates the listeners with callbacks loaded from data, using custom {@link DynamicOps}. This is useful when
+	 * loading using codecs that want a {@link net.minecraft.registry.RegistryOps} or similar.
+	 *
+	 * @param resourceManager the resource manager to load data from
+	 * @param ops             the dynamic ops to use to decode data
+	 */
+	public void update(ResourceManager resourceManager, DynamicOps<JsonElement> ops) {
+		Map<Identifier, Pair<Identifier, T>> dynamicListeners = new LinkedHashMap<>();
```
LambdAurora:
```suggestion
		var dynamicListeners = new LinkedHashMap<Identifier, Pair<Identifier, T>>();
```

library/core/data_callbacks/src/main/java/org/quiltmc/qsl/data/callbacks/api/DynamicEventCallbackSource.java
```diff
@@ -0,0 +1,216 @@
...
+					dynamicListeners.put(unwrappedIdentifier, pair);
+				} else {
+					LOGGER.error("Couldn't parse data file {} from {}: {}", unwrappedIdentifier, id, result.error().get().message());
+				}
+			} catch (IOException e) {
+				LOGGER.error("Couldn't parse data file {} from {}", unwrappedIdentifier, id, e);
+			}
+		}
+
+		updateDynamicListeners(dynamicListeners);
```
LambdAurora:
```suggestion
		this.updateDynamicListeners(dynamicListeners);
```

library/core/data_callbacks/src/main/java/org/quiltmc/qsl/data/callbacks/api/predicate/AlwaysPredicate.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * A predicate that is always true.
+ */
+public final class AlwaysPredicate<T> implements CodecAwarePredicate<T> {
+
+	public static final Identifier ID = new Identifier("quilt", "always");
+	public static final PredicateCodecProvider PROVIDER = AlwaysPredicate::makeCodec;
+
+	private AlwaysPredicate() {
+	}
```
LambdAurora:
```suggestion
	private AlwaysPredicate() {}
```

library/core/data_callbacks/src/main/java/org/quiltmc/qsl/data/callbacks/api/predicate/PredicateCodecMap.java
```diff
@@ -0,0 +1,118 @@
...
+import org.quiltmc.qsl.data.callbacks.api.CodecMap;
+
+/**
+ * A CodecMap for {@link CodecAwarePredicate}s that bundles a set of codecs created by shared providers. This class
+ * also handles registering shared providers for predicate codecs.
+ *
+ * @param <T> the type of the input tested by the predicates this map can help encode
+ */
+public class PredicateCodecMap<T> extends CodecMap<CodecAwarePredicate<T>> {
+	private static final Map<Identifier, PredicateCodecProvider> providers = HashBiMap.create();
```
LambdAurora: As it's a static final field it should be named `PROVIDERS`.

library/core/data_callbacks/src/main/java/org/quiltmc/qsl/data/callbacks/api/predicate/PredicateCodecMap.java
```diff
@@ -0,0 +1,118 @@
...
+	public Identifier lookup(Codec<? extends CodecAwarePredicate<T>> codec) {
+		checkCache();
+		return super.lookup(codec);
+	}
+
+	private void checkCache() {
+		if (!this.cached) {
+			synchronized (this) {
+				if (!this.cached) {
+					cacheProviders();
```
LambdAurora:
```suggestion
					this.cacheProviders();
```

library/core/data_callbacks/src/main/java/org/quiltmc/qsl/data/callbacks/api/predicate/PredicateCodecMap.java
```diff
@@ -0,0 +1,118 @@
...
+
+	@Override
+	public Codec<? extends CodecAwarePredicate<T>> lookup(Identifier id) {
+		checkCache();
+		return super.lookup(id);
+	}
+
+	@Override
+	public Identifier lookup(Codec<? extends CodecAwarePredicate<T>> codec) {
+		checkCache();
```
LambdAurora:
```suggestion
		this.checkCache();
```

library/core/data_callbacks/src/main/java/org/quiltmc/qsl/data/callbacks/api/predicate/PredicateCodecMap.java
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	@Override
+	public void register(Identifier id, Codec<? extends CodecAwarePredicate<T>> codec) {
+		super.register(id, codec);
+	}
+
+	@Override
+	public Codec<? extends CodecAwarePredicate<T>> lookup(Identifier id) {
+		checkCache();
```
LambdAurora:
```suggestion
		this.checkCache();
```

library/worldgen/biome/build.gradle
```diff
@@ -13,13 +13,17 @@ qslModule {
...
+			api("data_callbacks")
 			testmodOnly("resource_loader")
 		}
 	}
 	entrypoints {
 		events {
 			values = ["org.quiltmc.qsl.worldgen.biome.impl.TheEndBiomeData"]
 		}
+		init {
+			values =["org.quiltmc.qsl.worldgen.biome.impl.modification.BiomeModifierCodecs"]
```
LambdAurora:
```suggestion
			values = ["org.quiltmc.qsl.worldgen.biome.impl.modification.BiomeModifierCodecs"]
```

library/worldgen/biome/src/main/java/org/quiltmc/qsl/worldgen/biome/api/BiomeModifier.java
```diff
@@ -0,0 +1,58 @@
...
+
+	/**
+	 * Applies this biome modifier to the biome.
+	 */
+	void modify(BiomeSelectionContext selectionContext, BiomeModificationContext modificationContext);
+
+	/**
+	 * Stores biome selector codecs. Custom biome selector codecs should be added here during mod initialization.
+	 */
+	PredicateCodecMap<BiomeSelectionContext> BIOME_SELECTOR_CODECS = new PredicateCodecMap<>(BIOME_SELECTOR_CODEC);
```
LambdAurora: Might be better to move this at the top of the interface with the BIOME_MODIFIER_CODECS.

library/worldgen/biome/src/main/java/org/quiltmc/qsl/worldgen/biome/impl/modification/BiomeModificationReloader.java
```diff
@@ -0,0 +1,110 @@
...
+					// We have to catch the 'java.lang.IllegalStateException: Missing tag TagKey[minecraft:worldgen/biome / minecraft:increased_fire_burnout]'
+					// that can be thrown by the biome holder list codec...
+					LOGGER.error("Couldn't parse data file {} from {}", unwrappedIdentifier, id, e);
+				}
+			} catch (IOException e) {
+				LOGGER.error("Couldn't parse data file {} from {}", unwrappedIdentifier, id, e);
+			}
+		}
+
+		updateDynamicListeners(dynamicListeners);
```
LambdAurora:
```suggestion
		this.updateDynamicListeners(dynamicListeners);
```

library/worldgen/biome/src/main/java/org/quiltmc/qsl/worldgen/biome/impl/modification/BiomeModificationReloader.java
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	private void updateDynamicListeners(Map<Identifier, Pair<ModificationPhase, BiomeModifier>> dynamicListeners) {
+		combinedListeners.clear();
+		combinedListeners.putAll(listeners);
+		combinedListeners.putAll(dynamicListeners);
```
LambdAurora:
```suggestion
		this.combinedListeners.clear();
		this.combinedListeners.putAll(this.listeners);
		this.combinedListeners.putAll(dynamicListeners);
```

library/worldgen/biome/src/main/java/org/quiltmc/qsl/worldgen/biome/impl/modification/BiomeModificationReloader.java
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	public Map<Identifier, BiomeModifier> getCombinedMap(ModificationPhase phase) {
+		return combinedListeners.entrySet().stream().filter(entry -> entry.getValue().getFirst() == phase).collect(Collectors.toMap(Map.Entry::getKey, entry -> entry.getValue().getSecond()));
+	}
+
+	public void addModifier(ModificationPhase phase, Identifier identifier, BiomeModifier modifier) {
+		listeners.put(identifier, new Pair<>(phase, modifier));
```
LambdAurora:
```suggestion
		this.listeners.put(identifier, new Pair<>(phase, modifier));
```

library/worldgen/biome/src/main/java/org/quiltmc/qsl/worldgen/biome/impl/modification/BiomeModificationReloader.java
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	private void updateDynamicListeners(Map<Identifier, Pair<ModificationPhase, BiomeModifier>> dynamicListeners) {
+		combinedListeners.clear();
+		combinedListeners.putAll(listeners);
+		combinedListeners.putAll(dynamicListeners);
+	}
+
+	public Map<Identifier, BiomeModifier> getCombinedMap(ModificationPhase phase) {
+		return combinedListeners.entrySet().stream().filter(entry -> entry.getValue().getFirst() == phase).collect(Collectors.toMap(Map.Entry::getKey, entry -> entry.getValue().getSecond()));
```
LambdAurora:
```suggestion
		return this.combinedListeners.entrySet().stream()
				.filter(entry -> entry.getValue().getFirst() == phase)
				.collect(Collectors.toMap(Map.Entry::getKey, entry -> entry.getValue().getSecond()));
```

library/worldgen/biome/src/main/java/org/quiltmc/qsl/worldgen/biome/mixin/modification/MinecraftServerMixin.java
```diff
@@ -40,12 +41,15 @@ public abstract class MinecraftServerMixin {
...
+	public abstract ResourceManager getResourceManager();
+
 	@Inject(method = "<init>", at = @At(value = "RETURN"))
 	private void finalizeWorldGen(CallbackInfo ci) {
 		if (!(this.saveProperties instanceof WorldSaveProperties levelProperties)) {
 			throw new RuntimeException("Incompatible SaveProperties passed to MinecraftServer: " + saveProperties);
 		}
 
-		BiomeModificationImpl.INSTANCE.finalizeWorldGen(getRegistryManager(), levelProperties);
+		BiomeModificationImpl.INSTANCE.finalizeWorldGen(getRegistryManager(), levelProperties, getResourceManager());
```
LambdAurora:
```suggestion
		BiomeModificationImpl.INSTANCE.finalizeWorldGen(this.getRegistryManager(), levelProperties, this.getResourceManager());
```

LambdAurora (commented):
I really think the name should be singular as we already say "the Event API" and not "the Events API".

library/worldgen/biome/src/main/java/org/quiltmc/qsl/worldgen/biome/api/BiomeModifier.java
lukebemish: It has to be after `BIOME_SELECTOR_CODEC`, but I can move both up

LambdAurora: For sure, I just wanted the fields to be grouped at the beginning of the class.

lukebemish:
Assuming I didn't break anything else it should now be renamed to `data_callback`

library/worldgen/biome/src/main/java/org/quiltmc/qsl/worldgen/biome/api/BiomeModifier.java
lukebemish: I think it originally got ordered this way when I ran the checkstyle; not sure why

library/core/data_callback/src/main/java/org/quiltmc/qsl/data/callbacks/api/CodecHelpers.java
```diff
@@ -0,0 +1,39 @@
...
+
+import com.mojang.datafixers.util.Either;
+import com.mojang.serialization.Codec;
+
+/**
+ * Helper methods for common codec operations.
+ */
+public final class CodecHelpers {
+	private CodecHelpers() {
+	}
```
EnnuiL:
```suggestion
	private CodecHelpers() {}
```

library/core/data_callback/src/main/java/org/quiltmc/qsl/data/callbacks/api/CodecMap.java
```diff
@@ -0,0 +1,133 @@
...
+ * parameterized by the type of this map can be encoded and decoded using the map's codec, and those that are encodable
+ * should provide codec identifiers for a corresponding codec in this map.
+ *
+ * @param <T> the codec-aware type that this map dispatches to
+ */
+public class CodecMap<T extends CodecAware> {
+	private final BiMap<Identifier, Codec<? extends T>> codecs = HashBiMap.create();
+
+	public CodecMap() {
+	}
```
EnnuiL:
```suggestion
	public CodecMap() {}
```

library/core/data_callback/src/main/java/org/quiltmc/qsl/data/callbacks/api/predicate/AlwaysPredicate.java
```diff
@@ -0,0 +1,46 @@
...
+import com.mojang.serialization.Codec;
+
+import net.minecraft.util.Identifier;
+
+/**
+ * A predicate that is always true.
+ */
+public final class AlwaysPredicate<T> implements CodecAwarePredicate<T> {
+
+	public static final Identifier ID = new Identifier("quilt", "always");
```
EnnuiL: The last nitpick that I have about these var names is that I think `CODEC_ID` would be better than `ID` here; Could you change those?

library/core/data_callback/src/testmod/java/org/quiltmc/qsl/lifecycle/test/event/DataCallbackTests.java
```diff
@@ -0,0 +1,110 @@
...
+	// would be added to the base event itself instead; however, this is a test mod, and we would have to make a whole
+	// new useful server-side event otherwise.
+	public static final Event<ServerJoin> SERVER_JOIN = Event.createWithPhases(ServerJoin.class, callbacks -> (handler, sender, server) -> {
+		for (ServerJoin callback : callbacks) {
+			callback.onPlayReady(handler, sender, server);
+		}
+	}, BEFORE_PHASE, Event.DEFAULT_PHASE, AFTER_PHASE);
+
+	public static final CodecMap<ServerJoin> JOIN_SERVER_CODECS = new CodecMap<>((handler, sender, server) -> {
+	});
```
EnnuiL:
```suggestion
	public static final CodecMap<ServerJoin> JOIN_SERVER_CODECS = new CodecMap<>((handler, sender, server) -> {});
```

library/worldgen/biome/src/main/java/org/quiltmc/qsl/worldgen/biome/api/codec/AddCarversModifier.java
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * A biome modifier that adds carvers to a biome.
+ *
+ * @param step    the carver generation step to add the carvers to
+ * @param carvers registry keys for the carvers to add
+ */
+public record AddCarversModifier(CodecAwarePredicate<BiomeSelectionContext> selector,
+								 List<RegistryKey<ConfiguredCarver<?>>> carvers,
+								 GenerationStep.Carver step) implements BiomeModifier {
+	public static final Identifier ID = new Identifier("quilt", "add_carvers");
```
EnnuiL: Same thing here!

library/worldgen/biome/src/main/java/org/quiltmc/qsl/worldgen/biome/impl/modification/BiomeModificationImpl.java
```diff
@@ -56,6 +60,8 @@ public class BiomeModificationImpl {
 
 	private boolean modifiersUnsorted = true;
 
+	final BiomeModificationReloader reloader = new BiomeModificationReloader();
+
 	private BiomeModificationImpl() {
 	}
```
EnnuiL: I know it isn't your code, but:
```suggestion
	private BiomeModificationImpl() {}
```
why not :p

library/worldgen/biome/src/main/java/org/quiltmc/qsl/worldgen/biome/api/codec/NothingModifier.java
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * A biome modifier that does nothing; useful to override existing biome modifiers with a datapack.
+ */
+public final class NothingModifier implements BiomeModifier {
+	public static final Identifier ID = new Identifier("quilt", "nothing");
+	public static final NothingModifier INSTANCE = new NothingModifier();
+	public static final Codec<NothingModifier> CODEC = Codec.unit(INSTANCE);
+
+	private NothingModifier() {
+	}
```
EnnuiL:
```suggestion
	private NothingModifier() {}
```

library/worldgen/biome/src/main/java/org/quiltmc/qsl/worldgen/biome/api/codec/NothingModifier.java
```diff
@@ -0,0 +1,52 @@
...
+
+	@Override
+	public boolean shouldModify(BiomeSelectionContext context) {
+		return false;
+	}
+
+	@Override
+	public void modify(BiomeSelectionContext selectionContext, BiomeModificationContext modificationContext) {
+
+	}
```
EnnuiL:
```suggestion
	public void modify(BiomeSelectionContext selectionContext, BiomeModificationContext modificationContext) {}
```

EnnuiL (changes requested):
Congrats! You made me run out of nitpicks! This is a good thing!

library/core/data_callback/src/main/java/org/quiltmc/qsl/data/callbacks/api/predicate/AlwaysPredicate.java
lukebemish: These should all be fixed

EnnuiL (approved):
I'm satisfied; Have the Enny Stamp of Approval

LambdAurora (approved):

LambdAurora:
I declare this pull request now in Final Comment Period for 5 days!

Leo40Git (approved):
LGTM!

PR #185, opened 2022-08-28
Quilt Registry Builder

Leo40Git:
Add registry builders!
- The base `QuiltRegistryBuilder` class
- Built-in (code-driven) registries: `QuiltBuiltinRegistryBuilder`
- Dynamic (data-driven) registries: `QuiltDynamicRegistryBuilder`
  - Registers into the `DynamicRegistryManager`!

### TODO
- [ ] Document dynamic registries more!!!
- [ ] Add dynamic registry testmod

PR #271, opened 2023-02-25
Add a dynamic registry registration API

Pyrofab:
Dynamic registries are a way for mojang to load data automatically from datapacks and optionally synchronize it with clients - a super useful feature that is unfortunately super locked down due to the use of constant unmodifiable collections in lieu of proper registry-registries (I call those meta-registries for short).

A first attempt to resolve this issue was made in #185, but 1.19.3 changed this system significantly - to the point I believe a new proposal is necessary.

Since builtin registries are no longer a thing, I believe a builder-style API is not as relevant. This PR thus adds a single API class with 2 methods for registering new dynamic registries, either serverside-only or synced with clients.
The test mod showcases the creation of a modded dynamic registry, with entry registration done through both data files and registry event.


OroArmor (approved):
Looks good from a first glance!

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/dynamic/DynamicMetaregistry.java
```diff
@@ -0,0 +1,74 @@
...
+	 *
+	 * <p>Synced dynamic registries do not have to be explicitly registered clientside to be usable.
+	 *
+	 * @param <E>        the type of elements in the dynamic registry
+	 * @param key        a {@link RegistryKey#ofRegistry(Identifier) key for the new dynamic registry}
+	 * @param entryCodec the codec used to deserialize entries from datapacks
+	 * @param syncCodec  the codec used to (de)serialize entries to and from packets - may be the same as {@code entryCodec}
+	 * @throws IllegalStateException if this registry of registries already got frozen
+	 */
+	public static <E> void registerSynced(RegistryKey<? extends Registry<E>> key, Codec<E> entryCodec, Codec<E> syncCodec) {
```
Leo40Git: Maybe offer a `RegistryKey, Codec` overload that specifies the codec as both the data and sync codec?

Leo40Git (approved):
LGTM!

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/dynamic/DynamicMetaregistry.java
Pyrofab: Done 👍

library/core/registry/src/main/java/org/quiltmc/qsl/registry/mixin/RegistryLoaderMixin.java
```diff
@@ -16,29 +16,37 @@
...
-import net.minecraft.registry.RegistryKey;
-import net.minecraft.resource.ResourceManager;
-import net.minecraft.registry.DynamicRegistryManager;
-import net.minecraft.registry.RegistryLoader;
-
-import org.quiltmc.qsl.registry.api.event.RegistryEvents;
-import org.quiltmc.qsl.registry.impl.DynamicRegistryManagerSetupContextImpl;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
```
EnnuiL: Hm, I recommend doing a Checkstyle pass on imports

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/dynamic/DynamicMetaregistry.java
```diff
@@ -0,0 +1,92 @@
...
+	 * @see #registerSynced(RegistryKey, Codec, Codec)
+	 */
+	public static <E> void registerSynced(RegistryKey<? extends Registry<E>> key, Codec<E> entryCodec) {
+		DynamicMetaregistryImpl.registerSynced(key, entryCodec, entryCodec);
+	}
+
+	/**
+	 * Registers a dynamic registry which contents get synced between the server and connected clients.
+	 *
+	 * <p>Entries will be loaded from {@code "data/<namespace>/<registry_path>"} for every datapack {@code namespace},
```
EnnuiL:
```suggestion
	 * <p>
	 * Entries will be loaded from {@code "data/<namespace>/<registry_path>"} for every datapack {@code namespace},
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/dynamic/DynamicMetaregistry.java
```diff
@@ -0,0 +1,92 @@
...
+	 * @throws IllegalStateException if this registry of registries already got frozen
+	 */
+	public static <E> void register(RegistryKey<? extends Registry<E>> key, Codec<E> entryCodec) {
+		DynamicMetaregistryImpl.register(key, entryCodec);
+	}
+
+	/**
+	 * Registers a dynamic registry which contents get synced between the server and connected clients.
+	 *
+	 * <p>Entries will be loaded from {@code "data/<namespace>/<registry_path>"} for every datapack {@code namespace},
```
EnnuiL:
```suggestion
	 * <p>
	 * Entries will be loaded from {@code "data/<namespace>/<registry_path>"} for every datapack {@code namespace},
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/dynamic/DynamicMetaregistry.java
```diff
@@ -0,0 +1,92 @@
...
+	 * where {@code registry_path}'s value is {@code key.getLocation().getPath()}.
+	 * The registry's own namespace is essentially ignored when loading values,
+	 * meaning <strong>the registry {@code key}'s path must be unique by itself</strong>
```
EnnuiL: oh no, I think that's one of the reasons why the original PR stalled; we actually wanted a mechanism where if the namespace is omitted, it's assumed to be `minecraft`

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/dynamic/DynamicMetaregistry.java
```diff
@@ -0,0 +1,92 @@
...
+ * <li>through the {@link org.quiltmc.qsl.registry.api.event.RegistryEvents#DYNAMIC_REGISTRY_SETUP dynamic registry setup event}
+ * </ul>
+ *
+ * <p>This registry will be frozen at the same time as static registries.
+ */
+public final class DynamicMetaregistry {
+	/**
+	 * Registers a serverside dynamic registry.
+	 *
+	 * <p>Entries will be loaded from {@code "data/<namespace>/<registry_path>"} for every datapack {@code namespace},
```
EnnuiL:
```suggestion
	 * <p>
	 * Entries will be loaded from {@code "data/<namespace>/<registry_path>"} for every datapack {@code namespace},
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/dynamic/DynamicMetaregistry.java
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Allows registration of dynamic registries for use through {@link net.minecraft.registry.DynamicRegistryManager}.
+ *
+ * <p>Dynamic registries are only available in a world context. Entries must be added either:<ul>
+ * <li>as datapack files
+ * <li>through the {@link org.quiltmc.qsl.registry.api.event.RegistryEvents#DYNAMIC_REGISTRY_SETUP dynamic registry setup event}
+ * </ul>
+ *
+ * <p>This registry will be frozen at the same time as static registries.
```
EnnuiL:
```suggestion
 * <p>
 * Dynamic registries are only available in a world context. Entries must be added either:
 * <ul>
 * <li>as datapack files
 * <li>through the {@link org.quiltmc.qsl.registry.api.event.RegistryEvents#DYNAMIC_REGISTRY_SETUP dynamic registry setup event}
 * </ul>
 * <p>
 * This registry will be frozen at the same time as static registries.
```

EnnuiL (changes requested):
hm, I might not be the registry gal or anything, but oh dear, other than nitpicks, i do worry about one point:

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/dynamic/DynamicMetaregistry.java
Pyrofab: Maybe the javadoc isn't clear, or maybe I didn't understand you - what do you want to default to `minecraft` exactly ? The issue I'm trying to allude to is that `mod1:dynamic_registry` and `mod2:dynamic_registry` are loaded with the path `data/<every_namespace_that_exists>/dynamic_registry`

EnnuiL: ah, what I mean is that for modded dynamic registries, you should specify a namespace, or else `data/<namespace>/dynamic_registry` is assumed to refer to `minecraft:dynamic_registry` and therefore it would be equivalent to `data/<namespace>/minecraft/dynamic_registry`;
There has been convos before about having this implemented in general (mainly due to how REA has explicit namespaces but it doesn't allow to omit `minecraft` for vanilla registries), and yeah, this was a conclusion that we had, and yeah, I believe that if we end avoiding problems where registries must be like `mod_id:mod_id_gases`? it would be great

Pyrofab: So if I understand correctly, the solution you suggest would be to systematically prepend the registry path with `<registry_namespace>/`, and to patch registry loading such that vanilla registries can be loaded from `minecraft/<registry_path>` ?

lukebemish: This seems like a good system; it's also very similar to the system that my PR (#264) uses for non-registry data, so consistency is always nice.

EnnuiL: > So if I understand correctly, the solution you suggest would be to systematically prepend the registry path with `<registry_namespace>/`, and to patch registry loading such that vanilla registries can be loaded from `minecraft/<registry_path>` ?

I needed a discussion on Discord in order to clear my mind more; Conclusion? Yep, this would bring the functionality that I've been thinking of

Pyrofab: Looking again at the code, the path from which resources are loaded is tightly coupled to the key. I therefore see 3 possibilities for implementing this:
1. Patch the registry loading code such that resources for the key `<namespace>:<path>` are always loaded from `<namespace>/<path>`, with a hardcoded exception for the Minecraft namespace. Pro: mods can just use `RegistryKey.ofRegistry(new Identifier("<namespace>", "<path>"))`. Con: immediately incompatible with every mod that already registers dynamic registries (e.g. [Blabber](https://github.com/Ladysnake/Blabber)).
2. Force the registry keys used to register new dynamic registries to have a `<namespace>:<namespace>/<path>` structure. Pros: least effort, transparent for mods. Con: not statically verified.
3. Make all registration methods take an `Identifier` and return a `RegistryKey` built as `<namespace>:<namespace>/<path>`. Pro: statically ensures the path is correct. Con: those keys are usually stored as constants, meaning the registration would have to be done during static initialization and would become part of a mod's public API.

By myself I would lean towards option 2 as it seems the least intrusive. What do you think ?

TheGlitch76:
This PR needs to be updated to 1.19.4

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/dynamic/DynamicMetaregistry.java
EnnuiL: Hm, I'd be heavily inclined to Option 1 due to it looking like the best solution, with the mod breaking not being a concern as long as the ecosystem could move to it; Problem: unfortunately, we share the ecosystem with this hecker: *looks at fabric*;

Unfortunately, we'll have to compromise then with a solution that is backwards-compatible, which yeah, is complicated; I don't like the idea of having RegistryKeys not be the straightforward `<namespace>:<path>`, so I'm discarding options 2 and 3;

I do wonder if we could adopt option 1 and like, have some sort of solution on QFAPI's side that attempts to add exceptions for Fabric mods, similar to how QSL has content registries based on REA but QFAPI has deferring queues in order to adjust messy attempts to add to them by Fabric mods; I am having a hard time to think about how this exception system could be done though, since it could end effectively destroying the benefits of this system if done wrong; if done right, however, this option 1.5 would be the best option here, I believe

Pyrofab: Fair enough, I went ahead and implemented what I think is a decent spin on option 1 - making it so only dynamic registries registered through our API get affected by the path changes.

williambl (approved):

EnnuiL (approved):
We've determined that for now, Pyro's take on Option 1 will be the best solution for now, and it might remain like that in the future;
While I've thought backwards-support of new paths for other modded dynamic registries (this excludes minecraft's) would be nice and helpful, the hackery needed for that might make that not worth it; If it is to be gone through? It is best to be done as a separate PR

EnnuiL:
Maven Central might have burnt down, the Discord servers might be under a dictatorship, I don't care;
A Final Commend Period of 5 days for this PR has began and it *will* be over after 5 days
It doesn't matter if you own the Discords, this is our project and the show *must* go on

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/dynamic/DynamicMetaregistry.java
```diff
@@ -0,0 +1,89 @@
...
+import net.minecraft.util.Identifier;
+
+import org.quiltmc.qsl.registry.impl.dynamic.DynamicMetaregistryImpl;
+
+/**
+ * Allows registration of dynamic registries for use through {@link net.minecraft.registry.DynamicRegistryManager}.
+ * <p>
+ * Dynamic registries are only available in a world context. Entries must be added either:
+ * <ul>
+ * <li>as datapack files
```
LambdAurora:
```suggestion
 * <li>as data-pack files
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/dynamic/DynamicMetaregistry.java
```diff
@@ -0,0 +1,89 @@
...
+ * <li>through the {@link org.quiltmc.qsl.registry.api.event.RegistryEvents#DYNAMIC_REGISTRY_SETUP dynamic registry setup event}
+ * </ul>
+ * <p>
+ * This registry will be frozen at the same time as static registries.
+ */
+public final class DynamicMetaregistry {
+	/**
+	 * Registers a serverside dynamic registry.
```
LambdAurora:
```suggestion
	 * Registers a server-side dynamic registry.
```

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/dynamic/DynamicMetaregistryImpl.java
```diff
@@ -0,0 +1,56 @@
...
+import com.mojang.serialization.Codec;
+
+import net.minecraft.registry.Registry;
+import net.minecraft.registry.RegistryKey;
+import net.minecraft.registry.RegistryLoader;
+import net.minecraft.util.Identifier;
+
+import org.quiltmc.qsl.registry.mixin.DynamicRegistrySyncAccessor;
+
+public class DynamicMetaregistryImpl {
```
LambdAurora:
```suggestion
@ApiStatus.Internal
public class DynamicMetaRegistryImpl {
```

Will also need the import in case you try to add my suggestions directly in the GitHub UI.

LambdAurora (changes requested):

library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/dynamic/DynamicMetaregistryImpl.java
Leo40Git: You forgot to rename the API class too :P

LambdAurora: I missed that oops, anyway this needs IDE intervention anyways.

LambdAurora (approved):

PR #274, opened 2023-03-01
Initial feature flags API

Platymemo:
An API to make use of the new `FeatureFlags` system.

Originally limited to 64 flags, this API extends the size of feature flags using a `BitSet`, which grows as needed.

Additionally, the `GatedFeature#FILTERED_REGISTRIES` set is made mutable to allow for new registries to be added.


library/data/feature_flag/src/main/java/org/quiltmc/qsl/feature/flag/mixin/FeatureFlagBitSetMixin.java
```diff
@@ -0,0 +1,104 @@
...
+import net.minecraft.feature_flags.FeatureFlagGroup;
+
+import org.quiltmc.qsl.feature.flag.impl.QuiltFeatureFlagExtensions;
+
+@Mixin(FeatureFlagBitSet.class)
+public class FeatureFlagBitSetMixin {
+	@Unique
+	private final BitSet quilt$additionalFlags = new BitSet();
+
+	@Inject(method = "isIn", at = @At(value = "RETURN", ordinal = 2), cancellable = true)
```
TheGlitch76, line 39: I'd rather see use of `@Slice` over ordinals. Ordinals are super fragile (what if the method adds an early return at the head?) and if this code ever breaks, it's not going to be clear to me (or whoever else takes up the responsibility of porting) what you were targeting. Feel free to ping me if you need help.

Platymemo: In this case, ordinal is fairly safe because the function is 6 lines, an if-else chain with two early returns, then a simple bit mask check. If a slice is still recommended I can put one in, but this is one of the few cases where I think including a slice is unnecessary or possibly even more confusing.

library/data/feature_flag/src/main/java/org/quiltmc/qsl/feature/flag/api/QuiltFeatureFlags.java
```diff
@@ -0,0 +1,40 @@
...
+import net.minecraft.feature_flags.FeatureFlags;
+import net.minecraft.util.Identifier;
+
+import org.quiltmc.qsl.feature.flag.impl.QuiltFeatureFlagRegistryExtensions;
+
+/**
+ * Provides helper methods for working with {@link FeatureFlag}s.
+ */
+@ApiStatus.NonExtendable
+public interface QuiltFeatureFlags {
```
Leo40Git, line 31: This should be a (final) class for consistency.

Leo40Git (commented):
Can't do a proper review just yet (phone review time) so here's a nitpick

library/data/feature_flag/src/main/java/org/quiltmc/qsl/feature/flag/mixin/FeatureFlagBitSetMixin.java
TheGlitch76: If possible, I'd like enough description of some kind (either a few words in a comment or a slice) that gives me enough of an idea that I could fix the ordinals on this mixin in an update without needing to open up an old QSL workspace and reverse-engineer the injection points myself.

PR #278, opened 2023-03-05
Add dynamic registry manager references to Recipe API.

### LINKED ISSUE #265: Quilt Recipe API needs to be adapted for QSL 1.19.4
Basically, a major change that happened on the recipe system is that methods like `getOutput` now accept a DynamicRegistryManager, which means that in order to adapt for that, it might be time for the callbacks to provide one as well; While the actual usage of DRMs by recipes is limited to smithing ones, I'm worried about the implications for the future, and therefore, we need to adapt somehow

LambdAurora:


EnnuiL (approved):
wait oh my god i love synergy between APIs; i completely forgot about that possibility

PR #286, opened 2023-03-11
Refactor Quilt Screen API in order to fit 1.19.4

### LINKED ISSUE #282: Adapt Quilt Screen API to 1.19.4
There is currently a bug where whenever changes you do to the button list ends being lost upon the reinitialization done after resizing; Fabric fixes it by making the `resize` method invoke the init option, but something tells me that's either the wrong approach or a partial one

EnnuiL:
This moves the injection to the `init()V` calls inside the Screen class since that's what makes most sense within Vanilla's system; The parameter list has been changed as well, and some of the legacy remains have been cleared out;
With the new layout system, however, it's clear that further adaptating will be required in the future, since the ButtonList system does not seamlessly fit into some of the new screens, especially those who override `repositionElements`

oh dear, 1.19.4 is coming and it's coming fast and yet this feels like one of those PRs that need a 7 day FCP; i don't wanna have the old stuff remain ;-;

### DIFF
--- a/library/gui/screen/src/main/java/org/quiltmc/qsl/screen/api/client/ScreenEvents.java
+++ b/library/gui/screen/src/main/java/org/quiltmc/qsl/screen/api/client/ScreenEvents.java
@@ -137,12 +137,26 @@ public final class ScreenEvents {
 	@ClientOnly
 	@FunctionalInterface
 	public interface BeforeInit extends ClientEventAwareListener {
+		/**
+		 * An event that is called before a screen is initialized to its default state.
+		 *
+		 * @param screen the screen
+		 * @param client the screen's {@link MinecraftClient client} instance
+		 * @param firstInit {@code true} if the screen has been initialized for the first time, or {@code false} otherwise
+		 */
 		void beforeInit(Screen screen, MinecraftClient client, boolean firstInit);
 	}
 
 	@ClientOnly
 	@FunctionalInterface
 	public interface AfterInit extends ClientEventAwareListener {
+		/**
+		 * An event that is called after a screen is initialized to its default state.
+		 *
+		 * @param screen the screen
+		 * @param client the screen's {@link MinecraftClient client} instance
+		 * @param firstInit {@code true} if the screen has been initialized for the first time, or {@code false} otherwise
+		 */
 		void afterInit(Screen screen, MinecraftClient client, boolean firstInit);
 	}
 
--- a/library/gui/screen/src/main/java/org/quiltmc/qsl/screen/api/client/ScreenEvents.java
+++ b/library/gui/screen/src/main/java/org/quiltmc/qsl/screen/api/client/ScreenEvents.java
@@ -49,9 +49,9 @@ public final class ScreenEvents {
 	 *
 	 * @see ScreenEvents#AFTER_INIT
 	 */
-	public static final Event<BeforeInit> BEFORE_INIT = Event.create(BeforeInit.class, callbacks -> (screen, firstInit) -> {
+	public static final Event<BeforeInit> BEFORE_INIT = Event.create(BeforeInit.class, callbacks -> (screen, client, firstInit) -> {
 		for (var callback : callbacks) {
-			callback.beforeInit(screen, firstInit);
+			callback.beforeInit(screen, client, firstInit);
 		}
 	});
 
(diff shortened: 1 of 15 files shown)


maximumpower55 (approved):
LGTM

library/gui/screen/src/testmod/java/org/quiltmc/qsl/screen/test/client/ScreenTests.java
```diff
@@ -43,13 +42,17 @@ public class ScreenTests implements ScreenEvents.AfterInit, ScreenEvents.AfterRe
...
+			LOGGER.info(screen.getClass() + " has been initialized for the first time");
+		}
+
 		if (screen instanceof TitleScreen) {
 			final List<ClickableWidget> buttons = screen.getButtons();
 
 			buttons.add(
-					ButtonWidget.builder(Text.of("Hello world!!"), button -> LOGGER.info("Hello world!!"))
-							.positionAndSize((screen.width / 2) + 120, ((screen.height / 4) + 95), 70, 20)
+					ButtonWidget.builder(Text.of("Hello world!"), button -> LOGGER.info("Hello world!!"))
```
EnnuiL: whoops
```suggestion
					ButtonWidget.builder(Text.of("Hello world!!"), button -> LOGGER.info("Hello world!!"))
```

Leo40Git (approved):

LambdAurora (commented):
Why is MinecraftClient suddenly nullable?
That's quite bad imo

EnnuiL:
> Why is MinecraftClient suddenly nullable?
> That's quite bad imo

Hello! Stuck-at-phone Ennui here! It's pretty much because really, the original shadowed field is nullable, therefore, why not have the method that gets it nullable too?
I will double-check though once I'm home

LambdAurora:
> Hello! Stuck-at-phone Ennui here! It's pretty much because really, the original shadowed field is nullable, therefore, why not have the method that gets it nullable too? I will double-check though once I'm home

It's one of those cases where you should re-read `@Nullable` documentation. Where those events are called the client is never `null`.

EnnuiL:
> > Hello! Stuck-at-phone Ennui here! It's pretty much because really, the original shadowed field is nullable, therefore, why not have the method that gets it nullable too? I will double-check though once I'm home
> 
> It's one of those cases where you should re-read `@Nullable` documentation. Where those events are called the client is never `null`.
> 

Wait, the Before Init event would have had it as null! But now that I changed the injections? You're right, it'll never be null

LambdAurora:
Also, is the client field accessible from the screen parameter? If not then the client parameter should be kept.

EnnuiL:
> Also, is the client field accessible from the screen parameter? If not then the client parameter should be kept.

Yes, but what's the point of having it on the parameter list when screen.getClient() is guaranteed to have a result?
I can restore the `client` param though (or maybe I could transitive AW the field :p)

EnnuiL:
> > Also, is the client field accessible from the screen parameter? If not then the client parameter should be kept.
> 
> Yes, but what's the point of having it on the parameter list when screen.getClient() is guaranteed to have a result?
> I can restore the `client` param though (or maybe I could transitive AW the field :p)

ok no ignore the AW idea it's terrible because someone could set it with it

library/gui/screen/src/main/java/org/quiltmc/qsl/screen/mixin/client/ScreenMixin.java
```diff
@@ -47,35 +48,56 @@ abstract class ScreenMixin implements QuiltScreen {
...
 	@Final
 	private List<Drawable> drawables;
 
 	@Shadow
+	@Nullable
 	protected MinecraftClient client;
 
+	@Shadow
+	private boolean field_42156;
 	@Shadow
```
OroArmor, line 67:
```suggestion

	@Shadow
```

library/gui/screen/src/main/java/org/quiltmc/qsl/screen/api/client/ScreenEvents.java
```diff
@@ -139,13 +137,13 @@ public final class ScreenEvents {
 	@ClientOnly
 	@FunctionalInterface
 	public interface BeforeInit extends ClientEventAwareListener {
-		void beforeInit(Screen screen, MinecraftClient client, int scaledWidth, int scaledHeight);
+		void beforeInit(Screen screen, MinecraftClient client, boolean firstInit);
```
OroArmor, line 147: I would like to see some documentation on what `firstInit` is

OroArmor (changes requested):

library/gui/screen/src/main/java/org/quiltmc/qsl/screen/api/client/ScreenEvents.java
EnnuiL: I'll write some documentation for the events

library/gui/screen/src/main/java/org/quiltmc/qsl/screen/mixin/client/ScreenMixin.java
EnnuiL: augh

PR #291, opened 2023-03-30
Enchantment api

### LINKED ISSUE #13: EnchantmentTarget enum needs a proper fix

Platymemo:
This is a reboot of #20. Below are quoted points by @Vaerian when constructing the original PR.
> Preface: I recognize this is not where the repo's priority is right now, and considering how much of a rough draft this PR is, don't waste your time reviewing it unless you care deeply about the direction of this specific module (i.e. you provided/wanted to provide input on issue #13).

> This PR adds a custom enchantment API similar to the one discussed in #13. I'm making this a draft PR so the discussion around this API can have something to conceptualize what would be required to make these changes a reality. The specific structure of the implementation is still up for debate, but this is generally what it could look like.

> As previously mentioned this is still definitely a draft. It is likely that the majority of the structure of this PR will change including everything from javadocs and class names, to implementation details and method signatures. Don't waste your time nit-picking every individual detail at this point (I'm sure something somewhere is misspelled). Focus more on the big picture.

> Discussion Questions

> What additional information could developers benefit from having in the enchantment context?
Should the equipment slot array system be modified? If so, how? Is there a way to allow expandability of EquipmentSlots for something like Trinkets?
> Which library does the API belong in? Should it stay in content_other?
> How should event callbacks for random enchantments and anvil application work?
> What kind of information should developers have access to when determining the compatibility of enchantments?
> Enchantments use a variety of values such as the amount of bookcases, the "power" of the enchantment table, the level of the enchantment (i.e. I, II, III, etc.), the enchantment tier (the three boxes in the menu), etc. This gets kind of confusing in code. > Is there a way we can standardize these terms both in the API and in the mappings? If so, what should that look like?
> How can the API be modified to be more straightforward to use?
> What kind of enchantment helper methods should the API include for developers to use?
> Should there be some kind of framework/a couple methods to help developers create custom enchantment tables? What are the potential use cases for that? Would enough people use that? Is that in the scope of QSL?
> Closes #13

> This is a question concerning API design. Enchantments are used in a wide variety of places such as random loot, mob equipment with random enchantments, villager tool trades, the enchantment table, and more. Ideally this API would give developers control over whether their enchantment can be added in any of those scenarios. The only issue is that all total that makes up about 6 different contexts in which enchantments can be used. It doesn't really make sense to create a method for every single one of those contexts so instead should we create a common interface for enchantment contexts and allow developers to check if the context is an instance of a specific type of context? The only issue I see with that is each scenario is a little bit different and it doesn't really make much sense for someone to have to write an if else statement with 6 conditions every time they want to write complicated functionality, but then again it's a little weird for someone to have to write implementations for 6 different methods. Additionally, making it a common interface might make it easier for developers to create their own enchantment contexts with special logic.

Currently, I have implemented a common enchanting context, that is then expanded upon by more specific scopes. It currently requires 3 different contexts (base, entity, and player+block) to reach all 6 locations.

Progress

- [x] Support for customizable item groups for custom enchanted books
- [x] Powerful "enchantment targets" (using `Enchantment#isAcceptableItem` and `QuiltEnchantment#isAcceptableContext`)
- [x] Developers have access to the World, PlayerEntity, enchantment power, enchantment level, ItemStack, BlockPos, BlockState, and BlockEntity.
- [x] Enchantment weight can be defined per level/context
- [x] Enchantment weight free from the rarity enum
- [x] Add proper handling of treasure logic to the enchantment context
- [x] Support for more a powerful anvil application context
- [x] Support random enchants for entity spawns, enchanted tool trades, and loot functions
- [x] Custom interface for items to take priority over enchantment compatibility 
- [x] Event callbacks for random enchantment and anvil application
- [ ] ~~Enchantment compatibility takes into account the enchantment level (potentially the entire enchantment context)~~ (superceded by event callbacks which can do just as much)


library/content_other/enchantment/build.gradle
```diff
@@ -0,0 +1,18 @@
+plugins {
+	id("qsl.module")
+}
+
+qslModule {
+	name = "Quilt Enchantment API"
+	moduleName = "enchantment"
+	id = "quilt_enchantment"
+	description = "An "
+	library = "content_other"
```
OroArmor: I would make this part of the item api

library/content_other/enchantment/src/main/java/org/quiltmc/qsl/enchantment/impl/EnchantmentGodClass.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package org.quiltmc.qsl.enchantment.impl;
+
+import org.jetbrains.annotations.ApiStatus;
+
+import org.quiltmc.qsl.enchantment.api.EnchantingContext;
+
+@ApiStatus.Internal
+public final class EnchantmentGodClass {
```
OroArmor: Maybe something like `EnchantmentContextHolder`

OroArmor (changes requested):

library/content_other/enchantment/src/main/java/org/quiltmc/qsl/enchantment/impl/EnchantmentGodClass.java
EnnuiL: how dare you challenge the god class??? :p

Platymemo: https://github.com/QuiltMC/quilt-standard-libraries/issues/13#issuecomment-881958085

It's for i5, and since it's impl, I think we can get away with it

EnnuiL: i mean, it has been almost 2 years since that PR was written; we should aim for the standards of today;
on a glance on that PR though, what is the point of the class though? can't it be merged with QuiltEnchantmentHelper?

Platymemo: It can yeah, I only kept it to hide the threadlocal field and to keep `EnchantmentGodClass`

library/item/enchantment/src/main/java/org/quiltmc/qsl/enchantment/api/QuiltEnchantableItem.java
```diff
@@ -0,0 +1,41 @@
...
+	 * <p>
+	 * This takes highest priority for applying enchantments.
+	 * @param stack the stack containing this item
+	 * @param enchantment the enchantment to apply to this item
+	 * @return {@code true} if the enchantment can be applied, or {@code false} otherwise
+	 */
+	@Contract(pure = true)
+	default boolean canEnchant(ItemStack stack, Enchantment enchantment) {
+		return enchantment.isAcceptableItem(stack);
+	}
```
Platymemo, line 40: I wonder if a TriState might be better? Either override true, override false, or default to enchantment choice. As of now, it's a complete override if the item implements this interface, hence the default implementation, so calling super will allow the enchantment to decide (Note that the enchantment's decision likely doesn't take the other enchantments on the item into account, as there's the separate Enchantment#canCombine(Enchantment)).

TelepathicGrunt: You could do a tristate. Maybe an enum would be clear? Not sure what best way is to convey meaning of overriding behaviors

TelepathicGrunt: On the other hand, the modder does have the enchantment and can just check the enchantment's canEnchant directly if they want to know what the default was

Ampflower:
A random thought given the approach chosen here for handling contexts; it does seem fragile, especially for maintenance as there's a point of failure if the context fails to be cleared for a given thread.

Wouldn't a better approach to be sending an event with the context that could be in use, given it looks to be intended for mods?

There *doesn't* seem to be any concurrency issues with same thread, as the context looks to be created and destroyed within the same method, tho the potential for error is there if it gets spread around multiple threads tho.

library/item/enchantment/src/main/java/org/quiltmc/qsl/enchantment/api/QuiltEnchantableItem.java
EnnuiL, line 1: Shouldn't `QuiltEnchantableItem` and `QuiltEnchantment` become injected interfaces? It does seem like perfect usecases for those, which yeah, a PR from 2021 wouldn't really be able to use

EnnuiL (changes requested):
Not a full review yet, but

library/item/enchantment/src/main/java/org/quiltmc/qsl/enchantment/api/QuiltEnchantableItem.java
Platymemo: So the reason they aren't is that even implementing `QuiltEnchantableItem` or extending `QuiltEnchantment` changes the systems used for the enchantment process, so that definitely has to be opt-in.

EnnuiL: I see

Platymemo:
> A random thought given the approach chosen here for handling contexts; it does seem fragile, especially for maintenance as there's a point of failure if the context fails to be cleared for a given thread.
> 
> Wouldn't a better approach to be sending an event with the context that could be in use, given it looks to be intended for mods?
> 
> There _doesn't_ seem to be any concurrency issues with same thread, as the context looks to be created and destroyed within the same method, tho the potential for error is there if it gets spread around multiple threads tho.

I like events! They're in the planned todo above but I just hadn't gotten to them. But yeah, I wish there were a safer, more robust way to gather the context.

library/item/enchantment/src/main/java/org/quiltmc/qsl/enchantment/api/EnchantmentEvents.java
```diff
@@ -0,0 +1,70 @@
...
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
+import net.minecraft.screen.AnvilScreenHandler;
+
+import org.quiltmc.qsl.base.api.event.Event;
+import org.quiltmc.qsl.enchantment.api.context.EnchantingContext;
+
+public final class EnchantmentEvents {
+	/**
+	 * An event that is called after random enchantments are generated, but before they are selected to be applied on an item.
```
TheGlitch76: Maybe it's a me thing, but it took me way too long to guess what "random enchantments" were for
```suggestion
	 * An event that is called after random enchantments are generated (e.g. for an enchantment table), but before they are selected to be applied on an item.
```

library/item/enchantment/src/main/java/org/quiltmc/qsl/enchantment/api/context/PlayerUsingBlockEnchantingContext.java
Leo40Git, line 1: Maybe add a `BlockState getBlockState()` method here as a shorthand for `context.getWorld().getBlockState(context.getPos())`?

Leo40Git (approved):
LGTM!

library/item/enchantment/src/main/java/org/quiltmc/qsl/enchantment/api/context/PlayerUsingBlockEnchantingContext.java
Leo40Git: nvm, missed comment about removing context objects entirely

NathanX-S:
Any reason this hasn't been merged yet?

OroArmor:
It's still a draft PR, plus it doesn't target the current development version for QSL.

PR #295, opened 2023-04-15
Set the default mixin priority to `500`.

TheGlitch76:
This has a few, somewhat contradictory effects:
- QSL's mixins are applied *before* other mixins
- In the final class, QSLs injection points now ALWAYS occur:
  - After HEAD injections of other mods
  - Before RETURN injections of other mods

This change will **largely eliminate 50/50 incompatibilities with other mods**, Quilt or otherwise. Interactions between a mod and QSL will always either fail or succeed, unless its mixins also apply at 500 priority.

As to why 500 and not 1500 (applying after everything else): The decision was largely arbitrary, but personally I think it makes more sense for the "default mod" to always apply its mixins before the user-installed mods, similar to how Forge works

**IF ACCEPTED, EVERY OPEN PR WHICH ADDS A NEW MODULE WOULD NEED TO BE UPDATED TO INCLUDE THE NEW MIXIN JSON ENTRY**

### DIFF
--- a/library/block/block_content_registry/src/main/resources/quilt_block_content_registry.mixins.json
+++ b/library/block/block_content_registry/src/main/resources/quilt_block_content_registry.mixins.json
@@ -2,6 +2,7 @@
   "required": true,
   "package": "org.quiltmc.qsl.block.content.registry.mixin",
   "compatibilityLevel": "JAVA_17",
+  "mixinPriority": 500,
   "mixins": [
     "AxeItemMixin",
     "EnchantingTableBlockMixin",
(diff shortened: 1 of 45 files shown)


LambdAurora:
Testing this with modpacks would be ideal.

LlamaLad7:
Mixin has a concept of config inheritance specifically for cases like this. You can define a single json with common settings like required, defaultRequire, mixinPriority, etc, and then parent all the other configs to that so they inherit the defaults.
> In the final class, QSLs injection points now ALWAYS occur:
After HEAD injections of other mods
Before RETURN injections of other mods

99% sure that's not the case, lower priority `@Inject`s will run before higher priority ones regardless of injection point.

EnnuiL:
Something tells me that this PR should specifically target 1.20; Blanketcon 23 will be targetting it, and a big modpack is what we really need in order to make sure this works and is worth it

TheGlitch76:
Will reopen when 1.20.1 cycle begins

LambdAurora:
We're already in the 1.20.1 cycle...

TheGlitch76:
Er, I meant 1.20.2

OroArmor:
1.20.2 is in development (i think?) if not we should get on that, recipe api will be interesting (hopefully good)

PR #311, opened 2023-06-05
chat api improvements for probably after 1.20 because i took too long

SilverAndro:
This took a while, probably lots of nitpicking to do (especially with javadoc)

### Changes:
- Test mod is *significantly* improved
- `TypedChatApiHook` is no longer api
- Errors from events now try to show where the error is from
- Support for chat signing
- Support for commands
- Support for message removal (this isnt used in vanilla so, i *think* my impl works?)

### Testing:
This PR is *mostly* tested locally, but because i only have 1 minecraft account and no friends, i couldn't test several sections of this PR, notably:
- Cancelling `INBOUND CLIENT CHAT`
- How this works with commands like `/msg`, especially between players not privy to the conversation
- Probably other things i dont remember


EnnuiL:
i thought i could have merged this pr quickly..
but yeah.. i don't feel safe handling this pr's review process.. i never really felt safe after all of that..

SilverAndro:
would you mind elaborating? im a little confused on what youre referring to or why it involves closing my pr

LambdAurora:
> i thought i could have merged this pr quickly.. but yeah.. i don't feel safe handling this pr's review process..

There is multiple QSL maintainers. This PR can be handled by another team member.

library/management/chat/src/main/java/org/quiltmc/qsl/chat/api/ChatSecurityRollbackSupport.java
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+package org.quiltmc.qsl.chat.api;
+
+/**
+ * Provides rollback support for various classes relating to chat security. This is mostly used in the case of cancelling
+ * signed messages on the client.
+ */
+public interface ChatSecurityRollbackSupport {
```
TheGlitch76, line 23: Should this be an injected interface?

SilverAndro: hm probably

i think i left it out since its only used internally, but since the idea was to allow extending/replacing the signing system it makes sense

library/management/chat/src/main/java/org/quiltmc/qsl/chat/api/MessageChainLookup.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.util.WeakHashMap;
+
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.network.message.MessageChain;
+
+/**
+ * Provides a method to get a {@link ChatSecurityRollbackSupport} from a {@link MessageChain.Packer} or {@link MessageChain.Unpacker}.
+ * This class has been provided as API in the case that a user wishes to replace or extend the builtin chat signing system.
```
TheGlitch76, line 27: Dumb question: Why?
When might a mod want to replace/extend the chat signing system?

SilverAndro: Hm i think that might be some off docs on my part

essentially there isnt any way to automatically track the state behind custom (un)packers since theyre a `FunctionalInterface`, so the idea was that if you create a custom (un)packer for *any* reason, i.e a custom verification system (e2e) or just adding logging, other small tweaks, ect, you could easily register it here

the lookup has to exist anyways, so i didnt see much reason to not just, allow people to register custom ones easily

PR #316, opened 2023-06-25
Make error logging better for bad pack.mcmeta files in mods.

LambdAurora:
This was brought up a few days ago by @lukebemish, when some mods with wrongly formatted `pack.mcmeta` were loaded, causing a log spam without a clear cause.

This should allow to identify more easily mods that have a wrongly formatted `pack.mcmeta` by now logging the resource pack name as well.

The way this implementation is made is needed due to not being able to patch the original Minecraft method as it doesn't have enough context on the source resource pack.


lukebemish:
Ope, you got to it before I did! Looks good to me

EnnuiL (approved):

lukebemish (approved):

Leo40Git (approved):
LGTM
put em on blast

PR #321, opened 2023-07-09
Static Resources API

rinOfTheStars:
## This post is inaccurate to what now exists in the API, please see the comment made before marking this PR as ready for review!
This is an API for Static Resources: user-controlled data that is _accessible early_, _scope-independent_ and _does not change after it is loaded._ 
### Motivations
Due to the properties of Static Resources, they have uses in creating data-driven APIs that affect the game's state in a way that does not change, and/or in a way that must be done far before vanilla's Resource/Data Packs become available. This makes a Static Resources API _essential_ for a future Content Pack API, as the registries said API would need to modify are static in of themselves and are frozen early on in the game's loading process, making regular resources (in the form of Data Packs) highly unsuited for such an API.
### The `Resource` class and mutability of on-disk files
While the first two properties of Static Resources (early access and scope-independence) are already implemented in the current draft, the third property (immutability) is a far more complicated affair, as the needs of each consumer of Static Resources will likely vary heavily in terms of how many times said consumer accesses said resources. Since `Resource` objects are just pointers to a file and this API primarily returns `Resource` objects, it is possible for the contents of the file to be modified at runtime, which may cause unexpected behavior for the consumer. To be honest, I (rin) am not sure whether or not ensuring strict immutability of the data provided by `Resource` instances is something this API should even be concerned with doing—this will need discussion in the Quilt discord and/or in this PR's comments.
### The API itself: modder & QSL-side
The API is found in the `core/resource_loader` package, and is primarily accessed through two `StaticResourceLoaderImpl` objects. Said objects are accessible outside of the API via the static interface method `StaticResourceLoader#get`, to which a vanilla `ResourceType` object is provided—while this currently (subject to change if requested—this could easily be enforced) only dictates _where_ the API gets Resources from, it should be treated the same as in vanilla, where `ResourceType.CLIENT_RESOURCES` are only available on the client, and `ResourceType.SERVER_DATA` in all scopes. 

The implementation of `StaticResourceLoader` (and thus the two objects provided, depending on whether server data or client resources are requested) serves as a fairly thin wrapper around another API-internal class, `StaticResourceManager`, which is currently little more than a re-implementation of the vanilla `MultiPackResourceManager` class. This was done as `MultiPackResourceManager` is an essential part of the very much not static vanilla resource loading process, and is targeted by mixins changing things about said process—using an object of that class (or of a child of that class) for this API would lead to it being affected by those mixins. 

Currently, consumers of Static Resources are responsible for loading the files provided `Resource` objects point to themselves.

### The API itself: the stuff exposed to end-users
Static Resources are stored within the `static/resources` and `static/data` folders: the former being for client-side resources/data and the latter for server-side resources/data. Basically, these folders have a same purpose to vanilla's `resources` folder and the `datapacks` folder in each world save, respectively. Like vanilla resource packs and data packs, the files placed in these folders _must_ be contained in packs as well—though the distinction between resource packs and data packs is less concrete with static packs than it is with vanilla's packs. 

### TODO Checklist before this API is ready for review and merger

- [X] Implement the Static Resource Loader (`StaticResourceLoader`/`StaticResourceManagerImpl`) and the manager backing it (`StaticResourceManager`) 
- [ ] Write any additional API infrastructure needed for the Static Resource Loader. **Input is needed here!**
- [ ] Write test mod stuff for the Static Resource Loader
- [ ] Write Javadocs
#### If data immutability should be ensured by the API:
- [ ] Create some form of `CachedResource` class or something along those lines, some form of hashing of the loaded data, etc
- [ ] Figure out how to even ensure said disk-loaded data immutability and implement it


library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/StaticResourceLoader.java
```diff
@@ -0,0 +1,60 @@
...
+	Set<String> getNamespaces();
+	List<ResourcePack> getPacks();
+	Map<Identifier, List<Resource>> findAllResources(String startingPath, Predicate<Identifier> pathFilter);
+	Map<Identifier, Resource> findResources(String startingPath, Predicate<Identifier> pathFilter);
+	Stream<ResourcePack> streamPacks();
+	Optional<Resource> getResource(Identifier id);
+	Resource getResourceOrThrow(Identifier id) throws FileNotFoundException;
+	InputStream open(Identifier id) throws IOException;
+	BufferedReader openAsReader(Identifier id) throws IOException;
+	Map<Identifier, JsonElement> findJsonObjects(String namespace, String startingPath);
```
falkreon: We've tightly coupled search and decoding here. Even as a helper method, this is awkward. Consider something like
```java
findObjects(Identifier idWithStartingPath).getAsJson();
```

and while we're at it, consider using the quilt json5 parser instead of the Gson one.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/StaticResourceLoader.java
```diff
@@ -0,0 +1,60 @@
...
+ * Unlike 'regular' resources loaded through vanilla's resource loader and {@link ResourceLoader},
+ * static resources are accessible <i>as soon as a StaticResourceLoader instance exists to get them,</i>
+ * and are
+ */
+public interface StaticResourceLoader {
+
+	static StaticResourceLoader get(ResourceType type){
+		return StaticResourceLoaderImpl.get(type);
+	}
+	List<Resource> getAllResources(Identifier id);
```
falkreon: Seems like copying vanilla respacks and datapacks for no benefit. I'd much rather have a good search system and let people search recursively/non-recursively at root than have this method exclusively for a case basically no one needs.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/StaticResourceManager.java
```diff
@@ -0,0 +1,176 @@
...
+import java.io.BufferedReader;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.*;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+
+/**
+ * Basically a glorified re-implementation of {@link MultiPackResourceManager} to avoid mixins to said class effecting static data loading.
```
falkreon: affecting*

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/StaticResourceManager.java
```diff
@@ -0,0 +1,176 @@
...
+
+	@Override
+	public void close() {
+		this.resourcePacks.forEach(ResourcePack::close);
+	}
+
+	/**
+	 * Based heavily on {@link MultiPackResourceManagerMixin#quilt$recomputeNamespaces()}
+	 */
+	private void computeNamespaces(){
```
falkreon: again, seems like a lot of moving parts for what's essentially a copy of a little-used feature from vanilla.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/StaticResourceLoader.java
```diff
@@ -0,0 +1,60 @@
...
+	Set<String> getNamespaces();
+	List<ResourcePack> getPacks();
+	Map<Identifier, List<Resource>> findAllResources(String startingPath, Predicate<Identifier> pathFilter);
```
falkreon: Is there a reason the static resource doesn't contain provenance information? Is it just to reuse the Resource class? This could bake down into a List or a Stream if we wanted it to, and users could mapreduce it if they actually cared about the properties that make it a map here.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/StaticResourceLoader.java
```diff
@@ -0,0 +1,60 @@
...
+	Map<Identifier, Resource> findResources(String startingPath, Predicate<Identifier> pathFilter);
+	Stream<ResourcePack> streamPacks();
+	Optional<Resource> getResource(Identifier id);
+	Resource getResourceOrThrow(Identifier id) throws FileNotFoundException;
+	InputStream open(Identifier id) throws IOException;
```
falkreon: Probably tight coupling. Consider
```java
findFirst(id).open();
```

falkreon (commented):
I would have designed this a lot differently (see: StaticData mod). There are some core issues I'd like to address here before I'm happy with the API.

Dividing things into Static *Resources* versus *Data* is an interesting idea that I'm going to need to put some thought into, I'll definitely have more to say later on that.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourceLoader.java
```diff
@@ -50,6 +51,19 @@ public interface ResourceLoader {
...
+	/**
+	 * Gets the static resource manager for a given type.<br>
+	 * Static resources are different from regular resources in that they
+	 * are available as soon as the manager is requested and are (theoretically) immutable.<br>
+	 * It is <i>highly</i> recommended for mods to personally cache data they get from the provided manager in order to
+	 * prevent changes to userspace packs modifying data at runtime.
+	 * @param type the given resource type
+	 * @return the static resource manager instance
+	 */
+	static @NotNull StaticResourceManager getStaticResourceManager(@NotNull ResourceType type){
```
lukebemish: As `StaticResourceManager` is an internal type, this should probably return just a `ResourceManager` to not expose internals

rinOfTheStars:
After quite a bit of refactoring work, I'd say this is finally ready for review. Let's go over what's changed since the first post:
- `StaticResourceLoader` no longer exists, and is simply part of the main `ResourceLoader` class and impl.

- `StaticResourceManager` is now a child class of `MultiPackResourceManager` that implements a wrapper interface—the wrapper interface is also implemented by `MultiPackResourceManagerMixin` to prevent cast-related exceptions. Said wrapper interface's helper method is used in said mixin to check if the `MultiPackResourceManager` object mixin methods are currently running in are `StaticResourceManager` objects, at which point they immediately return (if a `void` method) or pass the original value (if a `@ModifyVariable` mixin method). _Future mixins added to `MultiPackResourceMangerMixin` should use this wrapper to prevent them from being ran within `StaticResourceManager`s._

- No helper methods for working with JSON exist anymore—these will probably be a future PR, and work with all `Resource` objects, not just static ones.

- Pack loading actually works now, and tests exist to prove that it works.

Thank you to everyone who gave feedback and advice in the discord and on here during the process of developing this PR, it wouldn't have been possible to write it otherwise.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourceLoader.java
rinOfTheStars: Thanks, wouldn't have caught that w.o this comment.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourceLoader.java
```diff
@@ -50,6 +51,19 @@ public interface ResourceLoader {
...
+	/**
+	 * Gets the static resource manager for a given type.<br>
+	 * Static resources are different from regular resources in that they
+	 * are available as soon as the manager is requested and are (theoretically) immutable.<br>
+	 * It is <i>highly</i> recommended for mods to personally cache data they get from the provided manager in order to
+	 * prevent changes to userspace packs modifying data at runtime.
+	 * @param type the given resource type
+	 * @return the static resource manager instance
+	 */
+	static @NotNull MultiPackResourceManager getStaticResourceManager(@NotNull ResourceType type){
```
LambdAurora: Any reason to return `MultiPackResourceManager` and not just `ResourceManager`?
So far `ResourceManager` is usually used for APIs that provide a resource manager.

Also, given the API structure, this should not be `static` but be something to implement directly on the `ResourceLoaderImpl` class like every other method present in this interface.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceLoaderImpl.java
```diff
@@ -98,13 +105,11 @@ public final class ResourceLoaderImpl implements ResourceLoader {
 			.toBooleanOrElse(QuiltLoader.isDevelopmentEnvironment());
 	private static final boolean DEBUG_RELOADERS_ORDER = TriState.fromProperty("quilt.resource_loader.debug.reloaders_order")
 			.toBooleanOrElse(false);
-
```
LambdAurora, line 101: I would highly appreciate to not remove blank lines that are present for readability.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceLoaderImpl.java
```diff
@@ -523,4 +536,36 @@ public static void appendLanguageEntries(@NotNull Map<String, String> map) {
 			}
 		}
 	}
+
+	/* Static pack stuff */
+	private static List<ResourcePack> findUserStaticPacks() {
+		List<ResourcePack> returnList = new ArrayList<>();
+		File directoryFile = QuiltLoader.getGameDir().resolve(STATIC_PACK_ROOT).toFile();
+		File[] potentialPackFiles = directoryFile.listFiles(filterFile -> {
```
LambdAurora: Given that most things use the NIO Path API, it might make more sense to use the `Files` NIO API to search packs.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/StaticResourceManagerWrapper.java
```diff
@@ -0,0 +1,28 @@
...
+
+import net.minecraft.resource.MultiPackResourceManager;
+import org.jetbrains.annotations.ApiStatus;
+
+/**
+ * Wrapper/duck interface for {@link MultiPackResourceManager} instances.
+ * @see StaticResourceManager
+ */
+@ApiStatus.Internal
+public interface StaticResourceManagerWrapper {
```
LambdAurora: This is useless, a simple `instanceof StaticResourceManager` is already sufficient for checking.

library/core/resource_loader/src/testmod/java/org/quiltmc/qsl/resource/loader/test/StaticResourcesTestMod.java
```diff
@@ -0,0 +1,60 @@
...
+public class StaticResourcesTestMod implements ModInitializer {
+	public static final Logger LOGGER = LoggerFactory.getLogger("QuiltStaticResourcesTest");
+	@Override
+	public void onInitialize(ModContainer mod) {
+
+		try {
+			MultiPackResourceManager clientManager = ResourceLoader.getStaticResourceManager(ResourceType.CLIENT_RESOURCES);
+			Resource cronch = clientManager.getResource(new Identifier("cronch", "test_client")).get();
+			BufferedReader readerCronch = cronch.openBufferedReader();
+			LOGGER.error(readerCronch.readLine() + "(Reading this line should be impossible!)");
```
LambdAurora: Please use the var-args system of the logger to avoid passing unsafe strings.
Given this is a test mod it's not as bad, but they're also references on how to use an API, so they should not accidentally give bad practices.

LambdAurora (changes requested):
If you are using Intellij, please import the checkstyle file as the code style for this project as several part of the PR are not compliant with the code style of QSL.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourceLoader.java
rinOfTheStars: The former I agree with and will be changed accordingly in the next commit, but the latter I have to disagree with—the creation and storage of the `StaticResourceManager`s that back the Static Resources API are both entirely static-scoped, so it's both less verbose and more aligned with the data structure of the API itself to get the static resource managers in a static method.

If this is a sticking point I can change it, but I feel like it makes more sense this way.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceLoaderImpl.java
rinOfTheStars: Fixed in next commit, hopefully?

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/StaticResourceManagerWrapper.java
rinOfTheStars: Screen Shot 2023-07-15 at 8 31 23 PM

Would the mixin need to `implements ResoruceManager` for that to work? I'm presuming yes, and am going to test it while working on the upcoming commit—if it doesn't work I'll post proof as screenshot or video.

rinOfTheStars: Couldn't get it to work with both `implements AutoCloseableResourceManager` & `implements ResourceManager`. I have exactly zero clue why this is the case—logic says it should work, but it just doesn't.

Screen Shot 2023-07-15 at 8 37 21 PM
Screen Shot 2023-07-15 at 8 37 48 PM

library/core/resource_loader/src/testmod/java/org/quiltmc/qsl/resource/loader/test/StaticResourcesTestMod.java
rinOfTheStars: Fixed in next commit.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/StaticResourceManagerWrapper.java
lukebemish: You need to cast to object before you instanceof. Java is smart and finds impossible conditions

rinOfTheStars: Yup, turns out it was an 'IDEA and the compiler don't understand mixins' moment and not an actual error—this has been fixed and thus said wrapper interface will be gone in next commit.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceLoaderImpl.java
rinOfTheStars: Tested implementing it that way, and it both works and is slightly cleaner overall so it'll be swapped to using `Files.walk` in the next commit where `listFiles` is currently.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourceLoader.java
```diff
@@ -50,6 +51,19 @@ public interface ResourceLoader {
...
+	/**
+	 * Gets the static resource manager for a given type.<br>
+	 * Static resources are different from regular resources in that they
+	 * are available as soon as the manager is requested and are (theoretically) immutable.<br>
+	 * It is <i>highly</i> recommended for mods to personally cache data they get from the provided manager in order to
+	 * prevent changes to userspace packs modifying data at runtime.
+	 * @param type the given resource type
+	 * @return the static resource manager instance
+	 */
+	static @NotNull ResourceManager getStaticResourceManager(@NotNull ResourceType type) {
```
rinOfTheStars, line 66: With the most recent commit, everything should be in order—bar the de-`static`ing of this method, which I'm opposed to, though if it's a sticking point for @LambdAurora I can change it. See  https://github.com/QuiltMC/quilt-standard-libraries/pull/321#discussion_r1264587712 for the reasoning on why that hasn't been done.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourceLoader.java
lukebemish: I'm inclined to agree that this should be static, given that a static resource manager is not tied to a `ResourceLoader` in any way

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceLoaderImpl.java
```diff
@@ -523,4 +536,32 @@ public static void appendLanguageEntries(@NotNull Map<String, String> map) {
...
+	private static List<ResourcePack> findUserStaticPacks() {
+		List<ResourcePack> returnList = new ArrayList<>();
+		try (Stream<Path> walkStream = Files.walk(QuiltLoader.getGameDir().resolve(STATIC_PACK_ROOT), 1)) {
+			for (Path path : walkStream.toList()) {
+				File pathAsFile = path.toFile();
+				String n = calcUserspacePackName(pathAsFile);
+				if (pathAsFile.isFile()) {
+					if (pathAsFile.toPath().toString().endsWith(".zip")) {
+						returnList.add(new ZipResourcePack(n, pathAsFile, false));
+					} else if (!pathAsFile.getName().equals(".DS_Store")) {
```
falkreon: if we're going to silently ignore filesystem helper files "like .DS_Store" instead of ignoring specifically .DS_Store, can we do something like
```java
private static final Set<String> IGNORED_FILES = Set.of(".DS_Store");

...

} else if (!IGNORED_FILES.contains(pathAsFile.getName())) {
  ...
}
```

Otherwise, if there will definitely never be any other fs helper files, we could just delete the comment

Cypher121: are thumbs.db files still around?

lukebemish: What does vanilla do for detecting packs in the resource packs folder? I feel like we should probably copy that

rinOfTheStars: Vanilla doesn't actually ignore .DS_Store files and similar (it throws an error if they're inside packs!), but it silently doesn't load loose files as a pack, same as is done here. Telling users that loose files aren't supported is a thing since other static data implementations do allow for them, so it felt pertinent to say 'hey this API doesn't do that'.

kb-1000: It's worth noting that both the macOS and Windows filesystems are case-insensitive by default, so a plain equals won't suffice, either. And... there are many more files like this, including but not limited to `._*` on macos, `.directory` on some Linux DEs, `desktop.ini` on windows, etc etc

rinOfTheStars: Anyways, the fs helper file thing has been reimplemented with `thumbs.db` and `.DS_Store` in the set, and has a message to make an issue on the main github if an OS filesystem helper is warned of as a loose file

rinOfTheStars: > It's worth noting that both the macOS and Windows filesystems are case-insensitive by default, so a plain equals won't suffice, either. And... there are many more files like this, including but not limited to `._*` on macos, `.directory` on some Linux DEs, `desktop.ini` on windows, etc etc

Yaaay... eating some food rn but will go back and do another pass on this afterwards—thankfully `Set` has a case-insensitive contains method built-in, so that part won't be too big of an issue...

rinOfTheStars:
Unless anyone else has suggestions for additions again, this should be ready for re-review by @LambdAurora

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceLoaderImpl.java
```diff
@@ -549,8 +549,20 @@ private static List<ResourcePack> findUserStaticPacks() {
...
-							//Ignores filesystem helper files, currently ignoring .DS_Store and thumbs.db
+						boolean fsHelper = false;
+						for (String fs : STATIC_USERSPACE_IGNORED_FILES) {
+							// Implementation detail: ._* and (some) *nix-derivative OSes (macOS, Linux distros, etc)
+							// ._* is a rare *nix filesystem helper file created to store file information normally placed in an extended attribute
+							// on HFS+ (Apple Native, typically found under macOS) or UFS (other *nix OSes) filesystems when interfacing with a filesystem
+							// that does NOT support such extended attributes (for instance, a FAT32 drive). Since these can have names of arbitrary lengths,
+							// the only way to detect them is via the shared prefix.
+							// Source: https://apple.stackexchange.com/a/14981
+							if (fs.equalsIgnoreCase(pathAsFile.getName()) || pathAsFile.getName().startsWith("._")) {
```
embeddedt: There is no benefit I see to repeating the `startsWith` check multiple times within the loop; guessing this is an oversight?

rinOfTheStars: Yeah, it is—I'll move it outside of the loop in a commit in a moment

rinOfTheStars: Fixed as of latest commit.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceLoaderImpl.java
```diff
@@ -523,4 +537,50 @@ public static void appendLanguageEntries(@NotNull Map<String, String> map) {
 			}
 		}
 	}
+
+	/* Static pack stuff */
+	private static List<ResourcePack> findUserStaticPacks() {
+		List<ResourcePack> returnList = new ArrayList<>();
```
LambdAurora, line 547:
```suggestion
		List<ResourcePack> returnList = new ArrayList<>();

```

By the way, we allow the usage of `var` for variables that use `new`. Letting you the choice of using it or not as I feel like it'd be weird to nitpick ahah.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceLoaderImpl.java
```diff
@@ -523,4 +537,50 @@ public static void appendLanguageEntries(@NotNull Map<String, String> map) {
...
+		try (Stream<Path> walkStream = Files.walk(QuiltLoader.getGameDir().resolve(STATIC_PACK_ROOT), 1)) {
+			for (Path path : walkStream.toList()) {
+				File pathAsFile = path.toFile();
+				String n = calcUserspacePackName(pathAsFile);
+				if (pathAsFile.isFile()) {
```
LambdAurora, line 554:
```suggestion

				if (pathAsFile.isFile()) {
```

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceLoaderImpl.java
```diff
@@ -523,4 +537,50 @@ public static void appendLanguageEntries(@NotNull Map<String, String> map) {
...
+
+						if (!fsHelper) {
+							for (String fs : STATIC_USERSPACE_IGNORED_FILES) {
+								if (fs.equalsIgnoreCase(pathAsFile.getName())) {
+									fsHelper = true;
+									break;
+								}
+							}
+						}
+						if (!fsHelper) {
```
LambdAurora:
```suggestion
```

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceLoaderImpl.java
```diff
@@ -523,4 +537,50 @@ public static void appendLanguageEntries(@NotNull Map<String, String> map) {
...
+						}
+					}
+				} else if (pathAsFile.isDirectory()) {
+					returnList.add(new NioResourcePack(n, path, false));
+				}
+			}
+		} catch (IOException eio) {
+			LOGGER.error("IO Exception thrown while loading userspace static packs: {}", eio.toString());
+		}
+		return returnList;
```
LambdAurora, line 578:
```suggestion

		return returnList;
```

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceLoaderImpl.java
```diff
@@ -523,4 +537,50 @@ public static void appendLanguageEntries(@NotNull Map<String, String> map) {
...
+	}
```
LambdAurora, line 579:
```suggestion
	}

```

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/StaticResourceManager.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.List;
+
+import org.jetbrains.annotations.ApiStatus;
+
+import net.minecraft.resource.MultiPackResourceManager;
+import net.minecraft.resource.ResourceType;
+import net.minecraft.resource.pack.ResourcePack;
+
+/**
+ * Exists solely for determining whether a manager is involved in static resource loading or regular resource loading.<br>
```
LambdAurora:
```suggestion
 * Exists solely for determining whether a manager is involved in static resource loading or regular resource loading.
 * <p>
```

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/mixin/MultiPackResourceManagerMixin.java
```diff
@@ -69,21 +71,25 @@ public abstract class MultiPackResourceManagerMixin implements QuiltMultiPackRes
...
 	public void quilt$appendTopPacks() {
+		if (((Object) this) instanceof StaticResourceManager) return;
 		if (!(this.packs instanceof ArrayList<ResourcePack>)) {
 			this.packs = new ArrayList<>(this.packs);
 		}
-
 		ResourceLoaderImpl.get(this.quilt$type).appendTopPacks((MultiPackResourceManager) (Object) this, this.packs::add);
 	}
-
+	@SuppressWarnings("ConstantConditions")
```
LambdAurora:
```suggestion

	@SuppressWarnings("ConstantConditions")
```

LambdAurora (changes requested):
Please space out your code for readability.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourceLoader.java
```diff
@@ -50,6 +51,19 @@ public interface ResourceLoader {
...
 	}
 
+	/**
+	 * Gets the static resource manager for a given type.<br>
+	 * Static resources are different from regular resources in that they
+	 * are available as soon as the manager is requested and are (theoretically) immutable.<br>
+	 * It is <i>highly</i> recommended for mods to personally cache data they get from the provided manager in order to
+	 * prevent changes to userspace packs modifying data at runtime.
+	 * @param type the given resource type
+	 * @return the static resource manager instance
```
LambdAurora:
```suggestion
	 * Gets the static resource manager for a given type.
	 * <p>
	 * Static resources are different from regular resources in that they
	 * are available as soon as the manager is requested and are (theoretically) immutable.
	 * <p>
	 * It is <i>highly</i> recommended for mods to personally cache data they get from the provided manager in order to
	 * prevent changes to user-space packs modifying data at runtime.
	 * <p>
	 * After the end of the initial boot process, as-in reaching the end of the start process of a server or of the client, the static resource manager may be closed and not re-used.
	 *
	 * @param type the type of resources expected from this resource manager
	 * @return the static resource manager instance
```

I knew I forgot something.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceLoaderImpl.java
rinOfTheStars: Should not be removed. The for loop just above this is able to set `fsHelper` to true and break out of the loop when it does so—removing the extra `if (!fsHelper)` check would cause the loose file log error to be displayed whenever the for loop is executed at all, not when the for loop never sets `fsHelper` to true. There's probably a more elegant way to write this segment, but this works while removing the second check does not.

LambdAurora: Then the ignored file list can be all lower case and change the condition to `pathAsFile.getName().starts with("._") || STATIC-USERSPACE_IGNORED_FILES.contains(pathAsFile.getName().toLowercase())`.

rinOfTheStars: Reimplemented in a very similar way to what you suggested as of latest commit, and so no longer uses a for loop or two checks of the same boolean as was the case before.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/impl/ResourceLoaderImpl.java
```diff
@@ -557,17 +560,7 @@ private static List<ResourcePack> findUserStaticPacks() {
...
-						if (!fsHelper) {
-							for (String fs : STATIC_USERSPACE_IGNORED_FILES) {
-								if (fs.equalsIgnoreCase(pathAsFile.getName())) {
-									fsHelper = true;
-									break;
-								}
-							}
-						}
-						if (!fsHelper) {
+						if (!pathAsFile.getName().startsWith("._") && !STATIC_USERSPACE_FILESYSTEM_HELPERS.contains(pathAsFile.getName().toLowerCase())) {
```
falkreon: More nitpicks:

I would strongly suggest not leaving this up to the system-default Locale. These are locale-insensitive Strings, so you want to use `toLowerCase(Locale.ROOT)`

rinOfTheStars: I wouldn't have thought of that, thanks!
Fixed as of latest commit.

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourceLoader.java
lukebemish: Given that static resource managers are unrelated to the normal resource loaders, really - which handle non static data - I am also opposed to de-`static`ing this

rinOfTheStars:
Could we potentially get some more movement on this soonish, now that QFAPI sync is back in order?

OroArmor (approved):

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourceLoader.java
```diff
@@ -50,6 +51,24 @@ public interface ResourceLoader {
...
+	/**
+	 * Gets the static resource manager for a given type.
+	 * <p>
+	 * Static resources are different from regular resources in that they
+	 * are available as soon as the manager is requested and are (theoretically) immutable.
+	 * <p>
+	 * It is <i>highly</i> recommended for mods to personally cache data they get from the provided manager in order to
+	 * prevent changes to user-space packs modifying data at runtime.
+	 * <p>
+	 * After the end of the initial boot process, as-in reaching the end of the start process of a server or of the client, the static resource manager may be closed and not re-used.
```
TheGlitch76: > the static resource manager may be closed and not re-used

I don't know what this means. Is it always closed, or just sometimes? Who can/does close it (mods, minecraft, qsl)? Does this mean "may [be closed] and [not re-used]" or "may [be closed and not re-used]"? Please clarify

TheGlitch76 (changes requested):

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/ResourceLoader.java
lukebemish: Actually, I'm even more confused; why do we need to close it at all? Why not leave it open indefinitely?

rinOfTheStars: Good catch—that looks to be an artifact of an older version of the API. Going to fix this (and even add code to prevent said closures, likely an override of the close method in the static resource manager class) later tonight.

rinOfTheStars: Fixed.

rinOfTheStars:
Should be ready for 1.20.2 now—I have absolutely no clue why there's conflicts in `ResourceLoaderImpl`, however. Both supposed conflicts seem to just be adding stuff? Not sure why its marking them as conflicts, but it took quite a few attempts to get this to merge at all locally, as it tended to get stuck in an infinite loop of trying to resolve the same conflicts over and over again.

OroArmor (approved):

rinOfTheStars:
Ready for 1.20.2 again if all checks pass—ResourceLoaderImpl's conflicts are git being weird just like last time

OroArmor:
this might be hard but i would try rebasing your PR off of 1.20.2. That should hopefully fix the conflict

rinOfTheStars:
Have tried doing that and it gets stuck in the infinite loop usually. Might try doing it again later

PR #326, opened 2023-07-20
Flatten Bundled Packets

TropheusJ:
This allows for nesting PacketBundles inside each other. This is currently used in the Entity Networking module for entities with extended spawn data. This ends up looking like this:
- root packet bundle
  - spawn packet bundle
    - base spawn packet
    - extended spawn data
  - tracked data
  - attributes
...

Without this PR, this causes a terrible error deep in packet code caused by trying to send a bundle packet as a normal packet.
This is quite an important fix, as the extended spawn data API currently does not work on dedicated servers.

I have not tested this in production yet.


LambdAurora (approved):

TropheusJ:
testmods work on a production server and client.

Issue #332, opened 2023-07-28
Nether in newly generated world does not gen properly if Nether was already visited in previous world


rinOfTheStars:
See https://github.com/QuiltMC/quilted-fabric-api/issues/109, confirmed this is a bug in QSL in a dev environment. 

No clue what's actually causing this, but considering it only happens when the Nether has been generated in a previous world, something probably isn't getting cleared from memory properly?

rinOfTheStars:
#333 should be a fix for this

EnnuiL:
To be fixed on the next update!

PR #333, opened 2023-07-29
Remove nether param. range list caching, fixing #332

rinOfTheStars:
Don't exactly understand why this works, but presumably the value stored in `quilt$CACHED_PARAMETER_RANGE_LIST` wasn't being nulled when a world was unloaded, leading to the funky behavior on second world's nether load
There's probably a better way to impl. this fix, but since @EnnuiL mentioned in the discord she might change this part of the Biomes API heavily soon anyways, this'll do for now as a quickfix


EnnuiL:
While I've went with the approach of switching back to an older method, I appreciate the efforts here! It did show that the newer method could be fixed and that is important, because I still don't discard a future where it might end being needed again :p
I should perhaps get deeper into like, a profiling habit though, in order to actually measure potential lag and such

PR #342, opened 2023-09-01
Add missing null handling to `GroupResourcePack`

soir20:
Currently, two methods in `GroupResourcePack` will throw `NullPointerException`s if there is no pack for the given namespace. I've tweaked these methods so that the `GroupResourcePack` won't crash the game in these circumstances.

This isn't an issue when no other mods are present because vanilla also scans packs by namespace. However, I develop a mod that originally listed resources in all packs, regardless of whether they contained a particular namespace, which caused the game to crash.

There's an existing null check in `open()`; this PR makes null handling consistent throughout the whole `GroupResourcePack`.

### DIFF
--- a/library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/GroupResourcePack.java
+++ b/library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/GroupResourcePack.java
@@ -77,7 +77,13 @@ public abstract class GroupResourcePack implements ResourcePack {
 	 * @return the list of the matching resource packs
 	 */
 	public @UnmodifiableView List<? extends ResourcePack> getPacks(String namespace) {
-		return Collections.unmodifiableList(this.namespacedPacks.getOrDefault(namespace, Collections.emptyList()));
+		var packs = this.namespacedPacks.get(namespace);
+
+		if (packs != null) {
+			return Collections.unmodifiableList(packs);
+		}
+
+		return Collections.emptyList();
 	}
 
 	/**
--- a/library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/GroupResourcePack.java
+++ b/library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/GroupResourcePack.java
@@ -77,7 +77,7 @@ public abstract class GroupResourcePack implements ResourcePack {
 	 * @return the list of the matching resource packs
 	 */
 	public @UnmodifiableView List<? extends ResourcePack> getPacks(String namespace) {
-		return Collections.unmodifiableList(this.namespacedPacks.get(namespace));
+		return Collections.unmodifiableList(this.namespacedPacks.getOrDefault(namespace, Collections.emptyList()));
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


LambdAurora:
I'm curious to see how you're listing those resources, in theory NPE should never be thrown if resources are interacted with correctly.

soir20:
My mod adds a resource pack that scans other packs for certain resources. I `streamResourcePacks()` from the `ResourceManager` and query resources from them directly to prevent infinite recursion and to query root resources.

Specifically, I was calling [`listResources()` directly on the resource packs](https://github.com/MoreMcmeta/core/blob/v1.20.1-4.2.1/common/src/main/java/io/github/moremcmeta/moremcmeta/impl/client/adapter/PackResourcesAdapter.java#L114), rather than going through the `ResourceManager`. This has since been fixed to check that the pack has the given namespace, so there's no longer a crash. However, I don't think Quilt should crash in these circumstances because vanilla packs do not; there should at least be a descriptive exception if this is intended.

LambdAurora (approved):

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/GroupResourcePack.java
```diff
@@ -77,7 +77,7 @@ public GroupResourcePack(@NotNull ResourceType type, @NotNull List<? extends Res
 	 * @return the list of the matching resource packs
 	 */
 	public @UnmodifiableView List<? extends ResourcePack> getPacks(String namespace) {
-		return Collections.unmodifiableList(this.namespacedPacks.get(namespace));
+		return Collections.unmodifiableList(this.namespacedPacks.getOrDefault(namespace, Collections.emptyList()));
```
LambdAurora: When merging (or before merge) this probably should be split to use an if instead (get and store into a local, if not null then return the unmodifiable call, else return emptyList()).

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/GroupResourcePack.java
```diff
@@ -131,9 +137,11 @@ public void listResources(ResourceType type, String namespace, String startingPa
 			ResourcePack.ResourceConsumer consumer) {
 		var packs = this.namespacedPacks.get(namespace);
 
-		// Iterating backwards as higher-priority packs are placed at the end.
-		for (var pack : packs) {
-			pack.listResources(type, namespace, startingPath, consumer);
+		if (packs != null) {
+			// Iterating backwards as higher-priority packs are placed at the end.
+			for (var pack : packs) {
```
Pyrofab, line 142: Is this comment actually relevant ? It seems copy-pasted from above

soir20: It's the same `for` statement, just moved under the `null` check, so the comment is still relevant.

OroArmor:
"Merged" with [50309c1](https://github.com/QuiltMC/quilt-standard-libraries/commit/50309c1e1eb25dfeb35f5696ba841ef505f0b1ca)

PR #336, opened 2023-08-15
[1.20.2] Initial port work; port resource loader

TheGlitch76:

PR #346, opened 2023-09-17
1.20.2

OroArmor:
Includes the work from #336.

Might add #337.

Still a WIP but its good progress.


OroArmor:
Stuff still needing to be tested:
- [x] Fabric Registry Sync API. This changed with 1.20.2
- [x] Status Effect Data Fixer

Stuff that still could be done:
- [x] Add #337

OroArmor:
@LambdAurora if you could push a commit with the licenses updated that would be great since I am stuck on windows.

EnnuiL:
> @LambdAurora if you could push a commit with the licenses updated that would be great since I am stuck on windows.

I'm a Linux user as well btw

TheGlitch76 (approved):

OroArmor:
> I'm a Linux user as well btw

Ah yes there was another I couldn't remember.

Codetoil:
`Caused by: org.gradle.api.GradleException: License header checks have failed on 12 out of 71 files.`
Might want to take a peek at that.

OroArmor:
Build passes now

build-logic/src/main/java/qsl/internal/extension/QslExtension.java
```diff
@@ -4,9 +4,7 @@
 import qsl.internal.Versions;
 
 public class QslExtension {
-	private static final String RESOLVE_VINEFLOWER_TASK = "resolveVineflower";
-	private static final String GEN_SOURCES_WITH_VINEFLOWER_TASK = "genSourcesWithVineflower";
-	private static final String[] TASKS_TO_DISABLE = {RESOLVE_VINEFLOWER_TASK, "genSourcesWithFernFlower"};
+	private static final String[] TASKS_TO_DISABLE = {"genSourcesWithFernFlower"};
```
EnnuiL: Why remove the exclusion? I believe we'll still need it in order to prevent the silly case of Decompiling All QSL Modules' Minecraft

library/core/networking/src/main/java/org/quiltmc/qsl/networking/api/ServerPlayNetworking.java
```diff
@@ -60,6 +80,7 @@ public final class ServerPlayNetworking {
 	 * @see ServerPlayNetworking#unregisterGlobalReceiver(Identifier)
 	 * @see ServerPlayNetworking#registerReceiver(ServerPlayNetworkHandler, Identifier, ChannelReceiver)
 	 */
+	@Deprecated
```
EnnuiL, line 85: still not a fan of these `@Deprecated` methods, but I'll leave them alone;
just make sure to have `@deprecated` notes on the javadoc

EnnuiL (changes requested):

build-logic/src/main/java/qsl/internal/extension/QslExtension.java
OroArmor: Loom 1.4 adds vineflower support so i can remove the special casing that we do for it

library/core/networking/src/main/java/org/quiltmc/qsl/networking/api/ServerPlayNetworking.java
OroArmor: Yeah im still not sure what i want to do with these. I originally had these marked for removal but i didnt like that.

build-logic/src/main/java/qsl/internal/extension/QslExtension.java
EnnuiL: Yes, I can see why the "make `genSourcesWithVineflower` depend on `resolveVineflower`" part was removed, but what I don't understand is why it was removed from `TASKS_TO_DISABLE`;
It is there in order to prevent a `./gradlew genSources` task from decompiling each module's Minecraft, so it still has to be there

TheGlitch76: This task existed on loom 1.3 too

OroArmor: Ah ok i see we disabled the resolve task which then in turn disables the gen sources task. ill fix this

OroArmor:
Ready for review!

Codetoil:
woo!

PR #330, opened 2023-07-26
Add system for Dynamic Registry Flags

rinOfTheStars:
This PR adds the beginnings of a system similar to (but far simpler than) the Static Registry Sync API for Dynamic Registries, allowing flags to be set which modify the behavior of Dynamic Registries.

Currently, only one flag, `DynamicRegistryFlag.OPTIONAL`, exists—this flag is a clone of the similar `SyncOption#skipWhenEmpty`, and thus only effects synchronized dynamic registries, and simply has no effect on non-synced dyn. registries. 

Both the `OPTIONAL` flag (and it's FAPI counterpart) set whether a dynamic registry should attempt to be synced with a connecting client which does not have said dynamic registry. This allows for clients without certain dynamic registries to still be able to connect, and thus is useful when creating entirely server-sided mods intended to be compatible with vanilla.

More flags can easily be added in the future, however the infrastructure for flags that apply to *all* dynamic registries, and not just synced ones, has not yet been set up—a future PR to add such a flag would need to add a mixin for that purpose somewhere other than `DynamicRegistrySync`. 

Note that the test mod for the registry API is ironically too complicated to serve as a proper test for this API addition—I have created a test mod locally (using the `mavenLocal()` repo) and tested having a vanilla client connect to a pure QSL + testmod server with an optional dynamic registry defined in the testmod, and it thankfully did work. If required, I can upload this testmod's repo to github with instructions on how to run it.

PR #329, opened 2023-07-23
Fix: Modify Tag path behavior in Dynamic Registries

rinOfTheStars:
Causes all modded Dynamic Registries to use `tags/NAMESPACE/path/to/file` for tag paths instead of `tags/path/to/file`, where `NAMESPACE` is the namespace of the Dynamic Registry. 

This prevents accidental namespace collision in tags, and could be considered a bugfix as it was intended behavior in the Dynamic Registry API to begin with.

PR #347, opened 2023-09-28
1.19.4 feature backports 

rinOfTheStars:
Backports the following changes required for or related to QFAPI sync:
- https://github.com/QuiltMC/quilt-standard-libraries/commit/2ff57f22e8dd610dcb2cfe234452c8319a0fc01b
- https://github.com/QuiltMC/quilt-standard-libraries/pull/330
- https://github.com/QuiltMC/quilt-standard-libraries/pull/329
- https://github.com/quiltmc/quilt-standard-libraries/commit/a8991ca686397828301b9d62d7289ecdc5707381
- https://github.com/quiltmc/quilt-standard-libraries/commit/c63c4893af453974f3798e9315cb05cb0a03b35a

Also ran the licenser, which created... _quite the few file changes_ that were just updating copyright year numbers. Tell me if I should revert that or not, please.

If there's more that needs to be done _please let me know_, this is a draft PR for a reason—I'm not sure if I caught everything that's absolutely necessary!


OroArmor:
Are you on windows? That messes up the licenser. If you revert the commit i can push a commit with the correct headers.

rinOfTheStars:
> Are you on windows? That messes up the licenser

nope, I'm on macOS, and it's worked fine in the past. If the headers are malformated in this case too, it might be a more general issue with the copyright year incrementing function? 
Also, looking at the headers they all look fine, as mentioned prior it's almost all just updating the copyright year range to contain 2023

OroArmor:
Hm interesting. Not sure at all, but yeah i would revert that last commit for now

OroArmor (approved):

EnnuiL:
> > Are you on windows? That messes up the licenser
> 
> nope, I'm on macOS, and it's worked fine in the past. If the headers are malformated in this case too, it might be a more general issue with the copyright year incrementing function? Also, looking at the headers they all look fine, as mentioned prior it's almost all just updating the copyright year range to contain 2023

The fix would be to backport the license header change, since yeah, the range is a lost cause

rinOfTheStars:
> > > Are you on windows? That messes up the licenser
> > 
> > 
> > nope, I'm on macOS, and it's worked fine in the past. If the headers are malformated in this case too, it might be a more general issue with the copyright year incrementing function? Also, looking at the headers they all look fine, as mentioned prior it's almost all just updating the copyright year range to contain 2023
> 
> The fix would be to backport the license header change, since yeah, the range is a lost cause

Ah, uh, how would that be done? Discussion is going on when it comes to licenser bs in the discord as well fyi, that's probably a better place to handle this technicality specifically.

rinOfTheStars:
Other than the licenser issues, this should be good to go ahead. Going to make it into a normal PR.

EnnuiL (approved):

PR #356, opened 2023-12-09
Support for 1.20.3/4

moehreag:
Tasks:
- [ ] decide on which version this should get (please just edit it in before merging)
- [ ] possibly create a new branch for this pr to target
- [ ] figure out a solution for the test mods as they currently do not get loaded correctly because loom requires the `mods` block

Information: builds succeed, but gradle will still spit out errors about the javadocs as well as about not finding some of minecraft's classes. I don't know if these "errors" were present before, but the mod works fine.


OroArmor:
Looks good from a cursory glance. Once I have time to fix 1.20.2, I'll merge that into this and then make a new 1.20.3 branch for 1.20.3/4 and then merge this

library/data/advancement/src/main/resources/quilt_advancement.mixins.json
```diff
@@ -3,7 +3,6 @@
   "package": "org.quiltmc.qsl.advancement.mixin",
```
OroArmor: This whole file can be deleted and add a `noMixins()` to the build.gradle for the module

library/data/recipe/src/main/resources/quilt_recipe.accesswidener
```diff
@@ -1,4 +1,2 @@
 accessWidener   v2  named
```
OroArmor: Similarly, this can be deleted.

library/worldgen/dimension/src/testmod/java/org/quiltmc/qsl/worldgen/dimension/QuiltDimensionTest.java
```diff
@@ -108,10 +107,6 @@ private int swapTargeted(CommandContext<ServerCommandSource> context) throws Com
 			var target = new TeleportTarget(new Vec3d(0.5, 101, 0.5), Vec3d.ZERO, 0, 0);
 			QuiltDimensions.teleport(player, modWorld, target);
 
-			if (player.getWorld() != modWorld) {
```
OroArmor, line 111: Why was this removed?

OroArmor (changes requested):

library/worldgen/dimension/src/testmod/java/org/quiltmc/qsl/worldgen/dimension/QuiltDimensionTest.java
moehreag: I removed it because at the time I was unaware of the alternatives to the previously used `CommandException`. It will be re-added with an updated exception type.

OroArmor (approved):

FirstMegaGame4:
Why is the target branch named 1.20.3 and not 1.20.4? 1.20.3 is just 1.20.4 with a missing markDirty

OroArmor:
20.3 and 20.4 are compatible. if we updated on time, we would stay on the 20.3 branch and add support from there.

PR #366, opened 2024-03-30
Fix Recipe Remainder API

### LINKED ISSUE #251: Duplication bug with furnace inputs and recipe remainders
QSL's recipe remainder hooks add remainder support to furnace inputs; this breaks smelting recipes that use water buckets as inputs - vanilla does not compute the remainder there, so the bucket should be destroyed. Quilt adds a remainder bucket instead, causing a duplication bug with MI.

https://github.com/AztechMC/Modern-Industrialization/blob/2f46d55691c1381b206adaefd1b91d794bf91918/src/main/resources/data/modern_industrialization/recipes/vanilla_recipes/steam_bucket.json

OroArmor:
A fairly substantial rework of the Recipe Remainder API.

The main idea behind this change is that the vanilla recipe remainders are used in two different contexts, and only those two (Crafting and Furnace Fuel). By arbitrarily extending this to all the supported locations, mods that relied on the behavior only being used in the vanilla locations broke. A great example was smelting a water bucket into a steam bucket, and duping the bucket (#251).

This PR fixes this by forcing mods to explicitly register their remainders for different locations than the default vanilla ones. This is a breaking change to the API, but no mods will crash, only their functionality will be impacted. 

This PR also fixes the remainders not working in the new smithing table and extends the system to work in both the template and base slots. 

Fixes #251


LambdAurora (approved):

library/item/item_setting/src/main/java/org/quiltmc/qsl/item/setting/api/RecipeRemainderLocation.java
```diff
@@ -0,0 +1,106 @@
...
+
+	/**
+	 * Remainder location for the smithing ingredient slot.
+	 */
+	RecipeRemainderLocation SMITHING_INGREDIENT = getOrCreate(new Identifier("minecraft:smithing_ingredient"));
+
+	/**
+	 * The default locations for a recipe remainder.
+	 */
+	RecipeRemainderLocation[] DEFAULT_LOCATIONS = {CRAFTING, FURNACE_FUEL};
```
Platymemo: Should we make default locations be a mutable list so modded locations can automatically opt-in? Would take some more finagling so that the remainders apply after all the locations are added though.

library/item/item_setting/src/main/java/org/quiltmc/qsl/item/setting/api/RecipeRemainderLocation.java
```diff
@@ -0,0 +1,106 @@
...
+ * Calling {@link #getOrCreate(Identifier)} allows mods to create their own remainder locations or get remainder locations without needing to compile against the other mod.
+ *
+ * <p> This class should not be extended.
+ */
+@ApiStatus.NonExtendable
+public interface RecipeRemainderLocation {
+	/**
+	 * Remainder location for Vanilla crafting. Used in Crafting Tables and the inventory crafting screen.
+	 */
+	RecipeRemainderLocation CRAFTING = getOrCreate(new Identifier("minecraft:crafting"));
```
Platymemo: Just a nitpick, but `new Identifier("minecraft", "crafting")` or `new Identifier("crafting")` are clearer/neater IMO.

Platymemo (approved):
The kind of solution where you kick yourself for not seeing it before! Looks great!

library/item/item_setting/src/main/java/org/quiltmc/qsl/item/setting/api/RecipeRemainderLocation.java
OroArmor: Maybe? But those would probably just be crafting. Most items will only target one of these locations.

Platymemo: Additionally, some items might want to blanket target all locations, that should be covered somehow

OroArmor: Ah I think i have an idea for this

OroArmor: done!

library/item/item_setting/src/main/java/org/quiltmc/qsl/item/setting/api/RecipeRemainderLocation.java
OroArmor: The first is more characters, so no real improvement, and the latter hides `minecraft`, which i want to stay clear.

OroArmor:
FCP for 2 days.

PR #373, opened 2024-06-03
Backport fixed Recipe Remainder API to 1.20.1

EnnuiL:
This also updates dependencies in preparation for another API-adding PR


OroArmor (approved):
Looks good. Didn't see any issues in the port. Was it fully tested?

EnnuiL:
> Looks good. Didn't see any issues in the port. Was it fully tested?

I ran through all tests in the testmod, and yeah, I can confirm it all works (especially after merging your fixes in, because without those, it crashed!); The only test I kinda forgot about is the brewing stand one though, but I could give a try now

EnnuiL:
> > Looks good. Didn't see any issues in the port. Was it fully tested?
> 
> I ran through all tests in the testmod, and yeah, I can confirm it all works (especially after merging your fixes in, because without those, it crashed!); The only test I kinda forgot about is the brewing stand one though, but I could give a try now

Update: I'll have to double-check the brewing stand stuff at home, the missing texture hecker got eaten by the brewing stand with no remainders left, which I believe is not supposed to happen with non-dragon's breath stuff

(oh also I might update CI workflows too)

PR #375, opened 2024-07-06
Port a few modules up to 1.21

OroArmor:
Also makes the DISABLE files per module, making it less of a diff to enable them.


eristhea:
the deletions?? you're cooking??

OroArmor:
> the deletions?? you're cooking??

Nope! Just using Mixin Extras

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/MutablePack.java
```diff
@@ -293,6 +294,14 @@ default void putImage(Identifier id, Supplier<NativeImage> imageSupplier) {
 		}));
 	}
 
+	/**
+	 * Adds a pack overlay to the mutable pack.
+	 *
+	 * @param overlay the overlay name
+	 * @param pack    the pack
+	 */
+	void putOverlay(@NotNull String overlay, ResourcePack pack);
```
OroArmor, line 303: @ix0rai @TheGlitch76 Im not too sure about this. Should it be a `ResourcePack`, where any pack can be part of the mutable pack, or does `Consumer<ResourcePack>` make more sense where the `MutablePack` implementation creates the pack in the way that it wants? Both options have their advantages.

eristhea:
@OroArmor can you compile a list of modules that you've ported in this PR?

OroArmor:
The list is:
- Everything

Ported already (thanks glitch):
- base
- crash

eristhea:
oh perf ok

library/core/resource_loader/src/main/java/org/quiltmc/qsl/resource/loader/api/MutablePack.java
TheGlitch76: I have zero expertise in this area, it's really up to you...

TheGlitch76 (approved):

PR #383, opened 2024-08-23
Improve mixin quality for Status Effect API

OroArmor:
Fixes a few crashes with this code. Should be fairly easy to backport to older versions if need be. I think some of the changes with `StatusEffect` -> `Holder<StatusEffect>` would need to be reverted.

### DIFF
--- a/library/entity/status_effect/src/main/java/org/quiltmc/qsl/entity/effect/mixin/LivingEntityMixin.java
+++ b/library/entity/status_effect/src/main/java/org/quiltmc/qsl/entity/effect/mixin/LivingEntityMixin.java
@@ -16,17 +16,20 @@
 
 package org.quiltmc.qsl.entity.effect.mixin;
 
+import java.util.Collection;
+import java.util.Iterator;
 import java.util.Map;
 
+import com.google.common.collect.Iterators;
+import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
+import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
 import org.jetbrains.annotations.NotNull;
 import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
-import org.spongepowered.asm.mixin.Overwrite;
 import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.Unique;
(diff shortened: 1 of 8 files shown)


library/entity/status_effect/src/main/java/org/quiltmc/qsl/entity/effect/mixin/LivingEntityMixin.java
```diff
@@ -41,8 +44,11 @@
 import org.quiltmc.qsl.entity.effect.api.StatusEffectRemovalReason;
 import org.quiltmc.qsl.entity.effect.api.StatusEffectUtils;
 import org.quiltmc.qsl.entity.effect.impl.QuiltStatusEffectInternals;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
 
-@Mixin(LivingEntity.class)
+// We want to make sure that our wrap operations are put before other mods, so that we wrap the vanilla call and not a mods call.
```
eristhea:
```suggestion
// We want to make sure that our wrap operations are put before other mods, so that we wrap the vanilla call and not a mod's call.
```

eristhea (approved):

library/entity/status_effect/src/main/java/org/quiltmc/qsl/entity/effect/mixin/LivingEntityMixin.java
```diff
@@ -41,8 +44,11 @@
 import org.quiltmc.qsl.entity.effect.api.StatusEffectRemovalReason;
 import org.quiltmc.qsl.entity.effect.api.StatusEffectUtils;
 import org.quiltmc.qsl.entity.effect.impl.QuiltStatusEffectInternals;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
 
-@Mixin(LivingEntity.class)
+// We want to make sure that our wrap operations are put before other mods, so that we wrap the vanilla call and not a mod's call.
+// This is because we do not call the vanilla method, so any mod adding something will not be called.
+@Mixin(value = LivingEntity.class, priority = 500)
```
TheGlitch76: it would probably be wise to set the default priority to around this number globally

TheGlitch76 (approved):

PR #388, opened 2025-03-09
Port to 1.21.5

Pixaurora:
So far, I've gotten everything to at least compile in 1.21.4 except for two modules: resource loader & registry.

@maloryware did more than half the work here, so I would like to thank him for all the time he put in!

Edit: Another big thanks to @supersaiyansubtlety and @alikindsys for their time contributing to this since they did stuff after I first made the draft PR! Also, it might be nice if you don't squash so it's easier to port, as I've already begun that.

PR #393, opened 2025-07-18
Port to 1.21.6

Pixaurora:
This PR is all in one commit, but relies on #388

### DIFF
--- a/library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/TextSerializationUtil.java
+++ b/library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/TextSerializationUtil.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2025 The Quilt Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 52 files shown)


library/core/registry/src/main/java/org/quiltmc/qsl/registry/impl/TextSerializationUtil.java
```diff
@@ -0,0 +1,43 @@
...
+		return JsonHelper.toSortedString(encodeResult.getOrThrow(message -> new RuntimeException("expected text to be encodeable")));
+	}
+
+	public static Text fromJson(String json, DynamicRegistryManager registryManager) {
+		var element = JsonParser.parseString(json);
+
+		var ops = RegistryOps.create(JsonOps.INSTANCE, registryManager);
+		var encodeResult = TextCodecs.CODEC.decode(ops, element);
+
+		return encodeResult.getOrThrow(message -> new RuntimeException("expected text to be decodeable")).getFirst();
```
supersaiyansubtlety, line 41: these exception  messages should probably include the error `message` from the result
minecraft also threw `JsonParseException`s on 1.21.5 which I think is good for specificity

PR #396, opened 2025-09-09
Registry Builder API

Leo40Git:
*Welcome back, #185.*  
Adds the `QuiltRegistryBuilder` class to build new registries via a fluent builder pattern.


Leo40Git:
Marked as a draft until I add tests, ~~and also until I figure out where to register the dynamic registry; `DynamicMetaRegistry` only handles the codecs, it seems~~.

Leo40Git:
So apparently you don't register dynamic registries, or even create `Registry<?>` objects for them?

Welp, makes my life easier.

Leo40Git:
Test mod is in!

library/core/registry/src/testmod/resources/quilt.mod.json
```diff
@@ -12,6 +12,7 @@
     "load_type": "always",
     "entrypoints": {
       "init": [
+        "org.quiltmc.qsl.registry.test.builder.RegistryLibBuiltinBuilderTest",
```
supersaiyansubtlety:
```suggestion
        "org.quiltmc.qsl.registry.test.builder.RegistryLibBuilderTests",
```
see <https://github.com/QuiltMC/quilt-standard-libraries/actions/runs/17590969310/job/49971374490#step:5:11466>

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/QuiltRegistryBuilder.java
```diff
@@ -0,0 +1,264 @@
...
+	public @NotNull QuiltRegistryBuilder<T> withBootstrap(Consumer<Registry<T>> bootstrap) {
+		this.bootstrap = bootstrap;
+		return this;
+	}
+
+	/**
+	 * Sets the default identifier of this registry.
+	 * <p>
+	 * Should a nonexistent entry be referenced in some way, the registry will instead reference the entry identified
+	 * by this instead.
```
supersaiyansubtlety:
```suggestion
	 * Should a nonexistent entry be referenced in some way, the registry will instead reference the entry identified
	 * by this.
```
double "instead"

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/QuiltRegistryBuilder.java
```diff
@@ -0,0 +1,264 @@
...
+		if (this.defaultId == null) {
+			registry = new SimpleRegistry<>(this.key, this.lifecycle, this.useIntrusiveHolders);
+		} else {
+			// this takes the identifier as a string, to guarantee that it's unique
+			registry = new DefaultMappedRegistry<>(this.defaultId.toString(), this.key, this.lifecycle, this.useIntrusiveHolders);
+		}
+
+		Registry.register((Registry<Registry<Object>>) Registries.ROOT, this.key.getValue(), (Registry<Object>) registry);
+
+		if (this.syncBehavior == SyncBehavior.REQUIRED || this.syncBehavior == SyncBehavior.OPTIONAL) {
```
supersaiyansubtlety:
```suggestion
		if (this.syncBehavior != SyncBehavior.SKIPPED) {
```
seems like this is more to the point

library/core/registry/src/main/java/org/quiltmc/qsl/registry/api/QuiltRegistryBuilder.java
```diff
@@ -0,0 +1,264 @@
...
+		 */
+		SKIPPED,
+		/**
+		 * The registry <em>will</em> be synchronized to the client,
+		 * and clients who do not have this registry on their side <em>will</em> be kicked.
+		 */
+		REQUIRED,
+		/**
+		 * The registry <em>will</em> be synchronized to the client,
+		 * and clients who do not have this registry on their side <em>will not</em> be kicked.
```
supersaiyansubtlety:
```suggestion
		 * but clients who do not have this registry on their side <em>will not</em> be kicked.
```
seems more appropriate, to differentiate it from `REQUIRED`

MrCat12-Studios:
I agree with supersaiyan

library/core/registry/src/testmod/resources/quilt.mod.json
Leo40Git: this is what I get for copy-pasting without checking.

MrCat12-Studios:
Why use copilot in the first place?

Leo40Git:
> Why use copilot in the first place?

I DID NOT USE COPILOT, it was a joke about how "generative" AI does not actually generate anything

the line was copied from the ancient original PR, and I forgot I renamed the class

supersaiyansubtlety (changes requested):
Looks good, just some nitpicks

There are checkstyle violations.
(we don't have a checkstyle task because a few things have to break it, but you can point [this plugin](<https://plugins.jetbrains.com/plugin/1065-checkstyle-idea>) to the project checkstyle and use it to scan)

*edit: fix link
Also [builder methods shouldn't have a "with" prefix](<https://github.com/QuiltMC/quilt-mappings/blob/25w06a/CONVENTIONS.md#factories-and-builders>).

Leo40Git:
Possible renames, based on the `Registries.create*Registry` methods:
- `withIntrusiveHolders()` to `frozen()` (apparently this is a side effect of the `useIntrusiveHolders` parameter)
- `withDefaultId(Identifier defaultId)` -> `defaultMapping(Identifier id)`

supersaiyansubtlety:
Oh I forgot one thing: since this uses the `@Experimental` `RegistrySynchronization`, this should also be `@Experimental`

supersaiyansubtlety (approved):

MrCat12-Studios (approved):
I don't think it needs any more changes. I've heard that this could be useful too so I think the job here is done.

**- Approved**

OroArmor (approved):
Was this tested with separate client and server instances?

Leo40Git:
not yet, will test today

Leo40Git:
Testmods have been added for synchronization!

Unforunately I was unable to actually _use_ them to test synchronization; I can't convince the server to let a vanilla client connect, Fabric clients don't work due to an unrelated issue in the registry module, and it treats other Quilt clients as vanilla(???).

MrCat12-Studios:
Is this the issue you opened? So that I can see the right log.

Leo40Git:
I cannot for the life of me get the testmods actually working; Quilt clients and Quilt servers are apparently incompatible.

MrCat12-Studios:
Is there any progress?

Leo40Git:
This PR is complete, but I can't do testing due to unrelated issues...

MrCat12-Studios:
What else do we need for it to be merged?

